PR #190, opened 2015-01-26
tox.ini: Add py34

msabramo:


HonzaKral:
Merged as 957792c
Thanks!

PR #194, opened 2015-02-10
Only update scroll id if all shards were successful for scanning

ntucker:
Sometimes a shard will fail on a scan. If it does, it won't be aware of the new scan id, so the old one should still be used until successful.

If you have many shards, decent load, and scan over a large enough set, you'll definitely run into this problem.

### DIFF
--- a/elasticsearch/helpers/__init__.py
+++ b/elasticsearch/helpers/__init__.py
@@ -34,7 +34,7 @@ def expand_action(data):
 
 
 def streaming_bulk(client, actions, chunk_size=500, raise_on_error=False,
-        expand_action_callback=expand_action, raise_on_exception=True, 
+        expand_action_callback=expand_action, raise_on_exception=True,
         **kwargs):
     """
     Streaming bulk consumes actions from the iterable passed in and yields
@@ -232,7 +232,8 @@ def scan(client, query=None, scroll='5m', preserve_order=False, **kwargs):
             break
         for hit in resp['hits']['hits']:
             yield hit
-        scroll_id = resp.get('_scroll_id')
+        if resp["_shards"]["successful"] == resp["_shards"]["total"]:
+            scroll_id = resp.get('_scroll_id')
         if scroll_id is None:
             break
 


HonzaKral:
Thanks for the patch, it makes sense. We should just build in some safeties - what if one shard is down, then this will end in a permanent loop. Also some people might prefer to ignore this or fail right away.

ntucker:
If the shard ever comes back it will be wrong to continue at that point. Therefore if a shard goes permanently down it needs to be removed to make progress.

Is there a way to ignore a shard after it fails? If so there could be an option to simply ignore failed shards from that point on.

As far as failing, that's easy to add as an option.

ntucker:
Alternatively we could have a timeout option that fails if it doesn't get all shards. This would cover both cases (in the second case set timeout to 0).

HonzaKral:
So I did some tests and unfortunately this approach doesn't work reliably.

in my test case I created an index with 3 shards, no replicas, on 2 nodes. I ran a few steps of the loop inside the scan helper then paused and killed one node, getting an unassigned shard. Then I let the loop iterate a few more times, paused again, started the node and waited for the shard to get into STARTED state. The scan still didn't pick up where it left off and some documents were missed. Occasionally I would even get a 404 error since the restarted node had no record of the search state.

Now some of the errors (404) were caused by the method of my test rather than the logic, but it still makes me want to not include this because it's unreliable.

I am afraid the best we can do is: add logging for missed shards (this is a must-have no matter what else we do) as well as some parameter to fail with an exception if not all shards were reached (with a safe - throwing errors - default).

Does this make sense to you?

thanks!

ntucker:
My company is currently using this logic to scan through up to 10 million documents in a cluster with 10 seperate machines, and many more shards.

I think maybe what we were experiencing and why this fixes it is more of a transient failure on part of the shard. i.e., it is timing out with its response, but the shard is not actually down and it is still communicating with it's neighbors. We actually don't usually have full outages, but this is still needed to deal with transient timeouts.

This is the difference between us being able to scan 10million documents and not fail, and just stopping after a short time, which there is no way to pick up. It absolutely does not work without handling this.

So I think if a shard is fully taken down we should have some sort of error for this method, but simply not getting a response is not enough to fail the scan or we would literally never be able to complete a scan.

HonzaKral:
The problem with this scenario is that it is unreliable, it can fail. If you imagine a scenario where a scroll request comes in with `scroll_id`- it is a last request to touch segment `A` which has already been merged and is kept around just because of the opened scroll. Which means that after request with `scroll_id` is finished it will get removed.

In this situation, if you were to repeat the request it will fail since it will reference a segment that is no longer there. The chances of that happening are low, of course, but it's possible. That is why I don't think it's a good choice for us - you can take the risk on your cluster but it cannot be part of the official API.

What I will do instead is document this and make sure scan fails when it encounters a failed shard (with an option to ignore failures).

Does this make sense?
Thanks

ntucker:
The thing is, the existing way is MORE unreliable. Would you be OK with an option to do what we need in our system? Currently we have to use a more verbose method instead of just iterating because we can't use this API and this method works for our case. I think many other people this will be true as well. What you described is in the realm of possibilty but what happens in practice for us is that shards just miss communication ocassionally, but enough that our scans never work without handling that case.

HonzaKral:
I understand your use case but unfortunately I cannot, in good conscience, include that in the client - it may break in unexpected fashion at unexpected times. It would be giving people false sense of security which is, in my opinion, one of the worst things a library can do.

It is also something that doesn't really *have* to be in the client for you to use it - you can still have your own function that solves your specific use case. I am happy to help you with the implementation (though you seem to be doing great on your own) and document the option for others though - I will add an example to the raw `scroll` method on how to use it if you wish to roll your own loop instead of using the helper.

I understand your frustration, but the thing is that your cluster is unstable, it shouldn't be (I will be happy to help you diagnose or fine tune the problem, feel free to reach out to me personally) since at that point most bets are off about what happens in the cluster. We cannot optimize for such a case.

PR #196, opened 2015-02-11
self.connections needs to be an array not a tuple

swapnam84:
When connecting to a single host in a cluster with sniffing enabled, it tried to add all the cluster members to the DummyConnectionPool. This then fails with:

TypeError: can only concatenate list (not "tuple") to list

Changing the connections to an array in the constructor gets it working...


HonzaKral:
Thanks for the report. This has been fixed in 7a674db

I will release 1.4 version shortly so this will be available on PyPI.

HonzaKral:
No time like the present - https://pypi.python.org/pypi/elasticsearch/1.4.0

PR #200, opened 2015-02-17
Added an option to suppress ssl check warning this warning appears in re...

imanhodjaev:
Added an option called *show_ssl_warnings=True/False* to suppress ssl check warning this warning appears in request client at 
*http_urllib3* and *http_requests* https://github.com/elasticsearch/elasticsearch-py/blob/master/elasticsearch/connection/http_urllib3.py#L49 https://github.com/elasticsearch/elasticsearch-py/blob/master/elasticsearch/connection/http_requests.py#L50

Now if we want to hide warnings like *Connecting to %s using SSL with verify_certs=False is insecure.* we can pass this keyword argument like
```python
Elasticsearch(hosts=HOSTS, show_ssl_warnings=False)
```

### DIFF
--- a/elasticsearch/connection/http_requests.py
+++ b/elasticsearch/connection/http_requests.py
@@ -25,7 +25,7 @@ class RequestsHttpConnection(Connection):
     """
     def __init__(self, host='localhost', port=9200, http_auth=None,
         use_ssl=False, verify_certs=False, ca_certs=None, client_cert=None,
-        **kwargs):
+        show_ssl_warnings=True, **kwargs):
         if not REQUESTS_AVAILABLE:
             raise ImproperlyConfigured("Please install requests to use RequestsHttpConnection.")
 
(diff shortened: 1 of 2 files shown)


HonzaKral:
What is the motivation behind this feature? I consider it a good thing that a library warns you when you are doing something insecure and I don't see much reasons to add an option to silence these warnings.

imanhodjaev:
I've been working on CLI utility to obtain statistics from Elasticsearch so these warnings are annoying. 
By default it will always show this warning except we decide to hide then intentionaly.

HonzaKral:
For that I would recommend explicitly silencing the warnings using warning filters (0). The best way, however, is of course to not allow users to run insecure configuration or let them see the warning if they do.

I am sorry but I cannot accept this PR because it can hide important messages and give users false sense of security. If they want to silence the warnings explicitly there is obviously nothing we can do, but I strongly believe these warnings should be there.

0 - https://docs.python.org/2/library/warnings.html#the-warnings-filter

imanhodjaev:
Thank you for your time and review

HonzaKral:
Btw - best way to handle this is of course to specify `verify_certs=True` and provide the CA pack by certifi, then you should be covered. Then, you just provide the users of your CLI with the option not to verify (if using self-signed certs for example) and they won't be surprised by the warning then. It's what I would do.

imanhodjaev:
Nice, honestly my bad that I forgot about warning filters. 
Some people in our team are fine with providing certificates others not so I decided suppress these warnings unless certificates provided.

PR #206, opened 2015-03-11
socket.error is now caught in perform_request aswell.

### LINKED ISSUE #205: max_retries is ignored when using thrift
connection/thrift.py does not catch socket.error in perform_request which cause max_retries to be ignored in the transport.

One can argue that this socket.error exception should be translated into a TException in the offical thrift-library but I think it's better to have it included here aswell.

sgadrian:
Fixes #205

### DIFF
--- a/elasticsearch/connection/thrift.py
+++ b/elasticsearch/connection/thrift.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 from socket import timeout as SocketTimeout
+from socket import error as SocketError
 import time
 import logging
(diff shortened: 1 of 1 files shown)


HonzaKral:
Thanks for this fix, could I please ask you to sign our CLA (0) so it can be merged?

https://www.elastic.co/contributor-agreement/

sgadrian:
Sure, it's done now!

HonzaKral:
Thanks!

PR #226, opened 2015-04-28
Fixes issue introduced in 482f166 where the tracer spams when

jimjkelly:
unconfigured.

In the aforementioned commit, the existing behavior was changed,
I believe in advertantly, to set propagate to False if the
trace logger already has been configured. This seems to be the
opposite of the internet, and this change restores what appears
to have been the intent, and silences the tracer unless
it has been configured, in which case it will propagate.

### DIFF
--- a/elasticsearch/connection/base.py
+++ b/elasticsearch/connection/base.py
@@ -12,7 +12,8 @@ logger = logging.getLogger('elasticsearch')
 # logger hasn't already been configured
 _tracer_already_configured = 'elasticsearch.trace' in logging.Logger.manager.loggerDict
 tracer = logging.getLogger('elasticsearch.trace')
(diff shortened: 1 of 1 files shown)


HonzaKral:
Good catch, thank you for the fix! Could I please ask you to sign our CLA (https://www.elastic.co/contributor-agreement/) so I can merge it?

jimjkelly:
Yup, should be all signed now. FYI on Safari the text of the agreement wasn't visible (obviously I can see it now that I've signed it).

HonzaKral:
Thank you!

jimjkelly:
No problem, thanks for all your work on the module! Also, ignore my comment on Safari, I think it was probably just Click2Flash.

Issue #185, opened 2015-01-14
SerializationError for datetime.time object


marekmalek:
JSONSerializer does not support time type.
```
elasticsearch/serializer.py

class JSONSerializer(object):
    mimetype = 'application/json'

    def default(self, data):
        if isinstance(data, (date, datetime)):
            return data.isoformat()
        elif isinstance(data, Decimal):
            return float(data)
        raise TypeError("Unable to serialize %r (type: %s)" % (data, type(data)))
```
 so I get a following error:

```
TypeError("Unable to serialize datetime.time
(10, 34) (type: <type 'datetime.time'>)"
```

A fix is quick and easy. Return time in isoformat like for date and datetime types:
```
class JSONSerializer(object):
    mimetype = 'application/json'

    def default(self, data):
        if isinstance(data, (date, time, datetime)):
            return data.isoformat()
        elif isinstance(data, Decimal):
            return float(data)
        raise TypeError("Unable to serialize %r (type: %s)" % (data, type(data)))
```

HonzaKral:
Hm, I originally didn't include time in the serialization because it may be misleading - we could serialize it automatically exactly as you describe but then elasticsearch won't recognize it and store it as string. To me that's slightly weird and that's why I avoided it, forcing people to deal with it manually.

What do you think, is it ok to just serialize it even when elasticsearch won't recognize it?

marekmalek:
I think if time doesn't have it's on core type in elasticsearch it should be indexed as string. If someone's implementation need different approach then they will take care of it.

HonzaKral:
I understand, I am just worried about the gap - with `datetime` we serialize it from python into string because we know elasticsearch will deserialize it correctly so it's transparent for the user. With `time` it wouldn't be transparent and imho could be very frustrating for the users. This way it fails and forces the user to deal with the situation. In the spirit of "no opinions in the client" I have to close this issue.

If you really want this behavior, it's very easy to subclass `JSONSerializer` and add the `time` handling into its `default` method yourself. You then just pass in a serializer instance into `Elasticsearch` when creating the client instance.

zeinababbasi:
@HonzaKral, would you please explain how to serialize a `datetime` object? Is there any known format like ISO or something else which I can use to provide a datetime object to elasticsearch?

HonzaKral:
[...]

PR #232, opened 2015-05-11
datetime.time support for JSONSerializer

mojeto:
datetime.datetime is subclass of datetime.date class.
This PR is adding support for serializing datetime.time objects.  

datetime.time.isoformat() is returning time in same ISO 8601 as datetime.datetime.isoformat() method does.

### DIFF
--- a/elasticsearch/serializer.py
+++ b/elasticsearch/serializer.py
@@ -2,7 +2,7 @@ try:
     import simplejson as json
 except ImportError:
     import json
-from datetime import date, datetime
+from datetime import date, time
 from decimal import Decimal
 
 from .exceptions import SerializationError, ImproperlyConfigured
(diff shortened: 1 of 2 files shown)


HonzaKral:
Thanks for the PR, this has already been proposed before as #185 and we have decided against the implementation mostly because it is unclear what elasticsearch will do with the data - it will, by default, just index it as string, which is most of the time the last thing you want.

What do you think?

mojeto:
A default elasticsearch behaviour is not the best in this case. You have to configure mapping for the field manually to date type with correct time format first. 
I agree that the its probably better to failure time object serialization to highlight it.
Creation of a child JSONSerializer class what will deal with time objects is good solution I'm good with it. 

But would be nice to have this problem and its solution documented somewhere. 

Feel free to decline this PR.

HonzaKral:
I totally get the desire to have a sensible time representation. In my eyes though that would need to be tied to the mappings etc, which this library has no concept of.

This is probably a good candidate for a field type for `elasticsearch-dsl` (0) which could handle it gracefully.

I will close the PR here and open an issue there.

Thanks again for the PR and discussion. If you feel like this is something you could use even without that, there is nothing stopping you from creating your own subclass of the serializer and passing it to the `Elasticsearch` class as `serializer` kwarg, it doesn't have to be a patch in the library itself.

0 - https://github.com/elastic/elasticsearch-dsl-py

PR #235, opened 2015-05-24
Add support for native app engine urlfetch

sadovnychyi:
Adds support for native urlfetch for app engine environments.
No test coverage for now, seems its a bit tricky about how to run them.

https://cloud.google.com/appengine/docs/python/urlfetch/

### DIFF
--- a/elasticsearch/__init__.py
+++ b/elasticsearch/__init__.py
@@ -20,6 +20,6 @@ from .connection_pool import ConnectionPool, ConnectionSelector, \
     RoundRobinSelector
 from .serializer import JSONSerializer
 from .connection import Connection, RequestsHttpConnection, \
-    Urllib3HttpConnection, MemcachedConnection, ThriftConnection
+    Urllib3HttpConnection, MemcachedConnection, ThriftConnection, \
+    UrlfetchHttpConnection
 from .exceptions import *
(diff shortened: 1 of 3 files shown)


HonzaKral:
Thanks for the patch!

Do you have any tests how this performs compared to the default backend? It should be compatible with app engine.

sadovnychyi:
Yes, the default backend is compatible. I don't have any tests, but it makes a bit of sense to use the native one. I did it only because I saw that both requests and urllib3 are supported -- they both do the same thing, so why not to add native urlfetch?

HonzaKral:
You have a good point about the `requests` backend - I have been considering removing it for a long time. The only reason why it's still there is that a lot of people with more advanced use cases (like different auth schemes) are more comfortable with extending the `requests`-based backend than the default one.

For a minority use case like app engine I see no reason to support two alternate backends in the same library. I'd be happy to have a link to the implementation in the docs for people who'd want to use it, but at present wouldn't want the code to live in `elasticsearch-py` itself.


Thanks for the patch, but currently it doesn't make sense to include it -  the code can live in it's own package as an addon without any loss of functionality or flexibility, it doesn't add anything new and introduces new overhead (more code, much harder testing).

PR #242, opened 2015-06-23
A few other cat APIs accept 'bytes'.

cxmcc:
/nodes
/shards
/segments


cxmcc:
local test passed, CLA signed, not sure what is going on here.

HonzaKral:
Thanks! Don't worry about the CI tests, they are currently broken because of an issue with the yaml tests in elasticsearch.

PR #251, opened 2015-07-09
Regex to allow hostnames as well as IP addresses

danpilch:
I have modified the regex to support obtaining "host.name.com" from '/_nodes/_all/clear' when using elasticsearch's sniff functionality.

e.g.
"http_address":"inet[host.name.com/1.2.3.4:443]"

I needed this because I am using a wildcard SSL certificate behind an nginx proxy and the current implementation produces:

SSLError: ConnectionError(hostname u'1.2.3.4' doesn't match either of '*.name.com', 'name.com') caused by: SSLError(hostname u'1.2.3.4' doesn't match either of '*.name.com', 'name.com')

Using this regex so the client uses the hostname to connect instead of the IP address solved my problem.

### DIFF
--- a/elasticsearch/transport.py
+++ b/elasticsearch/transport.py
@@ -8,8 +8,11 @@ from .serializer import JSONSerializer, Deserializer, DEFAULT_SERIALIZERS
 from .exceptions import ConnectionError, TransportError, SerializationError, \
                         ConnectionTimeout, ImproperlyConfigured
 
-# get host/ip and port from "inet[wind/127.0.0.1:9200]" or "inet[127.0.0.1:9200"
-ADDRESS_RE = re.compile(r'(?P<host>[-a-zA-Z0-9\.]+(?=/[\.:0-9a-f]*:)|(?<=/)[\.:0-9a-f]*)[^:]*:(?P<port>[0-9]+)\]?$')
+# get host/ip and port from one of the following format:
+#   "inet[wind/127.0.0.1:9200]"
+#   "inet[/127.0.0.1:9200]"
+#   "127.0.0.1:9200"
+ADDRESS_RE = re.compile(r'(?P<host>[-a-zA-Z0-9\.]+(?=/[\.:0-9a-f]*:)|(?<=/)[\.:0-9a-f]+|^[\.:0-9a-f]+)[^:]*:(?P<port>[0-9]+)\]?$')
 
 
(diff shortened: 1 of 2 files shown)


HonzaKral:
Thanks for the PR, could you please provide some tests with the code? Thanks

danpilch:
added test case

HonzaKral:
Thanks, I appreciate the tests!

Could I please ask you to sign our CLA (https://www.elastic.co/contributor-agreement/) so that I can use this code?

I have to check with the other clients to decide if we should make this the default since it might not work in some cases (internal hostnames, ...), we might also just expose it as an option.

danpilch:
We did sign the Corporate CLA but it doesn't seem to have been accepted (as organisation incopro)?

danpilch:
Hi, we have now attempted to sign the agreement twice (both times successfully).

danpilch:
any update on the CLA? we've signed two times but still shows that we haven't done so...

electrical:
Hi @danpilch for some reason the system is not updating the github PR's with the right status.
I do see the Corp and individual CLA's signed.
I'll make sure our system is updated.

I also see that the 2 commits are done by 2 different authors of which one not with the corp email address. See https://patch-diff.githubusercontent.com/raw/elastic/elasticsearch-py/pull/251.patch for details.

bowensong:
@electrical I have done a `git push --force` to make sure all commits' email addresses are CLA's signed.

bowensong:
@HonzaKral Could you please review this PR? Thanks.

HonzaKral:
I am really sorry this is taking so long, I have one last request of you - could you make it not use regular expressions? The solution was fine for the simple use case of extracting the ip address and port but now it's getting a bit too complex where what we really need is extract the inside of `[]` (which can be done with regex or by hard-coded slicing since the text is static) and then split by `/` and `:` (that way we can also reuse the code for 2.0 where the `inet[]` part is missing

i am more than happy to make the change (with full credit to you of course, you put a lot of work and patience into this), or feel free to change it. I promise I will be more responsive and merge asap.

Thanks and once again, apologies for the delays.

bowensong:
Sure, no problem, I will change it to not using complex regular expressions.

HonzaKral:
I decided on a slightly different approach where we don't use regexes at all, thanks again for the PR.

Issue #256, opened 2015-07-31
slightly malformed exception message ('Scrol request has failed on %d shards out of %d.', 110, 120)


sujaymansingh:
I assume it should be "Scroll" rather than "Scrol", also, I guess the exception string should be formatted?

HonzaKral:
Merged, thank you for the patch and my apologies for the delay!

PR #257, opened 2015-07-31
[#256] Fix exception message

sujaymansingh:
* Typo: Scrol -> Scroll
* We need to format the string with the args ourselves (ScanError won't
  do it).
  Otherwise the args in the ScanError won't be substituted.


HonzaKral:
Merged as e3d25e8
Thanks!

PR #266, opened 2015-09-09
Clarified _op_type options in streaming_bulk docstring.

veatch:


elasticsearch/helpers/__init__.py
```diff
@@ -66,8 +66,9 @@ def streaming_bulk(client, actions, chunk_size=500, raise_on_error=True,
     Alternatively, if `_source` is not present, it will pop all metadata fields
     from the doc and use the rest as the document data.
 
-    If you wish to perform other operations, like `delete` or `update` use the
-    `_op_type` field in your actions (`_op_type` defaults to `index`)::
+    The :meth:`~elasticsearch.Elasticsearch.bulk` api accepts `index`, `create`,
+    `delete`, and `update` actions. Optionally use the `_op_type` field to specify
```
HonzaKral: Thanks, this is much cleaner, but I think we should not say "Optionally" here - it might be misleading in that people would think that you don't need to specify the type of action even if you're not using `index`. What do you think?

veatch: Ok cool. I updated with that change.

HonzaKral:
Thanks!

PR #267, opened 2015-09-10
use GET method for count instead of POST

dmglab:
As documented in the elasticsearch API, count should be requested as GET.

My motivation is a simple rule to protect my data: public requests deny any POST/PUT/DELETE requests. Count is a read-only operation but requests it's data via POST.


HonzaKral:
Thanks for the patch!

PR #279, opened 2015-10-07
fixes issue on decoding str type body when logging on python3

am2rican5:
When using python3.5 an error is occured below
```
  File "<string>", line 3, in raise_exc_info
  File "/Users/leehyeon/.py3virtualenv/andong_api/lib/python3.5/site-packages/tornado/gen.py", line 876, in run
    yielded = self.gen.throw(*exc_info)
  File "/Users/leehyeon/.py3virtualenv/andong_api/lib/python3.5/site-packages/tornado_elasticsearch.py", line 175, in perform_request
    ignore=ignore)
  File "/Users/leehyeon/.py3virtualenv/andong_api/lib/python3.5/site-packages/tornado/gen.py", line 870, in run
    value = future.result()
  File "/Users/leehyeon/.py3virtualenv/andong_api/lib/python3.5/site-packages/tornado/concurrent.py", line 215, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 3, in raise_exc_info
  File "/Users/leehyeon/.py3virtualenv/andong_api/lib/python3.5/site-packages/tornado/stack_context.py", line 314, in wrapped
    ret = fn(*args, **kwargs)
  File "/Users/leehyeon/.py3virtualenv/andong_api/lib/python3.5/site-packages/tornado_elasticsearch.py", line 91, in on_response
    response.code, response.body, duration)
  File "/Users/leehyeon/.py3virtualenv/andong_api/lib/python3.5/site-packages/elasticsearch/connection/base.py", line 59, in log_request_success
    body = body.decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```

### DIFF
--- a/elasticsearch/connection/base.py
+++ b/elasticsearch/connection/base.py
@@ -55,7 +55,7 @@ class Connection(object):
 
         # body has already been serialized to utf-8, deserialize it for logging
         # TODO: find a better way to avoid (de)encoding the body back and forth
-        if body:
+        if body and isinstance(body, bytes):
             body = body.decode('utf-8')
 
         logger.info(
@@ -84,7 +84,7 @@ class Connection(object):
 
         # body has already been serialized to utf-8, deserialize it for logging
         # TODO: find a better way to avoid (de)encoding the body back and forth
-        if body:
+        if body and isinstance(body, bytes):
             body = body.decode('utf-8')
 
         logger.debug('> %s', body)


HonzaKral:
I believe this is an issue with `tornado_elasticsearch` since it's copy of the `Transport` class didn't include some logic, notably https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/transport.py#L288-L293

The client itself is structured to only use bytes to avoid any possible conflicts and to provide accurate header information (content-length). If you choose to replace some of it's internals (which is definitely supported and encouraged if the client doesn't work for you as-is), you need to adhere to the protocol.

Do you think it would be worth documenting this requirement for custom `Transport` classes?

Thanks!

am2rican5:
Thank you for your detailed answer!

PR #286, opened 2015-10-16
Move import statement into a function

### LINKED ISSUE #282: Does not work on app engine anymore because of new parallel bulk
This [line](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/helpers/__init__.py#L4) prevent it from working under app engine, where multiprocessing is not supported. Exception looks like:
```
 from multiprocessing.dummy import Pool
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/multiprocessing/__init__.py", line 65, in <module>
    from multiprocessing.util import SUBDEBUG, SUBWARNING
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/multiprocessing/util.py", line 41, in <module>
    from subprocess import _args_from_interpreter_flags
```

How do we can prevent multiprocessing module to being imported while running on app engine?
Its possible to do explicit check, like it is done in [NLTK](https://github.com/nltk/nltk/blob/develop/nltk/data.py#L770) or we could move import statement inside of parallel_bulk function, which is against PEP8 – but it will solve the problem.

sadovnychyi:
To fix a bug on App Engine where multiprocessing is not allowed.
Closes #282

### DIFF
--- a/elasticsearch/helpers/__init__.py
+++ b/elasticsearch/helpers/__init__.py
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 
 import logging
-from multiprocessing.dummy import Pool
 from operator import methodcaller
 
 from ..exceptions import ElasticsearchException, TransportError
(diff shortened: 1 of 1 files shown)


HonzaKral:
Merged as 2d04178

Thanks!

PR #290, opened 2015-10-26
Renaming manpage to elasticsearch-py

hydrapolic:
Currently elasticsearch-py installs the manpage as "elasticsearch" - that manpage should belong to elasticsearch itself.

before:
$ man elasticsearch

after:
$ man elasticsearch-py

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -228,7 +228,7 @@ latex_documents = [
 # One entry per manual page. List of tuples
 # (source start file, name, description, authors, manual section).
 man_pages = [
-    ('index', 'elasticsearch', u'Elasticsearch Documentation',
+    ('index', 'elasticsearch-py', u'Elasticsearch Documentation',
      [u'Honza Král'], 1)
 ]
 


idella:
Confused. I thought we did this by the sed statement and dropped this

hydrapolic:
Some background: 
with idella, we recently added version 1.8 to gentoo portage:
https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-python/elasticsearch-py/elasticsearch-py-1.8.0.ebuild

We already applied the aforementioned change:

python_prepare_all() {
	# rename manpage to elasticsearch-py
	sed \
		-e "s@('index', 'elasticsearch'@('index', 'elasticsearch-py'@g" \
		-i docs/conf.py || die

	distutils-r1_python_prepare_all
}

I opened this pull request so if this is accepted we can remove that line from the ebuild and it will be available for all. I'm making the same change here as we did in the ebuild. Clear?

HonzaKral:
Sorry it took so long, merged as 9415c74

hydrapolic:
Thanks

PR #298, opened 2015-11-03
Adding assert_hostname and assert_fingerprint parameters to Urllib3HttpConnection

### LINKED ISSUE #291: Allow setting assert_hostname when using urllib3
In some instances, it may be necessary to set the `assert_hostname` when using the built-in `urllib3.HTTPSConnectionPool` in order to disable hostname verification (e.g., internal SSL certificates that do not supply a hostname).

It would be great if the parameter could be passed in without having to subclass. It may be worthwhile to also to support `assert_fingerprint` at the same time.

pickypg:
Fixes #291


HonzaKral:
Merged as f3315dc, thanks!

PR #299, opened 2015-11-04
Addition of AwsHttpConnection

vsheffer:
I've added a Connection implementation that will provide the necessary HTTP payload to authenticate with an AWS elasticsearch cluster where access has been granted to a specific IAM user.

### DIFF
--- a/elasticsearch/__init__.py
+++ b/elasticsearch/__init__.py
@@ -20,6 +20,6 @@ from .connection_pool import ConnectionPool, ConnectionSelector, \
     RoundRobinSelector
 from .serializer import JSONSerializer
 from .connection import Connection, RequestsHttpConnection, \
-    Urllib3HttpConnection
+    Urllib3HttpConnection, AwsHttpConnection
 from .exceptions import *
 
--- a/elasticsearch/connection/__init__.py
+++ b/elasticsearch/connection/__init__.py
@@ -1,3 +1,4 @@
 from .base import Connection
 from .http_requests import RequestsHttpConnection
 from .http_urllib3 import Urllib3HttpConnection
+from .http_aws import AwsHttpConnection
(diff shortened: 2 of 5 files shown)


HonzaKral:
Thank you for the patch, I am not sure however if we want all the logic in the library itself. So far I was leaning towards just docummenting the approach mentioned in https://github.com/sam-washington/requests-aws4auth/pull/2

Is there anything your solution does that the external one doesn't? Thanks!

vsheffer:
I haven't looked at the botocore library, but I doubt very much that my implementation does anything different.  The algorithm is well documented and pretty simple and AWS has a site that even provides the example python code.  By leveraging your nice Connection/Transport framework, the implementation was quite easy and has no impact on the rest of the library.

The advantage, and it may be minor, to my solution is that it eliminates the dependency on botocore that you may not need installed.  I'm using elasticsearch-py with https://github.com/nbs-system/naxsi/tree/master/nxapi and that will be deployed into a Docker container.  In accordance with my general approach of minimizing as much as possible what goes into a Docker container, I want to minimize the dependencies as much as possible.  Botocore is one library, but it depends on a handful more, and each of those might depend on more.  I haven't done a full dependency analysis, so I'm not sure of the total transitive dependencies brought in by botocore.  Given the simplicity of adding it directly to elasticsearch-py, I thought it was easier to do that than worry about botocore.

I'm guessing that a lot of people who use elasticsearch-py are also using much of the rest of the AWS CLI, which means that dependency will already be installed.  For that use case, botocore will also already be installed.  In the case of my Docker container, however, it will only be running nxtools, and AWS cli will not be installed.

Thanks.

HonzaKral:
I am not worried about dependencies and few extra kB on disk or docker image. What I am worried about is maintenance of the code This is a lot of code that doesn't belong to the elasticsearch client. It belongs to a library that deals with AWS API. If I go with the other approach I don't have to maintain any aws-related code (I don't use aws and therefore that is hard for me) and can instead depend on people maintaining a solid aws api library. For those reasons I will go with the other approach.

Thanks again for the patch. For yourself, if you want to use it in your code, you can just add it to your app code or make it into a separate package. If you do, please let me know and I will link to the solution from the docs next to the botocore one.

PR #306, opened 2015-11-16
Add bytes param to cat.shards and cat.segments

michaelgibson:
'bytes' appears to be a valid parameter for the shards and segments endpoints in the CAT API.
Would like to be able to specify this in the python functions.


HonzaKral:
Merged as 70d6203

Thanks for the PR and sorry for the delay

PR #317, opened 2015-12-10
add the ability to serialize uuid types

apeters:
It would be great to apply this to both the 1.x and 2.x versions

### DIFF
--- a/elasticsearch/serializer.py
+++ b/elasticsearch/serializer.py
@@ -2,6 +2,7 @@ try:
     import simplejson as json
 except ImportError:
     import json
+import uuid
 from datetime import date, datetime
 from decimal import Decimal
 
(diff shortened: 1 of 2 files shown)


apeters:
Just signed the CLA so we should be good there.

HonzaKral:
Thanks!

Sphinxs:
@apeters , you have converted the UUID object to `str` ([here](https://github.com/elastic/elasticsearch-py/pull/317/commits/2e77648ebb978168adfa59fc4b76f3c02ef5fe3d#diff-09cd79840fe7f5e55c9cd9d36a70b58aR33)), in your mappings have you set `binary` or `string` ?

When you need to compare the UUID object to the value stored at Elasticsearch it's necesseraly to convert the object to string and then compare it ? If so, it looks kinda a manually work

It's possible to consider the UUID, GUID and the Postgre implementation adding new cases!

PR #321, opened 2015-12-14
Forward kwargs from scan helper to scroll method as well.

andreip:
### Description

I've called `elasticsearch.helpers.reindex` with a custom `request_timeout` present in both `scan_kwargs` and `bulk_kwargs`. But that's not sent down to `client.scroll` which would send it to `self.transport.perform_request` which would then use it.

The first call from `scan` helper does forward the `**kwargs`: `client.search(body=query, scroll=scroll, **kwargs)`, so I'd say `client.scroll` should as well. Because this wasn't happening, I eventually got a timeout near the end of a huge reindexing with the below output, even though I supplied a `request_timeout=600`.

``` bash
....
  File "/Users/andrei/.virtualenvs/genesis/lib/python2.7/site-packages/elasticsearch/helpers/__init__.py", line 342, in _change_doc_index
    for h in hits:
  File "/Users/andrei/.virtualenvs/genesis/lib/python2.7/site-packages/elasticsearch/helpers/__init__.py", line 283, in scan
    resp = client.scroll(scroll_id, scroll=scroll)
  File "/Users/andrei/.virtualenvs/genesis/lib/python2.7/site-packages/elasticsearch/client/utils.py", line 69, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/Users/andrei/.virtualenvs/genesis/lib/python2.7/site-packages/elasticsearch/client/__init__.py", line 662, in scroll
    params=params, body=body)
  File "/Users/andrei/.virtualenvs/genesis/lib/python2.7/site-packages/elasticsearch/transport.py", line 307, in perform_request
    status, headers, data = connection.perform_request(method, url, params, body, ignore=ignore, timeout=timeout)
  File "/Users/andrei/.virtualenvs/genesis/lib/python2.7/site-packages/elasticsearch/connection/http_urllib3.py", line 86, in perform_request
    raise ConnectionTimeout('TIMEOUT', str(e), e)
elasticsearch.exceptions.ConnectionTimeout: ConnectionTimeout caused by - ReadTimeoutError(HTTPSConnectionPool(host='139f46bdca40404618dec7e5a8ecad29.us-east-1.aws.found.io', port=9243): Read timed out. (read timeout=10))
```
### Later edit

I saw the failing tests, `**kwargs` wasn't thought as being forwarded on `scroll`. And adding an inline case for forwarding only `request_timeout` there is a hack. But I was trying to make use of the fact that it's gonna be added to params in [here](https://github.com/elastic/elasticsearch-py/blob/799918712fc1136b82eacf24fc2a5860da2dadaa/elasticsearch/client/utils.py#L66). If there's an easy fix here I'd appreciate a suggestion though.

### DIFF
--- a/elasticsearch/helpers/__init__.py
+++ b/elasticsearch/helpers/__init__.py
@@ -232,7 +232,9 @@ def parallel_bulk(client, actions, thread_count=4, chunk_size=500,
     pool.close()
     pool.join()
 
-def scan(client, query=None, scroll='5m', raise_on_error=True, preserve_order=False, **kwargs):
+def scan(client, query=None, scroll='5m', raise_on_error=True,
+        preserve_order=False, global_kwargs={}, search_kwargs=None,
+        scroll_kwargs=None, **kwargs):
     """
     Simple abstraction on top of the
     :meth:`~elasticsearch.Elasticsearch.scroll` api - a simple iterator that
@@ -254,6 +256,13 @@ def scan(client, query=None, scroll='5m', raise_on_error=True, preserve_order=Fa
         cause the scroll to paginate with preserving the order. Note that this
         can be an extremely expensive operation and can easily lead to
         unpredictable results, use with caution.
+    :arg global_kwargs: additional kwargs to be passed to both
+        :meth:`~elasticsearch.Elasticsearch.search` and to
+        :meth:`~elasticsearch.Elasticsearch.scroll`
+    :arg search_kwargs: additional kwargs to be passed to
+        :meth:`~elasticsearch.Elasticsearch.search`
+    :arg scroll_kwargs: additional kwargs to be passed to
+        :meth:`~elasticsearch.Elasticsearch.scroll`
 
     Any additional keyword arguments will be passed to the initial
     :meth:`~elasticsearch.Elasticsearch.search` call::
@@ -268,7 +277,10 @@ def scan(client, query=None, scroll='5m', raise_on_error=True, preserve_order=Fa
     if not preserve_order:
         kwargs['search_type'] = 'scan'
     # initial search
-    resp = client.search(body=query, scroll=scroll, **kwargs)
+    search_kwargs = search_kwargs or {}
+    search_kwargs.update(global_kwargs)
+    search_kwargs.update(kwargs)
+    resp = client.search(body=query, scroll=scroll, **search_kwargs)
 
     scroll_id = resp.get('_scroll_id')
     if scroll_id is None:
@@ -280,7 +292,9 @@ def scan(client, query=None, scroll='5m', raise_on_error=True, preserve_order=Fa
         if preserve_order and first_run:
             first_run = False
         else:
-            resp = client.scroll(scroll_id, scroll=scroll, **kwargs)
+            scroll_kwargs = scroll_kwargs or {}
+            scroll_kwargs.update(global_kwargs)
+            resp = client.scroll(scroll_id, scroll=scroll, **scroll_kwargs)
 
         for hit in resp['hits']['hits']:
             yield hit
--- a/elasticsearch/helpers/__init__.py
+++ b/elasticsearch/helpers/__init__.py
@@ -280,7 +280,7 @@ def scan(client, query=None, scroll='5m', raise_on_error=True, preserve_order=Fa
         if preserve_order and first_run:
             first_run = False
         else:
-            resp = client.scroll(scroll_id, scroll=scroll)
+            resp = client.scroll(scroll_id, scroll=scroll, **kwargs)
 
         for hit in resp['hits']['hits']:
             yield hit
--- a/test_elasticsearch/test_server/test_helpers.py
+++ b/test_elasticsearch/test_server/test_helpers.py
@@ -1,4 +1,4 @@
-from elasticsearch import helpers, TransportError
+from elasticsearch import helpers, TransportError, NotFoundError
 
 from . import ElasticsearchTestCase
 from ..test_cases import SkipTest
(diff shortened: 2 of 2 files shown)


HonzaKral:
Thanks for this PR! It makes a lot of sense but we need to revise it a little bit - the `search` request supports more `kwargs` than the `scroll` api, passing the same kwargs to both can therefore be dangerous and might produce errors.

What do you think would be a better solution?
* have a `global_kwargs` (or some better name) that would be passed to both + `search_` and `scroll_` kwargs?
* pass the kwargs to `search` and add another kwarg called `scan_kwargs` (similar to what reindex does)
* try to detect which kwarg goes where (my least favorite option since it would tie it to specific api version and would make maintenance more difficult)

andreip:
@HonzaKral I followed your first suggestion, that seemed a clean way to go for me. But now you can send kwargs to the `client.search` call in multiple ways:
- the old way of calling scan was by forwarding all from `**kwargs` of the `scan` method to the `client.search` call
- the new `search_kwargs` kwarg I added (due to symmetry and being more clean, since I also added `scroll_kwargs` and `global_kwargs` to the `scan` method)

If I were to remove the old way of sending parameters to the `client.search` (since having multiple ways might create confusion?), then that would be fine for this repo, only a few modifications I'd have to do, but the previous code of people using the library might not work with new release. Should I leave it backwards compatible as is now?

dtran320:
:+1: for this

elasticsearch/helpers/__init__.py
```diff
@@ -268,7 +277,10 @@ def scan(client, query=None, scroll='5m', raise_on_error=True, preserve_order=Fa
     if not preserve_order:
         kwargs['search_type'] = 'scan'
     # initial search
-    resp = client.search(body=query, scroll=scroll, **kwargs)
+    search_kwargs = search_kwargs or {}
+    search_kwargs.update(global_kwargs)
```
HonzaKral: I think these need to be reversed - start with `global_kwargs.copy()` and then do `.update(search_kwargs)` - that way `search_kwargs` will have priority and will be able to override `global_kwargs`. Other way might be counter intuitive.

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -184,26 +184,44 @@ def test_errors_are_collected_properly(self):
...
-        bulk = []
-        for x in range(100):
-            bulk.append({"index": {"_index": "test_index", "_type": "answers", "_id": x}})
-            bulk.append({"answer": x, "correct": x == 42})
-        self.client.bulk(bulk, refresh=True)
+    def test_general_kwargs_forwarded_to_search(self):
+        inexistent_index = 'test_index_123'
+        self.assertRaises(
+            NotFoundError,
+            lambda: list(helpers.scan(self.client, index=inexistent_index, doc_type="answers", size=2))
```
HonzaKral: no need for the `lambda` here, `self.assertRaises(NotFoundError, list, helpers.scan(...))` will do the trick

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -184,26 +184,44 @@ def test_errors_are_collected_properly(self):
...
+        )
+        global_kwargs = {'ignore': 404}
+        list(helpers.scan(self.client, index=inexistent_index, doc_type="answers", size=2, global_kwargs=global_kwargs))
```
HonzaKral: it would be nice to verify the response here, even when the real test is to see no error was raised (or at least have a comment to that effect so readers know what is being tested)

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -184,26 +184,44 @@ def test_errors_are_collected_properly(self):
...
+
+    def test_general_kwargs_forwarded_to_scroll(self):
```
HonzaKral: this is a complex and expensive test that can be a bit cofusing. Could we perhaps just have a unit test using a mocked client (using `ElasticsearchTestCase` you can find example of how to supply your own body here: https://github.com/elastic/elasticsearch-py/blob/master/test_elasticsearch/test_transport.py#L171) and just manually verify proper params were passed through for each call?

elasticsearch/helpers/__init__.py
```diff
@@ -254,6 +256,13 @@ def scan(client, query=None, scroll='5m', raise_on_error=True, preserve_order=Fa
...
+    :arg global_kwargs: additional kwargs to be passed to both
+        :meth:`~elasticsearch.Elasticsearch.search` and to
+        :meth:`~elasticsearch.Elasticsearch.scroll`
+    :arg search_kwargs: additional kwargs to be passed to
+        :meth:`~elasticsearch.Elasticsearch.search`
+    :arg scroll_kwargs: additional kwargs to be passed to
+        :meth:`~elasticsearch.Elasticsearch.scroll`
 
     Any additional keyword arguments will be passed to the initial
     :meth:`~elasticsearch.Elasticsearch.search` call::
```
HonzaKral: If we have this here we should remove the `search_kwargs` and just use `kwargs` there, it would also be backwards compatible just like the current solution and simpler.

HonzaKral:
Sorry for the delay (and thanks to @dtran320 for the ping!). I left some line notes to clear things up a bit. I am more than happy to implement these but would like your opinion.

Thanks!

fxdgear:
bum @andreip this PR is pretty out of date and has some merge conflicts. Would you be willing to update this?

andreip:
@fxdgear thanks for the ping! I see that @HonzaKral fixed the main problem in https://github.com/elastic/elasticsearch-py/commit/d4efb81b0695f3d9f64784a35891b732823a9c32 . This is a great short fix, that was the issue I had too, only scroll wasn't using `request_timeout`; bulk was forwarding that properly to `transport.perform_request`.

Gonna close this. 👍

PR #326, opened 2015-12-21
Adding msearch_template to Elasticsearch client

danygielow:
I added an end point call to /_msearch/template which is available since elasticsearch version 2.1.0 (see elastic/elasticsearch#12414).
Also added .idea and .eggs directories to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ coverage.xml
 nosetests.xml
 junit-*.xml
 build
+.idea/
+.eggs/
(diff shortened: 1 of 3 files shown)


HonzaKral:
Thanks for catching this!

It has not yet been properly documented in elasticsearch so I will hold on merging the PR until it is clear what the client API should look like (we need to maintain consistency across language clients). I filed an issue in ES to keep track of this - https://github.com/elastic/elasticsearch/issues/15674

HonzaKral:
This has been fixed by 87bf9866

Thanks for the patch!

PR #327, opened 2015-12-21
use self.host instead of 'localhost' in trace logs

jraby:


HonzaKral:
Thanks for the PR. This behavior is, however, as intended - the `trace` logger is meant as a replayable script or a transcript of your es session to be shared when communicating with other (like on forums, replicating bugs, ...). It is therefore set to `localhost:9200` to not leak your configuration as well as allow you to replay the session locally without editting the file. The `elasticsearch` logger logs properly and exactly the hostname and all other information.

Hope this helps

PR #353, opened 2016-02-09
Add doc for filter_path argument in search

Garrett-R:
I figured out that the `search` function supported this by digging in the source.

Let me know if my way of describing it is not clear...

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -63,6 +63,19 @@ connection class::
 
 .. py:module:: elasticsearch
 
+Response Filtering
+~~~~~~~~~~~~~~~~~~
+
+The `filter_path` parameter is used to reduce the response returned by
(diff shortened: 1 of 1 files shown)


Garrett-R:
Hmmm... I signed the CLA, but I'm still getting a fail for the CI...

HonzaKral:
Thanks for the PR, the `filter_path` parameter actually doesn't belong to `search` but it is a global parameter accepted by all rest endpoints. It's documentation belongs to https://github.com/elastic/elasticsearch-py/blob/master/docs/api.rst#global-options instead.

Garrett-R:
Oh yeah, good call!  Ok, how does it look now?

HonzaKral:
Great, thank you!

Garrett-R:
Whoa, that was fast! Thanks.

PR #362, opened 2016-02-22
Urllib3HttpConnection should have separate client_cert and client_key…

### LINKED ISSUE #344: Urllib3HttpConnection should have separate client_cert and client_key arguments
Many deployments have the client cert and key in separate files, for security purposes.  You usually want much tighter control of who can read the key file.

Note that the function that is eventually called with these arguments is `ssl.wrap_socket()` https://docs.python.org/3.5/library/ssl.html#ssl.wrap_socket which has separate keyfile and certfile arguments (although certfile supports a file with both the cert and key in it).

richm:
… arguments #344

https://github.com/elastic/elasticsearch-py/issues/344
Add a client_key parameter to complement the client_cert parameter.  If
client_key is used, it is assumed that client_cert will contain only the
certificate, not the combined cert and key in a single file.

Closes #344

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -194,6 +194,18 @@ elasticsearch cluster, including certificate verification and http auth::
         ca_certs=certifi.where(),
     )
 
+    # SSL client authentication using client_cert and client_key
+
+    es = Elasticsearch(
+        ['localhost', 'otherhost'],
+        http_auth=('user', 'secret'),
+        port=443,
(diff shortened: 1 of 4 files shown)


HonzaKral:
Merged as https://github.com/elastic/elasticsearch-py/commit/fe6e9a7f5c832fd02847b99f3e2cb84292eb660b

Thanks for the fix!

PR #364, opened 2016-02-24
Fixed a bare except

gst:
AFAIK the only exception which can occur here is one that the json.loads(..) could raise. So (TypeError or) ValueError.

So except on that. and also log a warning in such case.

bare excepts are 99.99% of the time simply bad/wrong. It's even more the case when the except block is a no-op / pass statement.. It's the case here.


gst:
I signed the CLA..

HonzaKral:
Merged as 879de75

Thank you!

PR #373, opened 2016-03-04
Explicit close methods

### LINKED ISSUE #372: Support for explicit closing and context managers
In our usage of the library we have found some cases where it is useful to explicitly close connections.  It seems that this can be done with:

``` python
client = Elasticsearch()
for conn in client.transport.connection_pool.connections:
    conn.pool.close()
```

Likewise, it could be useful to use the client as a context manager, so something like:

``` python
with Elasticsearch() as client:
    client.foo()
```

And have the connections explicitly closed on exit.  It turns out that this can be done pretty simply and I have a working implementation in a feature branch [here](https://github.com/wdm0006/elasticsearch-py/tree/adding-close). In particular, the additions are 202-231 [here](https://github.com/wdm0006/elasticsearch-py/blob/adding-close/elasticsearch/client/__init__.py#L202). 

I've updated the one of the examples to include close, and added another to demonstrate the context manager behavior and it seems to be working nicely, at least for my purposes.  If there is interest, I'd be happy to make any changes and submit a pull request to fold this functionality into the library.

wdm0006:
Fixes #372

### DIFF
--- a/elasticsearch/connection/pooling.py
+++ b/elasticsearch/connection/pooling.py
@@ -29,5 +29,5 @@ class PoolingConnection(Connection):
         """
         Explicitly close connection
         """
-        return True
+        pass
 
--- a/elasticsearch/connection/pooling.py
+++ b/elasticsearch/connection/pooling.py
@@ -24,3 +24,10 @@ class PoolingConnection(Connection):
 
     def _release_connection(self, con):
         self._free_connections.put(con)
+
+    def close(self):
+        """
+        Explicitly close connection
+        """
+        return True
+
--- a/elasticsearch/connection_pool.py
+++ b/elasticsearch/connection_pool.py
@@ -231,10 +231,8 @@ class ConnectionPool(object):
         """
         Explicitly closes connections
         """
-        for conn in self.connections:
+        for conn in self.orig_connections:
             conn.close()
-
-        return True
 
 class DummyConnectionPool(ConnectionPool):
     def __init__(self, connections, **kwargs):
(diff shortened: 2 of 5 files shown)


elasticsearch/connection_pool.py
```diff
@@ -227,6 +227,14 @@ def get_connection(self):
         # only one connection, no need for a selector
         return connections[0]
 
+    def close(self):
+        """
+        Explicitly closes connections
+        """
+        for conn in self.connections:
```
HonzaKral: we should probably try and close the dead connections too. Iterating over `orig_connections` instead should do it

elasticsearch/connection/pooling.py
```diff
@@ -24,3 +24,10 @@ def _get_connection(self):
 
     def _release_connection(self, con):
         self._free_connections.put(con)
+
+    def close(self):
+        """
+        Explicitly close connection
+        """
+        return True
```
HonzaKral: is there a reason why there it `True` being returned anywhere? I don't think we need a return value here

elasticsearch/connection_pool.py
```diff
@@ -241,6 +249,13 @@ def __init__(self, connections, **kwargs):
     def get_connection(self):
         return self.connection
 
+    def close(self):
+        """
+        Explicitly closes connections
+        """
+        for conn in self.connections:
```
HonzaKral: we only have one connection here so `self.connection.close()` would do it

HonzaKral:
Thanks for the PR! 

It looks good, just some minor comments on the diff. If I could also ask you to please sign our CLA (https://www.elastic.co/contributor-agreement/) so that I can merge it. Thanks!

wdm0006:
There should be a Predikto CLA in transit somewhere, other changes made.

HonzaKral:
Thanks for the very quick turnaround and the PR!

PR #377, opened 2016-03-08
documentation for basic auth over reverse proxy

thammegowda:
Added an example for connecting to elastic instance protected by a reverseproxy (like nginx) with basic authentication

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -219,6 +219,22 @@ description of the options.
 
 .. _certifi: http://certifi.io/
 
+Accessing over Reverse Proxy:
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you have a reverse proxy like nginx configured to protect your elastic server with basic http authentication, then remember the following :
+
++ if you have exposed your elastic search over http://example.com/examplepath, then ::
+
+    es = Elasticsearch(
+        ['http://example.com:80/examplepath'],
+        connection_class=RequestsHttpConnection,
+        http_auth=('user', 'secret'),
+        use_ssl=False,
+        verify_certs=False
+    )
+  
+NOTE: Failing to specify port :80 causes the client to assume default port `:9200` and thus it wont be able to talk to your server over reverse proxy.
+
 Logging
 ~~~~~~~
 


docs/index.rst
```diff
@@ -219,6 +219,22 @@ description of the options.
...
 
+Accessing over Reverse Proxy:
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you have a reverse proxy like nginx configured to protect your elastic server with basic http authentication, then remember the following :
+
++ if you have exposed your elastic search over http://example.com/examplepath, then ::
+
+    es = Elasticsearch(
+        ['http://example.com:80/examplepath'],
+        connection_class=RequestsHttpConnection,
```
HonzaKral: could I ask why you are changing the class here for `RequestsHttpConnection`? It is not required for this to work and is not recommended for simple use cases where the default `urllib3`-based connection works much better. Thanks!

fxdgear:
@thammegowda ping. 

Can you please address @HonzaKral's question. 

Thanks

docs/index.rst
thammegowda: I dont remember now ( after 2 years) why `RequestsHttpConnection` was used instead of the default transport.

1. I might have faced issue getting the default `connection_class` transport to use `http_auth` in the format I have specified. 
2. or maybe for the sake of getting logs from the transport (the default might not have provided the logs)

Going forward: If the default one works well, then there is no need to change `RequestsHttpConnection`

thammegowda:
@fxdgear I had missed this thread. Replied now! Thanks

fxdgear:
@thammegowda 

Great! Thank you. 

If you could please remove the `connection_class` and I'll get to merging this. 

Thank you

thammegowda:
@fxdgear I tried but I was not able to do so. See in the header the source of pull request is shown as `unknown repository` which means the connection to my forked repository is lost, maybe because I deleted my fork assuming I no longer need it. 

is it possible to make that change from your side - get diff/patch, edit, and then merge?
Alternatively, if you just want to discard/close this PR, thats okay too!

Issue #386, opened 2016-03-25
Compatibility with host parsing when sniffing host in elasticsearch ver 1.x


jffifa:
Host sniffing use "/_nodes/_all/clear" to get hosts information,
where address returned in elasticsearch ver 1.x in format `inet\[/IP:PORT\]` like
```
"transport_address" : "inet[/192.168.0.2:9300]"
```
while in ver 2.x in format `[HOSTNAME/]IP:PORT` like
```
"transport_address" : "192.168.0.2:9300"
```

Such case cause incompatibility with parsing host info using client ver 2.x in elasticsearch 1.x, which obstructs upgrading from 1.x to 2.x in some respects.

HonzaKral:
The client is not backwards compatible - if you use Elasticsearch 1.x you have to also use 1.x version of the client, same for 2.x - always keep the major version in sync.

There are other differences between the versions that cannot be bridged like this one so there is nothing we can do unfortunately - you have to chose the correct version to use.

PR #387, opened 2016-03-25
bugfix for host parsing in host sniff

jffifa:
bugfix for issue #386


HonzaKral:
Elasticsearch-py 2.x is not compatible with elasticsearch server 1.x. If you use elasticsearch 1.x please use 1.x version of the client as well.

Thanks

PR #390, opened 2016-03-29
Fixed query param 'timeout'

ayang:
Using 'timeout' param will raise an error, 'request_timeout' is the correct param name.


HonzaKral:
`request_timeout` is a global parameter accepted by all the APIs, at such it is defined here:

https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L66

and documented here:

http://elasticsearch-py.readthedocs.org/en/master/api.html#timeout

`timeout` is a different parameter that is only accepted by some APIs.

Hope this helps.

PR #411, opened 2016-06-03
Fix HTTPS URL support

hudolejev:
Current behavior:

```
>>> from elasticsearch import Elasticsearch
>>> x = Elasticsearch(['https://localhost'])
>>> x.transport.get_connection()
<Urllib3HttpConnection: http://localhost:443>
```

-- note that `Urllib3HttpConnection` contains `http://` URL -- should be `https://`.

With this fix applied:

```
>>> from elasticsearch import Elasticsearch
>>> x = Elasticsearch(['https://localhost'])
>>> x.transport.get_connection()
<Urllib3HttpConnection: https://localhost:443>
```

-- URL protocol in `Urllib3HttpConnection` is correct.

### DIFF
--- a/elasticsearch/client/__init__.py
+++ b/elasticsearch/client/__init__.py
@@ -42,12 +42,12 @@ def _normalize_hosts(hosts):
             if parsed_url.port:
                 h["port"] = parsed_url.port
 
+            if parsed_url.scheme:
+                h['scheme'] = parsed_url.scheme
+
             if parsed_url.scheme == "https":
                 h['port'] = parsed_url.port or 443
                 h['use_ssl'] = True
-                h['scheme'] = 'http'
-            elif parsed_url.scheme:
-                h['scheme'] = parsed_url.scheme
 
             if parsed_url.username or parsed_url.password:
                 h['http_auth'] = '%s:%s' % (parsed_url.username, parsed_url.password)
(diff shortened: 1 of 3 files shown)


wtorrealba:
Hi, this pull request has exactly what the driver needs to work with lambda function inside AWS

HTTPS URL is breaking the connection inside AWS lambda, any plan to make this proposed behavior part of the driver soon?

HonzaKral:
Thanks for the fix!

jgardezi:
I am having issues with connecting to AWS Elasticsearch when used in AWS Lambda function. When I run python function on local machine it works however when I put it in Lambda function it does not works. No connection error nothing on my cloudwatch logs.

This is code for python 2.7
```
    from elasticsearch import Elasticsearch, RequestsHttpConnection

    try:
        es = Elasticsearch(
            hosts=[{'host': config.es_host, 'port': 80, 'timeout': config.es_timeout}],
            connection_class=RequestsHttpConnection
        )
        print (es.info())
        # print(es.cluster.health())
        # print ('Connection info: '.format(es.info()))
    except ElasticsearchException as e:
        print(e)
        print('ElasticSearch client connection issue')
        raise e

    print('Loaded ES Client {}'.format(es))
```

HonzaKral:
@jgardezi thanks for the report, could you please provide any kind of error message or anything that could indicate what the problem is? Without this I cannot help you since it looks like the problem is in lambda and not the client since the code works from local machine.

Either way if you could also open it as a separate issue to keep things organized, thanks!

jgardezi:
@HonzaKral thank you for letting me know. I have posted my issue here https://github.com/elastic/elasticsearch-py/issues/623

PR #415, opened 2016-06-10
[Fix] fix sniff host when extract node address from es api response.

fatelei:
Current response from api `/_nodes/_all/clear` is below:

```
{
    "cluster_name": "test",
    "nodes": {
         "47UWFACERJuqujyi60jQMA": {
                 "name":"test-1",
                 "transport_address":"inet[/10.10.0.1:9300]",
                 "host": "test1",
                 "ip": "10.10.0.1",
                 "version":"1.7.2",
                "build":"e43676b",
                "http_address":"inet[/10.10.0.1:9200]",
                "attributes":{
                     "max_local_storage_nodes":"1",
                     "rack":"rack0","master":"true"
                }
         }
}
```

The format of  `transport_address` and `http_address` has changed. Do as the old way, it will lead `ValueError` when call `int(port)`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ coverage.xml
 nosetests.xml
 junit-*.xml
 build
+.eggs
--- a/elasticsearch/transport.py
+++ b/elasticsearch/transport.py
@@ -1,4 +1,6 @@
 import time
+import re
+
 from itertools import chain
 
 from .connection import Urllib3HttpConnection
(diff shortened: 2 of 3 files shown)


HonzaKral:
Can I ask what version of elasticsearch you are using? It looks like elasticsearch `1.x` for which you'd have to use one of the `1.x` releases or `elasticsearch-py`.

fatelei:
image

Yes, the version of es is `1.7.2`.

HonzaKral:
In that case you need to be using the 1.x version which contains the proper code to handle that format - https://github.com/elastic/elasticsearch-py/blob/1.x/elasticsearch/transport.py#L12

fatelei:
Thanks, @HonzaKral

PR #430, opened 2016-08-03
Update python doc: added timeout to params in delete() call

antonbormotov:
When I try to delete snapshot with huge indices (~5-10gb), delete call raises an HttpConnectionError exception: ReadTimeout

### DIFF
--- a/elasticsearch/client/snapshot.py
+++ b/elasticsearch/client/snapshot.py
@@ -21,7 +21,7 @@ class SnapshotClient(NamespacedClient):
         return self.transport.perform_request('PUT', _make_path('_snapshot',
             repository, snapshot), params=params, body=body)
 
-    @query_params('master_timeout')
+    @query_params('master_timeout', 'timeout')
     def delete(self, repository, snapshot, params=None):
         """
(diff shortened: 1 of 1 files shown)


HonzaKral:
Hi @antonbormotov, the `timeout` parameter you have added is an internal parameter to elasticsearch and won't change this behavior. To do that you need to specify `request_timeout` which is already supported - http://elasticsearch-py.readthedocs.io/en/master/api.html#timeout

Hope this helps

antonbormotov:
@HonzaKral That was I thought as well: delete call times out after 10 seconds with appropriate exception when `request_timeout` is specified 60 seconds.
When I specify `timeout` in params it does not timeout.
I can provide an example of output if necessary.

HonzaKral:
An example output would be great @antonbormotov since I have trouble replicating this. In the mean time I am reopening this since it looks like I was wrong and further investigation is needed.

Thank you!

HonzaKral:
I am closing this PR since there is no additional information and I am unable to reproduce this behavior. Please feel free to reopen with additional information. Thank you!

PR #433, opened 2016-08-05
Fix debugging log host

svdgraaf:
After some WTF moments during debugging, I found that the hostname that is printed in the logs, is actually hardcoded. I'm not sure as to why that is, but it took me a while to figure out that the calls where actually going to my ES cluster as they should, and not to my localhost (which accidentally also had an ES host running).

I hope this PR helps other people from not making the same mistake 😉

### DIFF
--- a/elasticsearch/connection/base.py
+++ b/elasticsearch/connection/base.py
@@ -73,7 +73,7 @@ class Connection(object):
             path = path.replace('?', '?pretty&', 1) if '?' in path else path + '?pretty'
             if self.url_prefix:
                 path = path.replace(self.url_prefix, '', 1)
-            tracer.info("curl -X%s '%s:9200%s' -d '%s'", method, self.host, path, _pretty_json(body) if body else '')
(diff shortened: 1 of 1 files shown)


svdgraaf:
I did sign the CLA, I think it takes some to time to process 😄

HonzaKral:
If you are interested in seeing what is actually going on I recommend using just the `elasticsearch` logger, not the `elasticsearch.trace` - that one is meant for different purpose - to be able to share your session and replay it later. That is also why  it has `localhost:9200` hardcoded, unlike the normal logger - https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/base.py#L64-L67

I understand that this is confusing, but it is the intedned behavior. Could we perhaps change the docs instead to make this clear?

Thanks!

svdgraaf:
@HonzaKral Thanks, that makes a lot of sense 👍  I'm not sure where I got the `elastisearch.trace` from, I guess someone on the internet was wrong 😉 

I'll close this PR.

PR #447, opened 2016-09-07
Custom headers in http requests with requests library 

haeygen:
I needed custom http headers in one of my projects. I have added a few lines of code to the requests library to support this. It is an additional parameter to init and adds these as headers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ python:
   - "3.4"
   - "3.5"
 
+jdk:
+  - "oraclejdk8"
+
 env:
   # different connection classes to test
   - TEST_ES_CONNECTION=Urllib3HttpConnection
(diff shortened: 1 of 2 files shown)


haeygen:
Ok, signed the CLA now.

HonzaKral:
This has been already implemented in 311ac7c and will be part of next release. Thank you for the patch!

haeygen:
HonzaKral, thanks for the info. Is it possible to see what is currently in development or planned already? I couldn't see anything about this in the milestones.

HonzaKral:
Typically I try to put all this information into the Changelog - https://github.com/elastic/elasticsearch-py/blob/master/Changelog.rst but sometimes I fall behind. Opening a ticket is a good way to learn as well.

PR #454, opened 2016-09-21
Optionally skip clear scroll in scan helper

wdm0006:
In the scan helper, currently the clear_scroll api is explicitly called at the end (on error or on success), which issues a DELETE to /_search/scroll.  As per the [elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html#_clear_scroll_api), this isn't actually required, but there is some cost to leaving the scroll there until it gets reaped, so if you can delete it, it's better to.  

In some cases, we will front elasticsearch with nginx and limit access to things like DELETE, which causes the scan helper to error out, because it gets 403 from nginx when trying to close the scroll that it really doesn't have to close. 

This pull request adds a parameter `clear_scroll` to the scan helper, defaulting to True, that specifies whether the DELETE call should be issued, so that by default nothing changes, but optionally a user can block delete calls with something like nginx and set `clear_scroll=False` to keep the scan helper functional.


wdm0006:
Not sure how the checker works, but I have already signed the CLA for a previous PR.   Travis seems to be failing trying to load the ES v5.0.0 temporary store in the test-suite setup, please advise on if that has anything to do with my PR or is just an issue in the travis setup.

Thanks, -Will

wdm0006:
@HonzaKral, just checking in, have you had a chance to look at this yet?

HonzaKral:
Could I please ask you to sign our CLA so that I can merge this? https://www.elastic.co/contributor-agreement

Thank you!

wdm0006:
My organization has previously signed it, see: https://github.com/elastic/elasticsearch-py/pull/373

HonzaKral:
Thanks, my bad!

wdm0006:
Is there any chance that this gets pulled into the 2.x branch and released in the near term? We'd rather not run off of our branch if at all possible.

PR #460, opened 2016-10-04
Fix delete by query link

lostick:
While I was going through the code, wondering why the delete-by-query plugin is not present in the branch 2.x, I realised that the doc link has been changed in the elasticsearch as this API is currently experimental.


lostick:
it looks tests are not passing because travis is not using java8

HonzaKral:
Thank you for the patch, the link, however, is correct. It is just for unreleased version of elasticsearch. Once 5.x gets released it will point to the correct URL.

PR #467, opened 2016-10-21
Attempt to force Travis CI to Java 8

brainix:


brainix:
Hi, @HonzaKral.

Tagging you so that you get a notification.  I at least got Elasticsearch 5.0.0 installing/running on Travis CI (with JDK 8) so that we can run the test suite.  However, as expected, some of the tests fail against Elasticsearch 5.0.0.

I can dig into the test failures, but that might be beyond the scope of this PR, which is just to get Travis CI up and running again!

Let me know what you think, and thanks for all of your hard work,
Raj

georgepsarakis:
@brainix I think you can use [Docker in Travis CI build](https://docs.travis-ci.com/user/docker/) with the [official Elasticsearch images](https://hub.docker.com/r/library/elasticsearch/tags/) in order to save you some effort and increase reliability.

PR #468, opened 2016-10-24
Support for Custom Authentication Handler in RequestsHttpConnection

mbown:
This allows for the use of the requests_kerberos custom authentication handler (amongst others).

Example use:

```
from requests_kerberos import HTTPKerberosAuth

es = Elasticsearch(
    hosts=["https://sample.com:443/"],
    auth=HTTPKerberosAuth(),
    connection_class=RequestsHttpConnection,
    use_ssl=True,
    verify_certs=True
)
```

### DIFF
--- a/elasticsearch/connection/http_requests.py
+++ b/elasticsearch/connection/http_requests.py
@@ -9,7 +9,6 @@ except ImportError:
 from .base import Connection
 from ..exceptions import ConnectionError, ImproperlyConfigured, ConnectionTimeout, SSLError
 from ..compat import urlencode, string_types
-from requests_kerberos import HTTPKerberosAuth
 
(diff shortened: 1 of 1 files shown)


HonzaKral:
If you pass the `HTTPKerberosAuth()` as value for `http_auth` then it will be passed through to `requests` and used. There is no need for extra parameter.

Hope this helps and apologies for the delay,
Honza

karmi:
Hi @mbown, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/468.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #471, opened 2016-10-26
allow to set a specific CA directory through ca_cert_dir option

zmathe:
I am having a problem setting up the authentication using host certificate, because I can not specify a directory path to the CA certificates. What I wanted to have:

``` pyhton
cl = Elasticsearch('https:', verify_certs=True,ca_cert_dir='/opt/xxx/yyyy/security/certificates/')
```

I know I can create a file which contains all CAs in /opt/xxx/yyyy/security/certificates/ directory, but this means I have to maintain this file always. 
I exposed the ca_cert_dir option and it works well. Could you please include to the next release?
Thanks,
Zoltan

### DIFF
--- a/elasticsearch/client/__init__.py
+++ b/elasticsearch/client/__init__.py
@@ -123,6 +123,19 @@ class Elasticsearch(object):
             ca_certs='/path/to/CA_certs'
         )
 
+    If using SSL with a certain CA directory, it can be specified in the following way::
+
+        es = Elasticsearch(
+            ['localhost:443', 'other_host:443'],
+            # turn on SSL
+            use_ssl=True,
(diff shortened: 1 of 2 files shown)


fxdgear:
I'm going to say lets now go down this route. 

I've added the ability to use SSLContext, which would account for the ability to specify a ca_cert or a ca_path which would do the same thing you're trying to do. 

But it would also be a little more flexible in allowing how ca's are handled on a case by case basis. 

Please see: https://github.com/elastic/elasticsearch-py/pull/635

If the SSLContext does not address your needs please re-open.

zmathe:
Thank you very much SSLContext even better for me! I am going to have a look you PR.

PR #473, opened 2016-10-27
add '_mapping' to exists_type path

nkwood:
This adds ‘_mapping’ to the path for exists_type to avoid the
deprecation warning consistent with
(https://github.com/elastic/elasticsearch/commit/a4ea7e72234b480f89e8c59bb12061be11172db1)


HonzaKral:
Thank you for this patch!

Issue #476, opened 2016-11-02
RequestsHttpConnection::perform_request except clause uses possibly-unbound 'response' variable


brianfaull:
When using helpers.bulk (which is fantastic!), I received:
    UnboundLocalError: local variable 'response' referenced before assignment

This was raised by connection/http_requests.py RequestsHttpConnection::perform_request(); see: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/http_requests.py#L70

The (small) error is that the `try:` block in that function attempts to set `response`, but this assignment could fail if an exception is thrown from requests.Session::request(); in this case, `response` is unbound in the `except:` clause, but is unconditionally dereferenced. 

An example of one possible patch; not fully tested:
https://github.com/brianfaull/elasticsearch-py/commit/89f51899a671ebd195eff22f20f624fa8783139f

Overview: Split the assembly of the request from the sending of the request, and use the request object (not the response object) for logging. Clearly there are other possibilities also.

This is not an issue for the analog within http_urllib3.py (however it seems that url and full_url/path_url arguments to self.log_request_fail() are transposed between http_urllib3.py and http_requests.py unless I misunderstand.)

I note your preference for signing the CLA and submitting a PR; please let me know if that is still your preference for a fix as small as this.

Many thanks for a wonderful and useful library!

HonzaKral:
Thank you for the find! I would greatly appreciate a PR (and yes, please do sign the CLA, it helps us protect our contributors as well as the project).

May I also ask why you are using the `requests` backend? Just curiosity - trying to compile a list of use cases.

Thanks!

jmphilli:
+1 (having same issue)

jimstjean:
+1 (just hit this too)

Ins1ghtLabs:
+1

orbuch:
+1

jimstjean:
Would be very helpful to get this fix reviewed and committed if possible.
In my case, we are using the Request backend in order to work with Amazon ES service (AWS with IAM) as documented here: https://elasticsearch-py.readthedocs.io/en/master/index.html

HonzaKral:
I have merged the PR and will push out a new release asap.

I apologize for the delays.

ThomasdOtreppe:
If I may ask, any update on the release?

PR #481, opened 2016-11-04
Fix UnboundLocalError response (#476)

javaguirre:
Hi! I've checked out @brianfaull's solution and works great! It's also how requests [`session.request` method works underneath](https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L411-L490), so It's good to go IMHO. :-)


HonzaKral:
Thank you for the PR!

Issue #477, opened 2016-11-02
Elasticsearch 5.0.0 does not work with sniffing


ShaneHarvey:
See https://github.com/elastic/elasticsearch/issues/21292.

Elasticsearch 5.0.0 does not return the http_address field on _nodes/_all/clear. Therefore, https://github.com/elastic/elasticsearch-py/blob/5.0.1/elasticsearch/transport.py#L206 needs to use `/_nodes/_all/http` or some other mechanism to determine which nodes have http interfaces.

This bug causes causes all sniffing to fail against a 5.0.0 server:
```python
>>> import elasticsearch
>>> from pprint import pprint
>>> elasticsearch.__version__
(5, 0, 1)
>>> es = elasticsearch.Elasticsearch('localhost:9200')
>>> pprint(es.transport.perform_request('GET', '/_nodes/_all/clear')['nodes'])
{u'EFrXB9v-SC6EgISgHyQK5w': {u'build_hash': u'253032b',
                             u'host': u'127.0.0.1',
                             u'ip': u'127.0.0.1',
                             u'name': u'EFrXB9v',
                             u'roles': [u'master', u'data', u'ingest'],
                             u'transport_address': u'127.0.0.1:9300',
                             u'version': u'5.0.0'}}
>>> pprint(es.transport.perform_request('GET', '/_nodes/_all/http')['nodes'])
{u'EFrXB9v-SC6EgISgHyQK5w': {u'build_hash': u'253032b',
                             u'host': u'127.0.0.1',
                             u'http': {u'bound_address': [u'[fe80::1]:9200',
                                                          u'[::1]:9200',
                                                          u'127.0.0.1:9200'],
                                       u'max_content_length_in_bytes': 104857600,
                                       u'publish_address': u'127.0.0.1:9200'},
                             u'ip': u'127.0.0.1',
                             u'name': u'EFrXB9v',
                             u'roles': [u'master', u'data', u'ingest'],
                             u'transport_address': u'127.0.0.1:9300',
                             u'version': u'5.0.0'}}
>>> es = elasticsearch.Elasticsearch('localhost:9200', sniff_on_start=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/shane/venv/elastic2-doc-manager/python2.7/lib/python2.7/site-packages/elasticsearch/client/__init__.py", line 171, in __init__
    self.transport = transport_class(_normalize_hosts(hosts), **kwargs)
[...]

PR #482, opened 2016-11-07
fix sniffing for ES v5.x, see #477

daveio:
This needs testing on clusters with more than one node, as I don't have a production cluster to test against.


daveio:
CI is broken, see #483

HonzaKral:
This has been fixed by fe53ab5

Issue #483, opened 2016-11-07
CI is broken


daveio:
CI is failing on all recent builds because https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=org.elasticsearch.distribution.zip&a=elasticsearch&e=zip&v=5.0.0-SNAPSHOT does not exist.

This should probably be updated to 5.1.0 - https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=org.elasticsearch.distribution.zip&a=elasticsearch&e=zip&v=5.1.0-SNAPSHOT

ShaneHarvey:
#478 also fixes this issue.

daveio:
@ShaneHarvey's PR is much cleaner than mine. Closing mine.

GlenRSmith:
Merged and passing. Thanks again @daveio and @ShaneHarvey

PR #478, opened 2016-11-03
Travis: install Java 8 and test against Elasticsearch 5.0.0

ShaneHarvey:

PR #484, opened 2016-11-08
fix 404 while pulling maven zip by bumping to 5.1.0-SNAPSHOT, see #483

daveio:
5.0.0-SNAPSHOT zip file is no longer available.


daveio:
Work on this PR now finished and it's ready for merge - there are build issues, but people better at ES than I need to look at those. At least the build runs properly now.

daveio:
Closing - #478 is a much cleaner solution.

Issue #380, opened 2016-03-16
Confusion at tracelog


sinanislekdemir:
Hi all;

During a debug trace, a tracer.info message fooled me for a while and caused me to loose some time before figuring out what the problem actually was. It made me think that I was making the requests to localhost instead of the remote host;

In connection/base.py:73

`tracer.info("curl -X%s 'http://localhost:9200%s' -d '%s'", method, path, _pretty_json(body) if body else '')`

"localhost:9200" needs to be replaced with %s -> self.host I believe,

Shall I create a PR for this?

https://github.com/sinanislekdemir/elasticsearch-py/commit/6624fc42bc542a85c3faeca73984180bba33c019

HonzaKral:
This is by design so that you can share the trace log. the regular log (`elasticsearch` logger) has the proper host/port.

Thanks for the report.

pataquets:
From #327:
Suggestion: [Allow] Use 'elasticsearch' fixed hostname, so people interested on using it can tinker with local DNS resolution or allow to customize the base url (http://localhost:9200) part via command line switch or config option.
Would that be acceptable? (just leaving open to possible PRs)

PR #488, opened 2016-11-11
Fix hardcoded path when logging

msherry:


HonzaKral:
This is on purpose and is actually a feature :)

Closing this as a duplicate of #380

PR #497, opened 2016-12-01
Updated example scripts for ES v5

veatch:
`example/load.py` was failing to create an index because the `multi_field` type has been removed from ES 5. Index creation was failing silently because of the use of `ignore=400`, so now TransportErrors are caught and only ignored if the error is `index_already_exists_exception`. Scripted updates are disabled by default in ES 5, so removing the `update` example seemed like the cleanest fix. `example/queries.py` has also been updated to fix breaking changes in query syntax in ES 5.

### DIFF
--- a/example/load.py
+++ b/example/load.py
@@ -159,7 +159,7 @@ REPO_ACTIONS = [
     {'_type': 'repos', '_id': 'elasticsearch', '_source': {
         'owner': {'name': 'Shay Bannon', 'email': 'kimchy@gmail.com'},
         'created_at': datetime(2010, 2, 8, 15, 22, 27),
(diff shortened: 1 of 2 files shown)


karmi:
Hi @veatch, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/497.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

veatch:
Oops! I replaced the email with filter-branch, and it looks like the CLA check has passed now. Thanks

PR #498, opened 2016-12-09
Tweaks to setup.py

pjediny:
- Use context manager for opening README
- Specify encoding of README file
- Normalize path of README file
- The unittest2 should be in tests_require and not in install_requires
- Use single quotes strings consistently everywhere
- pep8 reformat

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,16 @@
 # -*- coding: utf-8 -*-
-from os.path import join, dirname
-from setuptools import setup, find_packages
 import sys
+from io import open
+from os.path import join, dirname, normpath
+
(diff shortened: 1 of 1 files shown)


pjediny:
OK, CLA signed, how to trigger re-check?

HonzaKral:
Thank you for the PR, I am, however, unsure of what it's supposed to fix. It breaks the distribution by moving `unittest2` dependency for python 2.6 and then provides a bunch of stylistic changes that are not needed and would obfuscate the history.

I am closing this PR, please feel free to reopen with additional information as to why this is needed. Thank you!

PR #501, opened 2016-12-11
support max_raw_size_mb

dusking:


HonzaKral:
Thanks for the PR, unfortunately this is not sufficient validation since the number can be changed by configuration, it can also be different per node.

To implement this properly we would have to provide a configuration option per node to specify the max size and also retrieve the number while sniffing. Even then it would be a little brittle if it were turned on by default since it would require people to either use sniffing or always change the configuration in 2 places (elasticsearch and client config) which is trappy.

What do you think?

dusking:
Thank you for your response.

The reasoning behind the PR is an issue we had in our product.
In case we're getting a response from elastic search with raw data larger
than ~60mb, after the deseralization (json.loads) we have ~600mb in memory.
Since we've a few gunicorns, after few requests the machine memory blows
up..

Also, I didn't understand your suggestion, is it relevant to my scenario?

HonzaKral:
Oh, I apologize, I completely misread the PR the first way around and thought you were validating request size.

There is no reason to arbitrarily limit the amount of data coming back from elasticsearch and it is not in the scope of this library. Feel free to implement this in your connection class by subclassing one from the library and do your check there:
```
class MyConnection(Urllib3HttpConnection):
    def perform_request(self, *args, **kwargs):
        status, headers, data = super().perform_request(*args, **kwargs)
        if len(data) > 60*1024*1024:
            raise XXX
        return status, headers, data

es = Elasticsearch(connection_class=MyConnection)
```

Ultimately though you need to know your data and be aware of what you are trying to do, I recommend you have a look at source filtering and other options to reduce the response size, if that is a concern for you.

PR #506, opened 2016-12-18
Fix error where address is null in _get_host_info

ytanay:
When the HTTP address is null for a given node (occurs in some configurations on ES5+), filter it instead of throwing an exception (checking for `':' in x` is invalid when `x is None`)


ytanay:
@HonzaKral any idea why the CLA is not marked as being signed?

PR #510, opened 2016-12-24
Fixed issues related to apps that using App Engine as server.

soulhave:
This new connector can be used for dev's that need using elastichsearch on App Engine.

### DIFF
--- a/elasticsearch/__init__.py
+++ b/elasticsearch/__init__.py
@@ -20,6 +20,5 @@ from .connection_pool import ConnectionPool, ConnectionSelector, \
     RoundRobinSelector
 from .serializer import JSONSerializer
 from .connection import Connection, RequestsHttpConnection, \
-    Urllib3HttpConnection
+    Urllib3HttpConnection, UrlFetchAppEngine
 from .exceptions import *
-
--- a/elasticsearch/connection/__init__.py
+++ b/elasticsearch/connection/__init__.py
@@ -1,3 +1,4 @@
 from .base import Connection
 from .http_requests import RequestsHttpConnection
 from .http_urllib3 import Urllib3HttpConnection
+from .http_urlfetch import UrlFetchAppEngine
--- a/elasticsearch/connection/http_urlfetch.py
+++ b/elasticsearch/connection/http_urlfetch.py
@@ -1,0 +1,76 @@
+from .base import Connection
+from ..exceptions import ConnectionError
+from ..compat import urlencode
+import time
+import json
+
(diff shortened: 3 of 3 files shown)


soulhave:
Whats the problem ? I try submit a new pull request but got a error and i couldn't see more details about it. Someone can help me please ?

HonzaKral:
Hi, thank you for this patch!

I am, however, not sure this is in scope of this project. From all I can see (didn't try it myself, please correct me if I am wrong) `urllib3` should work fine on app engine and so do `requests` so there is no immediate need to use an alternate connection. It is also not needed to have this code inside of the library as a module and can be instead used from the outside if anybody needs it.

What do you think? What are the advantages of this approach compared to using the stock connection classes? I don't use app engine myself so would be glad for any information.

Thank you!

soulhave:
Hey @HonzaKral, thanks for answering me

So, you are right about the urllib3, it should working but not work =(. 

Before it, I did a lot of tests using urllib3 and doesn't work well using the Google App Engine (GAE) Standard. 

Actually, I'm working with a lot of python apps in the GAE Standard and I can ensure for you that urlib3 isn't a good choice for GAE in the standard version. Google recommends use UrlFetch.

Because of this, for my applications that use elasticsearch today, I use my forked version. Unfortunately, the App Engine Standard, although is an easy and fast solution, also is a sandbox and we have a little bit problems because of that. 

The documentation says a little bit about some caveats, see more in the document below:
https://media.readthedocs.org/pdf/urllib3/latest/urllib3.pdf

Before this solution, I researched some ways to fix it without impacts in my applications and involved third parts. So this new additional connection looks for me a more elegant solution. I found in the github other solutions like it, trying to solve the same problem.

HonzaKral:
There is no need to use a fork, the connection class can live outside of the code base - see for example https://github.com/Loudr/elasticsearch-urlfetch

Because that is possible I don't think we need to exclude that in the library itself.

soulhave:
@HonzaKral I understand your point ... I did the same thing before, and just for you know works fine! Thanks for the tip ..  sorry for my insistence, just for clarify my idea ... The point here is adding more this connector as a new option and not exclude anything =]

HonzaKral:
Well, we have to draw the line somewhere about what to include and where to rely on external packages. This seems niche enough use case with easy enough external solution that maintaining the connector in the library itself is not worth it, just getting proper testing would be hard etc...

I appreciate your point of view though, but in the end this is exactly why we made the library extendable...

PR #512, opened 2017-01-03
Native Google App Engine Support with URLFetch

chmoder:
I made a connection class for Google App Engine.  This class is less configuration dependent than python-requests in app engine and will hopefully save engineers time adding elasticsearch-py to their projects.

See here for more information on google app engine supported request library support:
https://cloud.google.com/appengine/docs/python/issue-requests

Thanks!

### DIFF
--- a/elasticsearch/__init__.py
+++ b/elasticsearch/__init__.py
@@ -20,6 +20,6 @@ from .connection_pool import ConnectionPool, ConnectionSelector, \
     RoundRobinSelector
 from .serializer import JSONSerializer
 from .connection import Connection, RequestsHttpConnection, \
-    Urllib3HttpConnection
+    Urllib3HttpConnection, URLFetchHttpConnection
 from .exceptions import *
 
--- a/elasticsearch/connection/__init__.py
+++ b/elasticsearch/connection/__init__.py
@@ -1,3 +1,4 @@
 from .base import Connection
 from .http_requests import RequestsHttpConnection
 from .http_urllib3 import Urllib3HttpConnection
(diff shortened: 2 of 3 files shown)


chmoder:
I have signed the CLA as of now.

HonzaKral:
Thank you for the patch, may I ask what is the motivation for an app-engine specific connection class? I was given to understand that `urllib3` should work just fine on app engine, is that not the case?

Also have you seen #510? It looks like a duplicate.

chmoder:
@HonzaKral 

Thanks for the question and showing me that duplicate.  It's pretty funny how someone else was building the same thing as me at around the same time.  Nice work on elasticsearch-py, and thanks for your interest in adding this connection class.

In the link I provided it shows the supported "request" libraries in app engine.  urllib2 is supported unfortunately urllib3 is not.

I did get python-requests and elasticsearch-py working on app engine, but requests is a new feature and I needed to import a lot of libraries and add some configuration to our project for just this one library.  

So I decided to make a urlfetch connection class.  This gives us a more reliable and automatic (no import+configuration+testing) option.  All the non-urlfetch libraries are converted to urlfetch in the end so it also has a performance improvement.  

In summary URLFetchHttpConnection saves app engine users time, reduces application complexity, improves performance, and is more supported than the proxy libraries.

HonzaKral:
I understand, unfortunately we cannot support all platforms and, fortunately, we don't need to. This can live outside of `elasticsearch-py` with no loss of functionality as demonstrated by https://github.com/Loudr/elasticsearch-urlfetch

With that in mind I will close this PR, thank you for your work!

PR #513, opened 2017-01-06
Write log with customized port number

chaopli:
When I am monitoring the debug log from elasticsearch-py, it cannot show the correct port number which I am using. I found that this debug message is hardcoded with 'http://localhost:9200', so I changed it into using the customized info from self.host.


HonzaKral:
This is for the trace logger which is hard-coded to provide standardized logs (see [0]). For logging with correct host/port use the `elasticsearch` logger.

0 - http://elasticsearch-py.readthedocs.io/en/master/#logging

PR #518, opened 2017-01-13
Bulk Index timeout - Closes #517

### LINKED ISSUE #517: Bulk Index timeout
passing `params=dict(request_timeout=100)` as
```
.bulk(..., params=dict(request_timeout=100))
```
works but when doing chunks it doesn't
```
.bulk(..., chunk_size=100, max_chunk_bytes=100000, params=dict(request_timeout=100))
```
because `params` pass to `elasticsearch.helpers.streaming_bulk` as `**kwargs` are then passed all the way down to `elasticsearch.transport.Transport#perform_request` and `kwargs` gets mutated by `timeout = params.pop('request_timeout', None)` which results in subsequents requests to be missing some parameters.

alkaros2506:
Make sure the params are not being mutated on subsequent requests
Closes #517

### DIFF
--- a/elasticsearch/helpers/__init__.py
+++ b/elasticsearch/helpers/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import copy
 import logging
 from operator import methodcaller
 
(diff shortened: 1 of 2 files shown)


karmi:
Hi @alkaros2506, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/518.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

alkaros2506:
Hi @karmi, I believe I've added both email addresses on my profile now, and signed twice the agreement, every time with the different email. Sorry about that. Looks like my config files need some cleanup

HonzaKral:
Thank you so much for the PR, I didn't end up using it but it was super helpful in debugging!

The proper fix turned out to be https://github.com/elastic/elasticsearch-py/commit/f38fa51fea4e0b23054509a8e6feb8fb486209e7#diff-4b82dd6c9c34917fa2179655e8bb0b51R62 which fixes this for every occurence and not just in the `bulk` helpers.

PR #519, opened 2017-01-13
Relocate package name

jolynch:
This moves the package from "elasticsearch" to "elasticsearch5" and fixes the test references.

All the unit tests pass run locally.


jolynch:
Actually, I don't think I should open this until we actually talk about it in the issue. Sorry about that, I'll re-open if we actually move forward with that.

PR #520, opened 2017-01-13
Fix: _cluster/nodes/hotthreads contains unrecognized parameter: doc_type

nfsec:
doc_type causes an error:

`{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"request [/_cluster/nodes/hotthreads] contains unrecognized parameter: [doc_type]"}],"type":"illegal_argument_exception","reason":"request [/_cluster/nodes/hotthreads] contains unrecognized parameter: [doc_type]"},"status":400}`

Fix tested under 5.1.1  version ( https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-hot-threads.html ).

### DIFF
--- a/elasticsearch/client/indices.py
+++ b/elasticsearch/client/indices.py
@@ -661,7 +661,7 @@ class IndicesClient(NamespacedClient):
 
     @query_params('allow_no_indices', 'expand_wildcards', 'field_data',
         'fielddata', 'fields', 'ignore_unavailable', 'query', 'recycler',
-        'request')
+        'request_cache')
     def clear_cache(self, index=None, params=None):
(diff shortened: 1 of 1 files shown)


karmi:
Hi @nfsec, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/520.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

nfsec:
Sorry. I forgot to switch git settings. E-mail added.

PR #525, opened 2017-01-21
Fix hot_threads type and clear_cache request

nfsec:
request = request_cache in code and  	_cluster/nodes/hotthreads contains unrecognized parameter: doc_type.


nfsec:
The request_cache can't be skipped: https://github.com/elastic/elasticsearch/pull/22731

PR #543, opened 2017-02-25
Fixed issue#542

natureCode:
This fix can make indices.exists_type works also for Elasticsearch2.x


HonzaKral:
I am sorry, this is not how the exists type API works in `5.0` - https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-types-exists.html

PR #550, opened 2017-03-12
wait for completion is on by default

allias888:
By default calls to _delete_by_query and _reindex return some result different from task result.

Only when we explicitly whait_for_completion=false task id is returned

https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html


HonzaKral:
Thanks for the patch!

Issue #560, opened 2017-03-24
update from 5.0.1 to 5.2


gabriel-tessier:
Hi,
I freshly upgrade from 5.0.1 to 5.2 and I have the following error:

> File "/home/src/lib/elasticsearch/client/utils.py", line 62, in _wrapped
>     params = kwargs.pop('params').copy()
> AttributeError: 'NoneType' object has no attribute 'copy'
> 

Somehow params = None and it crash, on 5.0.1 I didn't have this problem.

Thanks

fatelei:
in 5.0.1 the related code is:

![image](https://cloud.githubusercontent.com/assets/961094/24537017/b7335a5a-1611-11e7-9dac-071cbaccd500.png)

HonzaKral:
I don't see how this can happen unless you explicitly pass `params=None` in your code to any of the APIs which is something the library cannot reasonably deal with and shouldn't have a special code for imho.

I am closing this issue for now, please feel free to reopen with a sample of the code that is causing this exception to trigger, thank you!

PR #563, opened 2017-03-31
Fix issures 560

fatelei:
If params in kwargs and params is None, kwargs.get('params') is `False`
If params not kwargs, kwargs.get('params') is `False`

### DIFF
--- a/elasticsearch/client/utils.py
+++ b/elasticsearch/client/utils.py
@@ -58,7 +58,7 @@ def query_params(*es_query_params):
         @wraps(func)
         def _wrapped(*args, **kwargs):
             params = {}
-            if 'params' in kwargs:
+            if kwargs.get('params'):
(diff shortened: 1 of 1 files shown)


karmi:
Hi @fatelei, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/563.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

fatelei:
https://github.com/elastic/elasticsearch-py/issues/560

fatelei:
@karmi I have added missing email address

HonzaKral:
My question is how would `params` be `None`? For that to happen someone would explicitly have to pass it into an api method which is not supported or documented. Do you have a use case you are trying to solve?

Thank you!

PR #569, opened 2017-04-13
Unquote user:password in hosts

### LINKED ISSUE #568: user:password in hosts are not url decoded
Hello,

If I have a character like "]" in my username or password I must urlencode it or `urlparse` will interpret the host as IPv6. Unfortunately, `urlparse` does not urldecode fragments : 
```
The components are not broken up in smaller parts (for example, the network location is a single string), and % escapes are not expanded.
```
(https://docs.python.org/3/library/urllib.parse.html)

`_normalize_hosts` should urldecode explicitely, which I will propose in a PR.

pmezard:
Before turning them into an authentication header, the user or password
supplied in hosts must be URL decoded.

The RFC https://tools.ietf.org/html/rfc3986 describes userinfo like:

  userinfo    = *( unreserved / pct-encoded / sub-delims / ":" )

and python documentation about urlparse says:

  The components are not broken up in smaller parts (for example, the
  network location is a single string), and % escapes are not expanded.

It is the caller job to unquote them before using them.

For instance, a character like "]" cannot be supplied in a host or
urllib will incorrectly interpret it using ipv6 syntax.

Fix #568

### DIFF
--- a/elasticsearch/client/__init__.py
+++ b/elasticsearch/client/__init__.py
@@ -3,7 +3,7 @@ import logging
 
 from ..transport import Transport
 from ..exceptions import TransportError
-from ..compat import string_types, urlparse
+from ..compat import string_types, urlparse, unquote
 from .indices import IndicesClient
 from .ingest import IngestClient
 from .cluster import ClusterClient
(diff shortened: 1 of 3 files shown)


HonzaKral:
Thanks for the patch!

Without this PR you could always just have passed in the `http_auth` parameter unquoted explicitly which would bypass the parsing: `Elasticsearch(["localhost"], http_auth=('user', 'secre]'))`

pmezard:
Thanks @HonzaKral. FWIW, the reason we are using the (deprecated) URL userinfo is we are migrating our codebase from base ES to ES+Shield and the ability to inject credentials without touching ES client constructors saves a lot of time.

HonzaKral:
I totally understand, just wanted to leave the structured solution here for people who don't want to do the encode and/or wait for the next release.

Issue #574, opened 2017-04-20
RPM name conflict


0r1t0m:
I create a RPM file according the Python [manual](https://docs.python.org/2/distutils/builtdist.html#creating-rpm-packages): 
```
> python setup.py bdist --formats=rpm
```
This command generate a RPM specification file which contains the same name like the 'real' Elasticsarch [RPM](https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.3.0.rpm) and cause a conflict when you install both RPM files.

```
> rpmbuild -bb elasticsearch.spec
> rpm -qip elasticsearch-5.3.0-1.noarch.rpm  | head
Name        : elasticsearch
Version     : 5.3.0
Release     : 1
Architecture: noarch
```

**Solution**
- Change the name in setup.py to 'python-elasticsearch'

HonzaKral:
I am sorry, but we cannot rename the package. What I would suggest to you is to create the `spec` file for yourself or temporarily change the `name` in `setup.py` when calling the `bdist_rpm`.

0r1t0m:
Okay, just for curiosity: why the package name can not be changed?

HonzaKral:
It's ok to change the rpm name (by producing your own spec file), but not the name of the python package, which this change would - it would break everyone's install script, dependencies etc.

untergeek:
@0r1t0m Use fpm to build your RPMs, and you can re-name the the RPM however you like.

0r1t0m:
@HonzaKral: Thanks for clarification
@untergeek: Thanks for the hint

PR #575, opened 2017-04-20
CHG: project name in setup script

0r1t0m:
FIX Issue #574


HonzaKral:
unfortunately we cannot change the name of the package, it would break installation for many people.

PR #590, opened 2017-05-15
Add script to run ES in docker container

fxdgear:
* Adding start_elasticsearch.sh script to run elastic search on the host
in a docker container at http://localhost:9200

* update readme in instructions on how to use script


start_elasticsearch.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Start elasticsearch in a docker container
+
+[ -z "$ES_VERSION" ] && echo "Need to set ES_VERSION (2.4, 5.3, 5.4)" && exit 1;
```
HonzaKral: could we default to `5.4.0` for now? or, better yet, is there `latest` or `5.x` ?

fxdgear: > could we default to `5.4.0` for now?

Yes can make that the default

> better yet, is there `latest` or `5.x`

Not yet. but all I need to do is `docker tag fxdgear/elasticsearch:5.4 fxdgear/elasticsearch:latest` and then push and there will be a `latest`. 

and then I'll update the default to be `latest`

HonzaKral:
Thanks for the patch @fxdgear!

PR #594, opened 2017-05-20
404 - Page Not Found for Tasks Management API

nfsec:
Task Management API moved to single page.


HonzaKral:
thanks for the patch!

PR #597, opened 2017-05-25
add support for POST method in search

azban:
From https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html: "Both HTTP GET and HTTP POST can be used to execute search with body. Since not all clients support GET with body, POST is allowed as well."


HonzaKral:
If you are in an environment where `GET` with body is not supported please use the `send_get_body_as` parameter and set it to `'POST'` as per (0). That is a mre systematic solution that will work globally and won't rely on every individual call to be annotated with the desired method.

Hope this helps.

0 - http://elasticsearch-py.readthedocs.io/en/master/#environment-considerations

azban:
ack... thank you

PR #601, opened 2017-06-14
add optional `host` and `path` param on load.py

fxdgear:
If you want/need to load data into a different es cluster just specify either the: 
 * -H/--host option
 * -p/--path option

ex: `python load.py --host http://123.123.123.123:9200 -p /home/code/elasticsearch`

### DIFF
--- a/example/load.py
+++ b/example/load.py
@@ -6,6 +6,7 @@ from os.path import dirname, basename, abspath
 from datetime import datetime
 import logging
 import sys
+import argparse
 
 import git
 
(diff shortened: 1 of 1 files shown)


HonzaKral:
Nice, when you are adding `argparse` it would be nice to also have it handle the part to the git repo - https://github.com/elastic/elasticsearch-py/blob/master/example/load.py#L181

fxdgear:
@HonzaKral bump

PR #605, opened 2017-06-23
Update README

Decator:
Add a documentation link.


HonzaKral:
Thanks for the patch but I don't think we need to provide links for the individual subsections. The link you mentioned also points just to the `indices` namespace which micght be confusing

PR #613, opened 2017-07-02
Add Python 3.5 and 3.6 to tox.ini

jonafato:
Bring `tox.ini` up to date with `setup.py` and `.travis.yml`.


HonzaKral:
Thanks for the patch!

PR #625, opened 2017-07-25
Bulk Update Usage Clarification

funseiki:
Bulk update appears to require 'doc' to be wrapped inside of a '_source'. The documentation update reflects that requirement.


HonzaKral:
Hi, it actually doesn't require the wrapping in `_source`, if `_source` is not present we will just use the leftover keys in the top level dict once we have removed all the metadata options.

we also test this exact form: https://github.com/elastic/elasticsearch-py/blob/master/test_elasticsearch/test_server/test_helpers.py#L61

This form also makes more sense for the `update` operation since it mimics the actual update API

PR #628, opened 2017-08-07
better handling of bytestrings in _escape

### LINKED ISSUE #627: 'bytes' object has no attribute 'encode'
File "/root/.local/share/virtualenvs/celery-jR7fJ1bi/lib/python3.5/site-packages/elasticsearch/client/utils.py", line 32, in _escape
        return value.encode('utf-8')
    AttributeError: 'bytes' object has no attribute 'encode'

It appears that when _escape receives value, which in my case is b'celery-task-meta-c4f1201f-eb7b-41d5-9318-a75a8cfbdaa0', it gets checked against (<class 'str'>, <class 'bytes'>), and therefore passes the test but <class 'bytes'> does not have the **encode** method.

try does not catch as error is AttributeError: 'bytes' object has no attribute 'encode'

    def _escape(value):
        """
        Escape a single value of a URL string or a query parameter. If it is a list
        or tuple, turn it into a comma-separated string first.
        """

        # make sequences into comma-separated stings
        if isinstance(value, (list, tuple)):
            value = ','.join(value)

        # dates and datetimes into isoformat
        elif isinstance(value, (date, datetime)):
            value = value.isoformat()

        # make bools into true/false strings
        elif isinstance(value, bool):
            value = str(value).lower()

        # encode strings to utf-8
        if isinstance(value, string_types):
            try:
                return value.encode('utf-8')
            except UnicodeDecodeError:
                # Python 2 and str, no need to re-encode
                pass

    return str(value)

fxdgear:
Fixes #627

### DIFF
--- a/elasticsearch/client/utils.py
+++ b/elasticsearch/client/utils.py
@@ -26,13 +26,13 @@ def _escape(value):
     elif isinstance(value, bool):
         value = str(value).lower()
 
+    # don't decode bytestrings
+    elif isinstance(value, bytes):
+        return value
+
     # encode strings to utf-8
     if isinstance(value, string_types):
-        try:
-            return value.encode('utf-8')
-        except UnicodeDecodeError:
-            # Python 2 and str, no need to re-encode
-            pass
+        return value.encode('utf-8')
 
     return str(value)
 
--- a/test_elasticsearch/test_client/test_utils.py
+++ b/test_elasticsearch/test_client/test_utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from elasticsearch.client.utils import _make_path
+from elasticsearch.client.utils import _make_path, _escape
 from elasticsearch.compat import PY2
 
 from ..test_cases import TestCase, SkipTest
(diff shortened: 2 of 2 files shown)


elasticsearch/client/utils.py
```diff
@@ -26,6 +26,10 @@ def _escape(value):
     elif isinstance(value, bool):
         value = str(value).lower()
 
+    # decode bytestrings
+    elif isinstance(value, bytes):
+        value = value.decode()
```
HonzaKral: if it's `bytes` then we don't want to `decode` it, just pass it through to avoid any problems with en/decoding and the extra work.

test_elasticsearch/test_client/test_utils.py
```diff
@@ -17,3 +17,24 @@ def test_handles_utf_encoded_string(self):
...
+            _escape(string)
+        )
+    def test_handles_unicode(self):
+        string = "中文"
+        self.assertEquals(
+            b'\xe4\xb8\xad\xe6\x96\x87',
+            _escape(string)
+        )
+
+    def test_handles_bytestring(self):
```
HonzaKral: note that this test passes on python 2 even without the code change.

fxdgear: yeah I was running this in 2 different envs (python2 and python3) to make sure it worked in both. I just wanted to have an explicit test case for handling bytestring.

PR #629, opened 2017-08-08
Additional fixes for UTF-8 surrogate escapes

tylerjharden:
My previously merged PR did not handle some additional spots in the code where similar encoding is done, namely success logging.


karmi:
Hi @tylerjharden, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/629.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

tylerjharden:
@karmi I believe your bot is mistaken. The e-mails match, but I added my work e-mail just to be safe.

HonzaKral:
Thanks for the patch!

PR #635, opened 2017-08-18
Nick/ssl context

fxdgear:
use SSLContext when using https

### DIFF
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -1,4 +1,5 @@
 import time
+import ssl
 import urllib3
 from urllib3.exceptions import ReadTimeoutError, SSLError as UrllibSSLError
 import warnings
@@ -14,6 +15,19 @@ except ImportError:
 from .base import Connection
 from ..exceptions import ConnectionError, ImproperlyConfigured, ConnectionTimeout, SSLError
 from ..compat import urlencode
+
+
+def create_ssl_context(**kwargs):
+    """
+    A helper function around creating an SSL context
+
+    https://docs.python.org/3/library/ssl.html#context-creation
+
+    Accepts kwargs in the same manner as `create_default_context`.
+    """
+    ctx = ssl.create_default_context(**kwargs)
+    return ctx
+
 
 class Urllib3HttpConnection(Connection):
     """
@@ -47,7 +61,7 @@ class Urllib3HttpConnection(Connection):
     def __init__(self, host='localhost', port=9200, http_auth=None,
             use_ssl=False, verify_certs=True, ca_certs=None, client_cert=None,
             client_key=None, ssl_version=None, ssl_assert_hostname=None,
-            ssl_assert_fingerprint=None, maxsize=10, headers=None, **kwargs):
+            ssl_assert_fingerprint=None, maxsize=10, headers=None, ssl_context=None, **kwargs):
 
         super(Urllib3HttpConnection, self).__init__(host=host, port=port, use_ssl=use_ssl, **kwargs)
         self.headers = urllib3.make_headers(keep_alive=True)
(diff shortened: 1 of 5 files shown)


elasticsearch/connection/http_urllib3.py
```diff
@@ -62,33 +76,37 @@ def __init__(self, host='localhost', port=9200, http_auth=None,
                 self.headers[k.lower()] = headers[k]
 
         self.headers.setdefault('content-type', 'application/json')
-        ca_certs = CA_CERTS if ca_certs is None else ca_certs
         pool_class = urllib3.HTTPConnectionPool
         kw = {}
         if use_ssl:
+            if not ca_certs and not ssl_context and verify_certs:
```
HonzaKral: `ca_certs` have a default value so it's ok if it's empty. This could work if we haven't removed line 65

elasticsearch/connection/http_urllib3.py
```diff
@@ -62,33 +76,37 @@ def __init__(self, host='localhost', port=9200, http_auth=None,
...
+                warnings.warn(
+                    'Connecting to %s using SSL with verify_certs=False is insecure.' % host)
+
             kw.update({
                 'ssl_version': ssl_version,
                 'assert_hostname': ssl_assert_hostname,
                 'assert_fingerprint': ssl_assert_fingerprint,
+                'ssl_context': ssl_context,
+                'cert_file': client_cert,
+                'key_file': client_key,
```
HonzaKral: Shouldn't `cert_file` and `key_file` also be part of `ssl_context`? seems inconsistent to pass some values directly and some via ssl context...

elasticsearch/connection/http_urllib3.py
fxdgear: `ca_certs` have a default value IFF `certifi` is installed.

elasticsearch/connection/http_urllib3.py
fxdgear: I couldn't find any documentation about client cert/key being part of SSLContext. My understanding is that SSLContext is for handling CA stuff. 

The `cert_file` and `key_file` are keys and certs for the client to do key based auth as opposed to http_auth, right?

elasticsearch/connection/http_urllib3.py
HonzaKral: yes, but that is in the code after this `if` so here it can still be empty since you removed the application of the default from line `65` - https://github.com/elastic/elasticsearch-py/pull/635/files#diff-e03df718a08bdad9d78d94c50dd7db90L65

GlenRSmith:
```======================================================================
ERROR: test_children_are_reindexed_correctly (test_elasticsearch.test_server.test_helpers.TestParentChildReindex)
----------------------------------------------------------------------
  File "/home/travis/build/elastic/elasticsearch-py/test_elasticsearch/test_server/test_helpers.py", line 305, in setUp
    self.client.indices.create(index='test-index', body=body)
  File "/home/travis/build/elastic/elasticsearch-py/elasticsearch/client/utils.py", line 73, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/home/travis/build/elastic/elasticsearch-py/elasticsearch/client/indices.py", line 91, in create
    params=params, body=body)
  File "/home/travis/build/elastic/elasticsearch-py/elasticsearch/transport.py", line 312, in perform_request
    status, headers, data = connection.perform_request(method, url, params, body, ignore=ignore, timeout=timeout)
  File "/home/travis/build/elastic/elasticsearch-py/elasticsearch/connection/http_requests.py", line 90, in perform_request
    self._raise_error(response.status_code, raw_data)
  File "/home/travis/build/elastic/elasticsearch-py/elasticsearch/connection/base.py", line 125, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(status_code, error_message, additional_info)
elasticsearch.exceptions.RequestError: TransportError(400, 'mapper_parsing_exception', 'No handler for type [join] declared on field [question_answer]')
```

fxdgear:
@GlenRSmith turns out this PR is breaking because of this commit: https://github.com/elastic/elasticsearch-py/commit/4e6f63571105545914c07fa4846f996d6049f44b

fxdgear:
@HonzaKral is this looking any better??

elasticsearch/connection/http_urllib3.py
zmathe: @fxdgear @HonzaKral is right. If you have ssl_context you do not need cert_file and key_file. But it is depend on the python version.
If someone using python 2.7.8 ( ssl_context is introduced in 2.7.9!), can not use ssl_context. So, cert_file and key_file is used.
For example:
python 2.7.8:
```python
import ssl 
ssl.wrap_socket(socket, keyfile, certfile, ca_certs, cert_reqs=ssl.CERT_REQUIRED)
```
python 2.7.9:
```python
import ssl
context = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
#load the certificates
....
context.wrap_socket(socket, serverside=False)
```
I propose to remove cert_file and key_file, because it is inconsistent and up to the developer to create the ssl_context. We are moved to python 2.7.13 this solution is okay for me. But when I opened #471 is was not the case!

elasticsearch/connection/http_urllib3.py
```diff
@@ -62,33 +76,44 @@ def __init__(self, host='localhost', port=9200, http_auth=None,
                 self.headers[k.lower()] = headers[k]
 
         self.headers.setdefault('content-type', 'application/json')
-        ca_certs = CA_CERTS if ca_certs is None else ca_certs
         pool_class = urllib3.HTTPConnectionPool
         kw = {}
-        if use_ssl:
+
+        # if providing an SSL context, raise error if any other SSL related flag is used
+        if ssl_context and (use_ssl or verify_certs or ca_certs or ssl_version):
```
zmathe: If ssl_context is True, will not work, because use_ssl is true if https protocol is used.
This line will overwrite use_ssl to False or True depending on the protocol: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/transport.py#L153, because of _normalize_hosts (https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/__init__.py#L170)
More details: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/__init__.py#L19

Could you please remove use_ssl from the check, because when https is used use_ssl always true?

fxdgear: Thanks for the feedback @zmathe

fxdgear:
@zmathe working on some travis issues :( goal is soon. I appreciate your feedback on the PR.

elasticsearch/connection/http_urllib3.py
zmathe: @fxdgear Could you please also remove use_ssl from (use_ssl or verify_certs or ca_certs or ssl_version) tuple? Because if the protocol is https, use_ssl is True and can not set to False
Thanks again for making this available.

fxdgear:
@zmathe I hadn't forgotten. was working on some other conditionals for skipping some broken tests. just added it now

PR #644, opened 2017-08-29
Add requirements.txt in the example

iblancasa:
I think it is more natural to install the Python packages from a requirements.txt file.

Thank you for your time.


HonzaKral:
Thank you for the patch, could I please ask you to not pin the *exact* version of the dependencies though? Just a major version should be sufficient as otherwise it might lead to a higher chance of conflict and/or us missing an important minor (bugfix) update.

iblancasa:
Sure! I fixed it.

I thought that it was a good idea to be totally sure that the example is going to work.

Thank you @HonzaKral.

HonzaKral:
thanks!

PR #652, opened 2017-09-28
fix: AttributeError: 'bytes' object has no attribute 'encode'(PY3)

brunobell:
Hi, I had this error in python3.6.2 while using elasticsearch as backend of celery. So I tracked down the error message and come to a simple fix for this. It's working now in my local python3.6.2 virtualenv, and I guess this should work for all. I'd like to hear your comments :)

### DIFF
--- a/elasticsearch/client/utils.py
+++ b/elasticsearch/client/utils.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals
 import weakref
 from datetime import date, datetime
 from functools import wraps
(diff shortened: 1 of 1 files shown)


fxdgear:
Can you please rebase off master and try again. I had to upgrade the version of ES Travis uses.

brunobell:
@fxdgear Hi, I've done rebase and passed local tests with python version 2.7.14 and 3.6.2, as well as travis ci test. Glad to know if you'd like to merge this.

PR #654, opened 2017-10-11
Add `bytes` param to cat.shards and cat.segments

GlenRSmith:


HonzaKral:
Thanks for the patch, @GlenRSmith !

Issue #615, opened 2017-07-07
Use autogenerating IDs for documents


soopanova:
In the Elastic API they state that you can create a document in an index with automatic id generation. Can this be replicated in the module. Currently you must provide an ID when creating a new document.

HonzaKral:
You don't have to provide an `id` when creating a new document:
```
from elasticsearch import Elasticsearch
es = Elasticsearch()
es.index(index='i', doc_type='t', body={})
```

soopanova:
This is not true for the es.create. If no id is given an exception is thrown. Is index the preferred method of document creation?

HonzaKral:
That is true - see my answer in https://github.com/elastic/elasticsearch-py/issues/516

Thanks!

Issue #516, opened 2017-01-10
Need an ID to create a document


alanbacon:
in version 5 it appears that a document ID must be specified to create a document. If one is not specified the following error appears:

    TypeError: create() missing 1 required positional argument: 'id'

This is nonsensical. If I do not specify an ID elasticsearch should create one for me.

HonzaKral:
`create` API is nonsensical without the ID, if you don't have an ID you don't need the `create` API as you can just call `index` API.

`create` API exists only to **force** creation (and fail if document already exists).

Duplicate of #474

alanbacon:
Fair enough. Thanks

tigerneil:
so use `index` instead? 

BTW `create` can assign a customized id :)

swethmandava:
When I try to create with an id, I get:
`TypeError: create() got an unexpected keyword argument 'id'`
is it customizing the id by default then?

When I try to check if index exists, it requires an id: I'm not sure how I can check if index exists if I can't declare an id myself. Any thoughts?

Issue #474, opened 2016-10-28
Creating document with automatic ID generation doesn't work in 5.0


piquadrat:
It seems to me that a2cd22dbf6 removed the possibility to create a document using [automatic ID generation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#_automatic_id_generation). Is that an oversight?

If yes, I'll be happy to provide a PR that restores this functionality, by changing `id` back to a keyword argument, and using `POST` if it is `None`.

HonzaKral:
If you are not providing an id there is no benefit in using the `create` API and you should instead be using the `index` API which still provides the functionality.

The only benefit of the `create` API is if you have specified an ID the operation will fail if a document with that ID already exists. This is not applicable without an ID value so the parameter has been switched to required.

Hope this helps

piquadrat:
Ah, that makes sense! Sorry for the noise!

PR #655, opened 2017-10-16
generate id automatically use POST method

bts-webber:
Hi,
   I modified Elasticsearch().create() to allow it to automatically generate document id using the "POST" method.We can find this fearture in doc:https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html

### DIFF
--- a/elasticsearch/__init__.py
+++ b/elasticsearch/__init__.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import
 VERSION = (5, 4, 0)
 __version__ = VERSION
 __versionstr__ = '.'.join(map(str, VERSION))
-test
+
 import sys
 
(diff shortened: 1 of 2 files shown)


karmi:
Hi @bts-webber, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/655.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

bts-webber:
Hi @karmi ,done.

HonzaKral:
If you want to auto-generate the ID, please use the `index` API, using `create` API for that case is not supported. See #615, #516 and #474 for more details.
Thanks!

PR #656, opened 2017-10-16
Avoid redefining builting id function

iblancasa:
Previously, the [builtin function id()](https://docs.python.org/2/library/functions.html#id) was redefined.

Thank you!

### DIFF
--- a/elasticsearch/client/ingest.py
+++ b/elasticsearch/client/ingest.py
@@ -46,16 +46,16 @@ class IngestClient(NamespacedClient):
             'pipeline', id), params=params)
 
     @query_params('verbose')
-    def simulate(self, body, id=None, params=None):
+    def simulate(self, body, pipelineID=None, params=None):
         """
         `<https://www.elastic.co/guide/en/elasticsearch/plugins/current/ingest.html>`_
 
         :arg body: The simulate definition
-        :arg id: Pipeline ID
+        :arg pipelineID: Pipeline ID
         :arg verbose: Verbose mode. Display data output for each processor in
             executed pipeline, default False
         """
         if body in SKIP_IN_PATH:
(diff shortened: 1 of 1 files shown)


elasticsearch/client/ingest.py
```diff
@@ -46,16 +46,16 @@ def delete_pipeline(self, id, params=None):
             'pipeline', id), params=params)
 
     @query_params('verbose')
-    def simulate(self, body, id=None, params=None):
+    def simulate(self, body, pipelineID=None, params=None):
```
georgepsarakis: Usually in Python in order not to shadow builtins you can append an underscore: `id_`.
If you want to use this name, I think you should change it to `pipeline_id`.

elasticsearch/client/ingest.py
```diff
@@ -46,16 +46,16 @@ def delete_pipeline(self, id, params=None):
             'pipeline', id), params=params)
```
georgepsarakis: I guess the same should change should be applied in this method as well?

iblancasa: Oh! Yes. I will change this ASAP. Thanks for your feedback!

HonzaKral:
Thank you for the patch, @iblancasa, but I don't believe this change is worth breaking the API for some people. The `id` function is not being redefined except inside the API method where it is not needed so no actual harm is done.

In the beginning, when designing this library, we decided to rename the variable `type` to `doc_type` and leave the `id` as is. The motivation for this was that we felt that `type` is more widely used and people would adopt it as variable name in their code, possibly leading to some issues. With `id` this is less of a problem and also widely accepted throughout python codebases.

I know this is not ideal, but right now making the change would introduce more harm and confusion. Thank you for understanding!

PR #657, opened 2017-10-19
Drop support for EOL Python 2.6, 3.2 and 3.3

hugovk:
https://github.com/elastic/elasticsearch-py/pull/548#issuecomment-291575889:

> `2.6` and `3.3` should be dropped. They're obsolete now and nobody on those versions is going to be upgrading to new versions of this library.

Here's the pip installs for elasticsearch from PyPI for the last month, showing low usage of the old ones:
```
$ pypinfo --percent --pip elasticsearch pyversion
python_version percent download_count
-------------- ------- --------------
2.7              73.0%        550,247
3.6              12.0%         90,671
3.5               8.6%         65,031
3.4               4.9%         36,811
2.6               1.4%         10,329
3.3               0.0%            299
3.7               0.0%             64
3.2               0.0%             11
None              0.0%              5

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 
 python:
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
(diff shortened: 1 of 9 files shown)


HonzaKral:
Thank you for the PR! For now though I will close it since there is no overhead with managing the different python versions and there are still people stuck on older versions of python. If we ever want to develop new features that would be impossible or harder on those version I will definitely revisit the idea but for now it is not worth it even if only few people are using the library with python `2.6`.

Thanks!

PR #658, opened 2017-10-24
Dump as minified JSON as possible in its serializer

illia-v:


HonzaKral:
Thank you for the patch, @illia-v!

PR #676, opened 2017-12-04
fix typo in comment (minor)

marcocova:
Fix a typo in a comment.


karmi:
Hi @marcocova, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/676.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #618, opened 2017-07-13
Set application/x-ndjson content type on bulk requests.

jmcarp:
Resolves #609

PR #682, opened 2017-12-13
Use provided request headers in urllib3

danielmitterdorfer:
#618 provided the ability to pass custom http headers.  However, 
`Urllib3HttpConnection` did not use these headers but only its 
default ones.

With this commit, `Urllib3HttpConnection` will either use the default
ones or merge the default headers with the provided ones if there are
any.

Relates #618

### DIFF
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -142,8 +142,7 @@ class Urllib3HttpConnection(Connection):
                 method = method.encode('utf-8')
 
             request_headers = dict(self.headers)
-            if headers:
-                request_headers.update(headers or {})
+            request_headers.update(headers or {})
             response = self.pool.urlopen(method, url, body, retries=False, headers=request_headers, **kw)
             duration = time.time() - start
             raw_data = response.data.decode('utf-8')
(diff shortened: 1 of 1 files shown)


elasticsearch/connection/http_urllib3.py
```diff
@@ -141,10 +141,10 @@ def perform_request(self, method, url, params=None, body=None, timeout=None, ign
             if not isinstance(method, str):
                 method = method.encode('utf-8')
 
+            request_headers = dict(self.headers)
             if headers:
```
georgepsarakis: The `if` block seems redundant now, since the condition that custom headers have been passed is applied in the `update` function call. What do you think?

danielmitterdorfer: Good point. I pushed another commit to address this.

synhershko:
+1 for this, we just hit this ourselves. This is a quite severe oversight.

We used the following fix though, which also optimizes for the path where no custom headers are sent:

```
            if headers:
                request_headers = dict(self.headers)
                request_headers.update(headers)
            else:
                request_headers = self.headers
```

HonzaKral:
merged as a023bddb4dc572bed41433d211e0e65249ac4cc3, thanks for the patch, @danielmitterdorfer !

PR #686, opened 2017-12-23
It's possible to send content-type to _raise_error

fbacchella:
Not all error return json content (think about a broken proxy). This requests add an optional content_type argument to _raise_error , that is checked to avoid calling json.loads() on non-json content.
It default to application/json to avoid breaking existing code.

### DIFF
--- a/elasticsearch/connection/base.py
+++ b/elasticsearch/connection/base.py
@@ -109,12 +109,12 @@ class Connection(object):
         if response is not None:
             logger.debug('< %s', response)
 
-    def _raise_error(self, status_code, raw_data):
+    def _raise_error(self, status_code, raw_data, content_type='application/json'):
         """ Locate appropriate exception and raise it. """
         error_message = raw_data
         additional_info = None
         try:
-            if raw_data:
+            if raw_data and content_type == 'application/json':
                 additional_info = json.loads(raw_data)
                 error_message = additional_info.get('error', error_message)
                 if isinstance(error_message, dict) and 'type' in error_message:


fbacchella:
CLA signed.

HonzaKral:
Thank you for the suggestion! If a proxy is broken then the existing code will still work by catching the `ValueError` and providing information about the response being undecodeable, so I am wondering if the 3 µs speedup (measured on my laptop, not scientific at all) is worth the extra parameter. Is that really an issue for you?

Thanks!

fbacchella:
It’s  not a critical issue, but it make debuging error harder.

HonzaKral:
Thanks for the quick reaction! How does it make the debugging harder, if I may ask? With the patch you still get the same amount of information in the logs just with the additional `warning` about the undecodeable data in the logs. The actual exception remains unchanged. Is that the only concern or is there something I am missing?

fbacchella:
With my use case, I have a SSO system in front of Elastic. It can return an HTTP 200, with HTML content. So the current code will complain about json decoding, that's false indication and give harder to read error messages, with layered stacks.

HonzaKral:
Thank you for the explanation. If a 200 is returned, it will not be sent into this method anyway since this one is only used for failed requests so it won't help anyways. It also sounds like something that ought to be implemented on a different layer - either as your custom `Connection` class or, maybe, as a plugin for `requests` similar to what `requests-aws4auth` is doing (0), which you can then use in the client by passing it in in a similar fashion (1)

Either way I am afraid this change doesn't fit into the library itself.

0 - https://pypi.python.org/pypi/requests-aws4auth
1 - http://elasticsearch-py.readthedocs.io/en/master/#running-on-aws-with-iam

PR #687, opened 2017-12-25
A connection should be marked live with any non 500 error code and with HEAD too.

fbacchella:


HonzaKral:
The connection is considered live as long as it's getting a response from the server, if the connection itself was successfully established. That includes `5xx` responses.

I do agree that we should also consider `HEAD` requests but we should not be checking the http status. Do you have any special reason for wanting to check that?

Thanks!

fbacchella:
Nop, I just wanted to keep the logic I found but just improved it. There is argument both for and against using the status code.

fbacchella:
Think about a apache used as a reverse proxy on the same host. One might want one to do that for authentication without x-pack. If ES is down on that host, you'll get a 500 then.

HonzaKral:
If you are using a reverse proxy on the same host then it should take care of handling the node going down and fallback to other nodes in the cluster. Also at that point you should only configure `elasticsearch-py` to talk to that proxy only which means the `mark_live`/`mark_dead` become no-op since those are only used if multiple nodes are configured.

Either way this change introduces behavior that is not needed so I am closing this PR.

PR #688, opened 2017-12-27
A connection should be marked live with HEAD too.

fbacchella:


HonzaKral:
Thank you for the fix @fbacchella, much appreciated!

Issue #669, opened 2017-11-14
"elasticsearch.exceptions.ImproperlyConfigured: Root certificates are missing..." even with certifi installed


anfreita:
Creating an Elasticsearch client with `use_ssl=True` and `verify_certs=True` fails with `elasticsearch.exceptions.ImproperlyConfigured: Root certificates are missing for certificate validation. Either pass them in using the ca_certs parameter or install certifi to use it automatically.`

running this in python 2.7 with certifi installed:
```
$ python --version                                                                                                                                                                                              11:39:54
Python 2.7.12+
$ pip list                                                                                                                                                                                                      11:38:52
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
backports.shutil-get-terminal-size (1.0.0)
bcrypt (3.1.3)
boto3 (1.4.7)
botocore (1.7.43)
certifi (2017.11.5)
cffi (1.10.0)
chardet (3.0.4)
decorator (4.1.2)
Django (1.11.6)
django-crispy-forms (1.6.1)
docutils (0.14)
elasticsearch (6.0.0)
enum34 (1.1.6)
falcon (1.2.0)
futures (3.1.1)
idna (2.6)
ipython (5.5.0)
ipython-genutils (0.2.0)
jmespath (0.9.3)
lxml (3.8.0)
numpy (1.13.3)
olefile (0.44)
pandas (0.21.0)
pathlib2 (2.3.0)
pexpect (4.3.0)
pickleshare (0.7.4)
Pillow (4.2.1)
pip (9.0.1)
pkg-resources (0.0.0)
prompt-toolkit (1.0.15)
ptyprocess (0.5.2)
pycparser (2.18)
Pygments (2.2.0)
python-dateutil (2.6.1)
python-mimeparse (1.6.0)
pytz (2017.3)
requests (2.18.4)
requests-aws4auth (0.9)
requests-futures (0.9.7)
s3transfer (0.1.11)
scandir (1.6)
setuptools (36.0.1)
simplegeneric (0.8.1)
six (1.11.0)
SkySerpent (0.1.55)
traitlets (4.3.2)
trollius (2.1)
urllib3 (1.22)
wcwidth (0.1.7)
webkit2png (0.8.2)
wheel (0.30.0a0)
```

[...]

PR #689, opened 2017-12-28
Updates to SSL Context changes

fxdgear:
Bug with OSX made breaking changes to the way certifi was being imported
and used.

Unable to reproduce with Linux.

Found improper logic. Fixed now.
Also changing removing the flag for checking `verify_certs` as
it is `True` by default. Don't want to make users flag that as
`False` just to make things work.

A connection to an SSL instance should be:

    es = Elasticsearch("https://host:port", http_auth=(user,pass))

* If using Elastic Cloud or some other publicly recognized certificate
  that certifi can recognize.

### DIFF
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -80,13 +80,14 @@ class Urllib3HttpConnection(Connection):
         kw = {}
 
         # if providing an SSL context, raise error if any other SSL related flag is used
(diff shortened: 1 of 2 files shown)


fxdgear:
@HonzaKral Please review. I was able to verify https://github.com/elastic/elasticsearch-py/issues/669

This fixes the issue.

HonzaKral:
LGTM

PR #690, opened 2017-12-28
Updating Docs to reference SSL Context

fxdgear:


fxdgear:
@HonzaKral how does this look?

HonzaKral:
It would be nice to have the helper function exposed in `elasticsearch.connection` directly and also add some information (mainly that it is python standard and you can just pass it in) and links to ssl context docs.

Otherwise lgtm

fxdgear:
@HonzaKral we already link to the SSL docs https://github.com/fxdgear/elasticsearch-py/blob/415b3e7dc75013ab1247e2788bf379e58190eea0/docs/connection.rst#urllib3httpconnection-default-connection_class

I'm just adding more

fxdgear:
> It would be nice to have the helper function exposed in elasticsearch.connection directly 

so move the `create_ssl_conext` out of `http_urllib3` and into `connection`?

HonzaKral:
I think it makes sense. We could also just import it from there, up to you...

fxdgear:
@HonzaKral ok I updated the import. are we happy? :smile:

PR #693, opened 2017-12-31
ImproperlyConfigured is imported but not used in transport.

fbacchella:


HonzaKral:
Thanks for another catch, much appreciated!

PR #694, opened 2017-12-31
Transport.perform_request mutate headers.

fbacchella:
`header` variable is used both for request and response headers. That makes the code less readable and  error prone.


HonzaKral:
Thank you for the patch and the attention to detail, @fbacchella !

Issue #696, opened 2018-01-03
Typo in docs


kxxoling:
There's a typo here: https://github.com/elastic/elasticsearch-py/blob/8c13060c879bbe2cedac4a699a162fb9a9cb487f/elasticsearch/transport.py#L135

`crate` should be corrected to `create`.

fxdgear:
@kxxoling can you please submit a PR? :)

ao:
Created one for you ;) https://github.com/elastic/elasticsearch-py/pull/699

fxdgear:
closeed via https://github.com/elastic/elasticsearch-py/pull/699

PR #699, opened 2018-01-05
Issue: 696 - crate should be corrected to create

ao:
`crate should be corrected to create.` as per https://github.com/elastic/elasticsearch-py/issues/696


fxdgear:
Thanks!

Issue #400, opened 2016-05-07
Support for Amazon AWS Hosted Elastic


tofler:
Amazon's hosted elastic deployment requires that all requests to the service be signed using their protocol - provided we don't want to specify specific IPs in their access policy. In most production deployments, this would be the case. However, elasticsearch-py does not have the capability (or even the instrumentation) to allow the user to plugin a "Request Signer". A request signer takes all the parameters of the request and generates the additional headers that need to be added to the request.

We have locally written extensions to cope with this, alongwith a RequestSigner class that can sign elasticsearch requests appropriately to work with AWS. Given AWS' popularity and the structure of their hosted solution, it would be great if this can be supported natively in elasticsearch-py. Other hosted services might also require request signing in the future and hence generic pluggable Request Signing would be a useful addition.

[P.S. - We would like to contribute back code to do this in a generic way; as well as provide a functioning request signer that works for Amazon's hosted Elastic service]

ddorian:
Can't this, like, be just a connection class ?

tofler:
Technically, yes. But then that just means we'd be replicating most of the good code in Urllib3HttpConnection or RequestsHttpConnection. Happy to contribute back a connection class as well, if desired.

HonzaKral:
Did you try the approach described in the docs? http://elasticsearch-py.readthedocs.io/en/master/#running-on-aws-with-iam from what I understand it should do what you are asking for, but it has been a while since using AWS - am I missing something?

thanks

HonzaKral:
Closing since there is no new information, please feel free to reopen with new information if proposed solution doesn't work for you.

sonnenhaft:
Any update on this?

PR #703, opened 2018-01-08
boto connection class

humeniukd:
May close issue #400


humeniukd:
missing boto requirement

PR #704, opened 2018-01-08
Adding GZip support to urllib3

robgil:
Adds new urllib3 class GzipEnabledConnection for handling gzip compression.

On bulk loads, this increased performance from 3000 doc/s to ~25k docs/s on my connection.

### DIFF
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -161,7 +161,7 @@ class Urllib3HttpConnection(Connection):
             if headers:
                 request_headers = request_headers.copy()
                 request_headers.update(headers)
-            if self.http_compress == True:
+            if self.http_compress == True and body:
                 body = gzip.compress(body)
             response = self.pool.urlopen(method, url, body, retries=False, headers=request_headers, **kw)
             duration = time.time() - start
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -83,6 +83,10 @@ class Urllib3HttpConnection(Connection):
             for k in headers:
                 self.headers[k.lower()] = headers[k]
 
+        if self.http_compress == True:
+            self.headers.update(urllib3.make_headers(accept_encoding=True))
+            self.headers.update({'content-encoding': 'gzip'})
+
         self.headers.setdefault('content-type', 'application/json')
         pool_class = urllib3.HTTPConnectionPool
         kw = {}
@@ -156,8 +160,6 @@ class Urllib3HttpConnection(Connection):
             request_headers = self.headers
             if headers:
                 if self.http_compress == True:
-                    headers.update(urllib3.make_headers(accept_encoding=True))
-                    headers.update({'content-encoding': 'gzip'})
                     body = gzip.compress(body)
                 request_headers = request_headers.copy()
                 request_headers.update(headers)
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -159,10 +159,10 @@ class Urllib3HttpConnection(Connection):
 
             request_headers = self.headers
             if headers:
-                if self.http_compress == True:
-                    body = gzip.compress(body)
                 request_headers = request_headers.copy()
                 request_headers.update(headers)
+            if self.http_compress == True:
+                body = gzip.compress(body)
             response = self.pool.urlopen(method, url, body, retries=False, headers=request_headers, **kw)
             duration = time.time() - start
             raw_data = response.data.decode('utf-8')
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -161,7 +161,7 @@ class Urllib3HttpConnection(Connection):
             if headers:
                 request_headers = request_headers.copy()
                 request_headers.update(headers)
-            if self.http_compress == True and body:
+            if self.http_compress and body:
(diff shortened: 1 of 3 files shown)


fxdgear:
@robgil great thanks for the PR. I'll take a look. 👍

fxdgear:
can you add a test? Something simple to make sure that it's grabbing the right Class?

robgil:
I'm trying to make this compatible with python2.7, but Elastic Cloud is barfing on this with 400 errors during bulk load. `gzip.compress()` seems to be the only one I can get to work. 
```
    def perform_request(self, method, url, params=None, body=None, timeout=None, ignore=(), headers=None):
        gzip_body = BytesIO()
        with GzipFile(fileobj=gzip_body, mode='wb', compresslevel=3) as f:
            f.write(body)
        headers.update(urllib3.make_headers(accept_encoding=True))
        headers.update({'Content-Encoding': 'gzip'})
        conn = super().perform_request(method=method, url=url, params=params,
                body=gzip_body, timeout=timeout, ignore=(), headers=headers)
        return conn
```

When I run the unit tests against a local ES, I don't get any errors.

test_elasticsearch/test_connection.py
```diff
@@ -70,6 +70,62 @@ def test_ssl_context_and_depreicated_values(self):
         self.assertRaises(ImproperlyConfigured, Urllib3HttpConnection, ssl_context=ctx, ca_certs="/some/path/to/cert.crt")
         self.assertRaises(ImproperlyConfigured, Urllib3HttpConnection, ssl_context=ctx, ssl_version=ssl.PROTOCOL_SSLv23)
 
+class TestGzipEnabledConnection(TestCase):
```
HonzaKral: these tests are only testing logic from the `Urllib3HttpConnection`, which is already tested and there is no need to retest it. What we need instead are tests for the `gzip` specific part - `gzip` accept headers and the fact that body is correctly compressed.

elasticsearch/compat.py
```diff
@@ -8,8 +8,13 @@
     from urlparse import  urlparse
     from itertools import imap as map
     from Queue import Queue
+    StringIO = BytesIO = StringIO.StringIO
```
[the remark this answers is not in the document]
robgil: Yes, see note above about 2.7 compatibility. I was unable to get GzipFile to work with Elastic Cloud, but it works ok with a local install.

test_elasticsearch/test_connection.py
robgil: For each of the tests, while redundant, they use the GzipEnabledConnection and tests all the same stuff but with the gzip class.

robgil:
@HonzaKral See first comment regarding performance. You'd need to test this over a resource constrained network to see the performance improvement. All of the tests assume you're running a local ES server.

HonzaKral:
@robgil you are right, I completely missed the  number in the original PR description, my bad!

In that case this makes sense. I think we'd want to have an optional flag which will set the correct headers in `__init__` and then compress the body in `perform_request` of the original class (no subclassing, just add a conditional compression). Then we won't have to deal with the error handling or anything like that as it will remain the same.

Thanks again for raising this and sorry for the confusion from my part!

robgil:
@HonzaKral @fxdgear re-worked and added the `http_compress` option. Any other thoughts on testing? That test I added is just a placeholder. I didn't notice a general "request" test that I could validate the headers and response compression. Any advice on unit testing would be great. I don't know whether this should be a mock or a _real_ request against a local ES.

fxdgear:
@robgil sorry to do this to you but can you please rebase? 

I wanted to get the SSL fix in first

fxdgear:
Also, can you please add to the docs 
1. how to use gzip
2. when you'd want to use gzip vs not using gzip?

Thanks

elasticsearch/connection/http_urllib3.py
```diff
@@ -152,6 +155,10 @@ def perform_request(self, method, url, params=None, body=None, timeout=None, ign
 
             request_headers = self.headers
             if headers:
+                if self.http_compress == True:
+                    headers.update(urllib3.make_headers(accept_encoding=True))
+                    headers.update({'Content-Encoding': 'gzip'})
```
HonzaKral: the headers should be modified in `__init__` so it only happens once at initializatin and not with every single request, see where `self.headers` are created. Also please use lowercase for `content-encoding` to be consistent.

test_elasticsearch/test_connection.py
```diff
@@ -32,6 +32,10 @@ def test_ssl_context(self):
         )
         self.assertTrue(con.use_ssl)
 
+    def test_http_compression(self):
+        con = Urllib3HttpConnection(http_compress=True)
+        self.assertTrue(con.http_compress)
```
HonzaKral: we should also verify that `con.headers` are set properly, once the headers are set in `__init__`.

elasticsearch/connection/http_urllib3.py
```diff
@@ -152,6 +159,8 @@ def perform_request(self, method, url, params=None, body=None, timeout=None, ign
 
             request_headers = self.headers
             if headers:
+                if self.http_compress == True:
```
HonzaKral: this needs to be outside of the `if headers` block, otherwise it will only be used if custome headers are specified.

if should be:
```python
if self.http_compress and body:
    body = gzip.compress(body)
```

robgil: Derp! Updating.

fxdgear: good eye @HonzaKral

PR #706, opened 2018-01-08
boto connection class

humeniukd:
May close issue #400 in case it will be reopened


humeniukd:
Build failed due to https://github.com/travis-ci/travis-ci/issues/7940

HonzaKral:
I am sorry, but we cannot add a big dependency like `boto` to this library. This PR also has no tests, documentation, nor error handling.

I would recommend releasing this as a separate package that supplies the `Connection` class that can then be used with this library.

Thank you for the PR but unfortunately we cannot merge it

PR #711, opened 2018-01-11
fixing SSL auth issue in 5.x

fxdgear:
If trying to connect to SSL instances with SSL Client Auth, but not verifying the CA cert, the client cert/key will not be added to the request. 

Instead lets add to the KW to be passed into the request, and instead if `verify_certs=False` lets set `cert_reqs=CERT_NONE` and if `verify_certs=True` we can set `cert_reqs=CERT_REQUIRED`


sevdog:
Seems it should work, nice job :+1:

PR #713, opened 2018-01-15
Adds dynamic class-based `__repr__`

3lnc:
* minor typo/stale fix
 * makes `__repr__` dynamic, based on class name. Gives ability to inherit `Elasticsearch` without reloading `__repr__` and still having proper representation

Faced inconvenience based on this while working with `elasticsearch-py-async`. Not sure that this is proper place for this code, but copy-pasting logic to dependent libs doesn't seem to be good way.


3lnc:
ping @HonzaKral

HonzaKral:
Thank you for the patch, and the ping, @3lnc!

PR #714, opened 2018-01-18
Use use ssl_context or don't but don't mix

fxdgear:


elasticsearch/connection/http_urllib3.py
```diff
@@ -85,16 +85,25 @@ def __init__(self, host='localhost', port=9200, http_auth=None,
...
+            else:
+                ca_certs = CA_CERTS
+
+            if verify_certs:
+                warnings.warn('Use of `verify_certs` has been deprecated in favor of using SSLContext', DeprecationWarning)
+
+            if not ssl_context:
+                # if no SSLContext we must make one
+
+                if not ca_certs and verify_certs:
```
fxdgear: ca_certs will only be Empty is certifi is not installed. Might be better if passed into the create_ssl_context function.

elasticsearch/connection/http_urllib3.py
```diff
@@ -85,16 +85,25 @@ def __init__(self, host='localhost', port=9200, http_auth=None,
...
+                    # If no ca_certs and no sslcontext passed and asking to verify certs
+                    # raise error
+                    raise ImproperlyConfigured("Root certificates are missing for certificate "
+                        "validation. Either pass them in using the ca_certs parameter or "
+                        "install certifi to use it automatically.")
+            if ssl_version:
```
fxdgear: if `ssl_verison` is passed in, we need to make sure this gets applied to the SSLContext object. Fixing coming soon.

elasticsearch/connection/http_urllib3.py
```diff
@@ -116,9 +125,6 @@ def __init__(self, host='localhost', port=9200, http_auth=None,
                 'assert_hostname': ssl_assert_hostname,
                 'assert_fingerprint': ssl_assert_fingerprint,
                 'ssl_context': ssl_context,
-                'cert_file': client_cert,
-                'ca_certs': ca_certs,
-                'key_file': client_key,
```
fxdgear: by leaving these here in the previous version of the SSLContext iterations, we're "half-assing" the ssl context implemetation. And as such it's causing issues. Since we are creating the SSLContext for the user if they don't pass it in, we shouldn't need to worry about passing through these variables anyway. So i'm removing them

elasticsearch/connection/http_urllib3.py
```diff
@@ -105,7 +114,7 @@ def __init__(self, host='localhost', port=9200, http_auth=None,
                 except AttributeError:
                     ssl_context = None
 
-                if not verify_certs and ssl_context is not None:
+                if not verify_certs:
                     ssl_context.check_hostname = False
```
HonzaKral: this will fail if `ssl_context` is `None`. Please add tests

fxdgear: Ssl_context should never be None, if it is we need to fail loud. 

I’ll raise an exception in that case.

elasticsearch/connection/http_urllib3.py
```diff
@@ -80,7 +80,9 @@ def __init__(self, host='localhost', port=9200, http_auth=None,
         kw = {}
 
         # if providing an SSL context, raise error if any other SSL related flag is used
-        if ssl_context and (ca_certs or ssl_version):
+        if ssl_context and ( use_ssl or verify_certs or ca_certs or client_cert or
```
HonzaKral: `verify_certs` defaults to `True`

PR #725, opened 2018-02-08
Add support for script actions to bulk helpers.

msukmanowsky:
Docs specify that script actions are possible with the BULK endpoint but `expand_action` doesn't support them https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html.

This PR adds support to `expand_action` for actions that look like:

```python
{
    '_op_type': 'script',
    '_script': {
        "id": "test",
        "source": "ctx._source.counter += params.param1",
        "lang": "painless",
        "params": {
                "param1": 1
        }
    },
    "scripted_upsert": True,
    "upsert": {
        "counter": 1
    }
}
```

Open to ideas on tweaking that format, but it seemed sane and in the spirit of what happens with doc updates.

I've already signed a CLA.

### DIFF
--- a/elasticsearch/helpers/__init__.py
+++ b/elasticsearch/helpers/__init__.py
@@ -36,6 +36,14 @@ def expand_action(data):
     data = data.copy()
     op_type = data.pop('_op_type', 'index')
     action = {op_type: {}}
+
+    if op_type == 'script':
+        for key in ('scripted_upsert', 'upsert'):
+            if key in data:
+                action[key] = data.pop(key)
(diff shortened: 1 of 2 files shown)


msukmanowsky:
Apologies, my misunderstanding here.

Scripts are already supported if you yield an action like so:

```python
{
    '_op_type': 'update',
    '_index': 'things',
    '_type': 'thing',
    '_id': 'thing|1',
    '_source': {
        'script': {
            'source': "ctx._source['@version'] = ctx._version",
            'lang': 'painless'
        }
    }
}
```

Issue #553, opened 2017-03-15
Provide complete error information in `TransportError.__str__`


Naitreey:
I have in times found that when a `TransportError` is raised, its string representation (mainly `error.root_cause[0].reason`) was just  not providing enough information to pinpoint the cause of error. Sometimes `error.reason` shows the more direct and relevant information about error. In these cases, I had to manually catch the exception and inspect complete `TransportError.info`.

Thus I propose we should provide complete `TransportError.info` or something similar in `TransportError.__str__` implementation.

HonzaKral:
Thank you for this issue. We made the decision to only include the crucial information in the `__str__` representation as the data in `info` can be quite large and thoroughly impractical for display purposes.

Then, no matter what we decide, we cannot display more information - I would much rather encourage people to catch the exception and inspect the `info` attribute. We could perhaps try to do something half way like including both `error.reason` and `root_cause[0].reason` (if they are not the same, which is typically the case) as well..

Does this make sense?

Naitreey:
@HonzaKral Yeah, I think including both `reason`s in some way could be better. It will make more exceptions easier to understand, and... less pretty.

How about this, just one small step further: We could include two pairs of `type` and `reason`,
one from the outer `error`, another
from `root_cause[0]`, something like the following in string representation:

```
TransportError(status_code, (type, reason), (root_cause[0].type, root_cause[0].reason))
```

This will certainly make traceback much longer (and more annoying...) than it is now. But it provides two set of info from different exception perspectives.

fxdgear:
@Naitreey please see this PR: https://github.com/elastic/elasticsearch-py/pull/726

It looks like someone is attempting to add more info into the exceptions.

sethmlarson:
`TransportError.__repr__` has changed quite a lot. Maybe now things are good enough? Can reopen if this is still a problem.

PR #726, opened 2018-02-11
optimize TransportError when show error msg is incomplete

iruimeng:
when update es, mapping fileds more than "index.mapping.total_fields.limit", I only got error message is "TransportError(400, illegal_argument_exception, [10.xx.xx.15][]dices:data/write/update[s]]". that is not really error msg, so that can't tell me what' wrong.
Eg curl repsonse msg:
{"error":{"root_cause":[{"type":"remote_transport_exception","reason":"[10.xx.xx.241][10.xx.xx.241:19413][indices:data/write/update[s]]"}],"type":"illegal_argument_exception","reason":"Limit of total fields [1000] in index [my_index] has been exceeded"},"status":400}

### DIFF
--- a/elasticsearch/exceptions.py
+++ b/elasticsearch/exceptions.py
@@ -55,9 +55,12 @@ class TransportError(ElasticsearchException):
         cause = ''
         try:
             if self.info:
-                cause = ', %r' % self.info['error']['root_cause'][0]['reason']
+                if self.info['error']['root_cause'][0]['type'] ==  self.info['error']['type']:
+                    cause = ', %r' % self.info['error']['root_cause'][0]['reason']
+                else:
+                    cause = ', error root_cause reason: %r, error reason: %r' % (self.info['error']['root_cause'][0]['reason'], self.info['error']['reason'])
         except LookupError:
-            pass
+            cause = str(self.info)
         return 'TransportError(%s, %r%s)' % (self.status_code, self.error, cause)
 
 


fxdgear:
@iruimeng Thank you for the pull request. It's greatly appreciated. It would be helpful if you could provide some tests as well for this PR.

iruimeng:
@fxdgear this PR use  es version 5.x. text case eg:
`import unittest
import random
import time

ES_CONF = {
    'index': 'es_index_test',
    'doc_type': 'pixel',
    '''edit your own es service host user&pwd'''
    'hosts': 'https://localhost:9200'
}

class PrTest(unittest.TestCase):

    def __doc(self, es):
        dsl = {'size': 10000, '_source': [], 'query': {'constant_score': {'filter': {'bool': {'must': [{'term': {'ds': '2017-01-01'}}, {'term': {'unit_name': 'ds__product_id'}}]}}}}}
        return es.search(index=ES_CONF['index'], doc_type=ES_CONF['doc_type'], body=dsl)

    def uperror_test(self):
        es = Elasticsearch(ES_CONF['hosts'])

        update_doc = {"ds": "2017-01-01", "product_id": "21", 'five_minutes_online_ips_distribution': {'190': 0}, "unit_name": "ds__product_id"}
        docs = self.__doc(es)
        if docs['hits']['total'] == 0:
            r = es.index(index=ES_CONF['index'], doc_type=ES_CONF['doc_type'], body=update_doc)
            time.sleep(2)

        docs = self.__doc(es)
        self.assertEquals(1, len(docs['hits']['hits']))
        doc_id = docs['hits']['hits'][0]['_id']

        for i in range(1110):
            update_doc['five_minutes_online_ips_distribution'] =  {i: random.randint(1, 100)}
            print(update_doc)
            r = es.update(index=ES_CONF['index'], doc_type=ES_CONF['doc_type'], id=doc_id, body={'doc':update_doc})


if __name__ == "__main__":
    unittest.main()`

fxdgear:
@iruimeng I meant if you could add to the tests repo a test for exception handling. https://github.com/elastic/elasticsearch-py/tree/master/test_elasticsearch

Just so we have some examples of exceptions as well as some examples of how the exceptions should look.

Naitreey:
@fxdgear Sorry, I'm not currently using elasticsearch at my job, so it would not be very responsible for me to put more opinion on this one. But I'll just provide my two cents here.

As I see from the PR, it does put `self.info['error']['reason']` into `TransportError`, as I originally intended. I think this will improve usefulness of exception traceback, instead of
having to manually catching error at application code. On the other hand, traceback can be
less readable, less pretty. If that is an issue. I personally think, usefulness of traceback should be
more important.

iruimeng:
@fxdgear what am I going do next? add  a test for  exception handling or close this pr.  I see the [issue553](https://github.com/elastic/elasticsearch-py/issues/553), HonzaKral says about display more information, "this not make snese"

fxdgear:
@iruimeng I'd like to see a test that raises the exception, and in the exception text is displayed the exception you expect to see.

PR #727, opened 2018-02-13
Use proper retry_on_conflict for helpers.expand_action

msukmanowsky:
As per
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update,
retry_on_conflict should not be preceded by an underscore.


msukmanowsky:
By the way, confirmed that sending `_retry_on_conflict` results in a deprecation warning in 6.2.1 (deprecation logs)

```
[2018-02-13T21:15:45,454][WARN ][o.e.d.c.ParseField       ] Deprecated field [_retry_on_conflict] used, expected [retry_on_conflict] instead
```

Issue #744, opened 2018-03-14
Unable to serialize set with terms filter query


jaisharma639:
Please refer [this issue](https://github.com/elastic/elasticsearch/issues/29047).
Basically the need is to have type casting for objects that can't be serialized. In this case that object is a **set** which should be type casted to a list.

fxdgear:
@jaisharma639 I'm sorry to hear you're having problems. But the good news is that there's a solution to your problem! The python client by default uses the [`JSONSerializer`](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/serializer.py#L24). But you are able to overwrite that serializer with any serializer you want to use. 

```python
es = Elasticsearch(serializer=MyCustomSerializer)
```

Just did a quick little search for `python json serializer sets` and this [stack overflow](https://stackoverflow.com/a/8230373) came up. By default a `set` is not a native datatype that the JSON spec can handle. So as such it needs to be special cased.

jaisharma639:
I'm more than thrilled to read your response, couldn't find this anywhere. I think the only problem here was not able to find this particular documentation anywhere. Maybe it didn't come up in the top search hits. Anyway, really appreciate the response.

fxdgear:
If you want to share your serializer I'll make an example of it and put it in the docs. 

Hopefully make things easier for the next person!

jaisharma639:
Here's how you implement your own custome serializer on client side.

```
from elasticsearch.serializer import JSONSerializer

class SetEncoder(JSONSerializer):
    def default(self, obj):
        if isinstance(obj, set):
            return list(obj)
        if isinstance(obj, Something):
            return 'CustomSomethingRepresentation'
        return JSONSerializer.default(self, obj)
```

and just call while initializing your connection instance:
`es = Elasticsearch(serializer=SetEncoder())`

@fxdgear , does it look ok?

fxdgear:
> Can i raise a PR updating the documentation with this example?

Yes that would be great! :)

jaisharma639:
@fxdgear , can you please review my PR?

PR #747, opened 2018-03-15
Adding documentation for passing custom serializer to es client instance

jaisharma639:
While using and query where passing an array is required, a set cannot be used. This is because sets are not json serializable.
Discussion has been captured here https://github.com/elastic/elasticsearch-py/issues/744

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -292,6 +292,29 @@ the `requests-aws4auth`_ package::
 
 .. _requests-aws4auth: https://pypi.python.org/pypi/requests-aws4auth
 
+Customization
+-------------
+
+Custom serializers
+~~~~~~~~~~~~~~~~~~
+
+By default, `JSONSerializer`_ is used to encode all outgoing requests.
+However, you can implement your own custom serializer::
+
+   from elasticsearch.serializer import JSONSerializer
+
+   class SetEncoder(JSONSerializer):
+       def default(self, obj):
+           if isinstance(obj, set):
+               return list(obj)
+           if isinstance(obj, Something):
+               return 'CustomSomethingRepresentation'
+           return JSONSerializer.default(self, obj)
+
+   es = Elasticsearch(serializer=SetEncoder())
+
+.. _JSONSerializer: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/serializer.py#L24
+
 Contents
 --------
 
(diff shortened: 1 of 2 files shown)


jaisharma639:
I tried to run the test suite but got some [errors](https://gist.github.com/jaisharma639/be74476563cea40e2d63b86650b1043d). The errors look unrelated to the change. Please let me know about the next steps.

jaisharma639:
I signed the CLA after raising the PR. The documentation says that adding comments 5 minutes after signing should refresh the CLA checker. Adding this comment for the sake of it.

fxdgear:
@jaisharma639 Thanks looks good so far. 

But could you also add it into the `docs` section. 

Maybe a blurb in https://github.com/elastic/elasticsearch-py/blob/master/docs/index.rst

this way it'ls show up on http://elasticsearch-py.readthedocs.io/en/master/ as well.

jaisharma639:
With the current comments added, it'd show up on http://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch, right?
Wouldn't showing it again at http://elasticsearch-py.readthedocs.io/en/master/ be redundant information at 2 places?

fxdgear:
I was thinking it'd be nice for it to show up on the index.rst. BUT having the example is nice regardless. 

I can move it around later.

jaisharma639:
I can add it to index.rst. What should the documentation explain there? Should it contain the example encoder? If you give me a brief idea about it, i'll update the PR.

fxdgear:
something like this maybe:

```rst
Cusomizaion
~~~~~~~~~~~~~~~~~~~~~~~~~~~

Custom Serializers
---------------------

By default, `JSONSerializer
<https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/serializer.py#L24>`_
is used to encode all outgoing requests.
However, you can implement your own custom serializer::

   from elasticsearch.serializer import JSONSerializer

   class SetEncoder(JSONSerializer):
       def default(self, obj):
           if isinstance(obj, set):
               return list(obj)
           if isinstance(obj, Something):
               return 'CustomSomethingRepresentation'
           return JSONSerializer.default(self, obj)

   es = Elasticsearch(serializer=SetEncoder())
```

fxdgear:
Thanks!

PR #749, opened 2018-03-16
add pr#726 test

iruimeng:
add test for [https://github.com/elastic/elasticsearch-py/pull/726](pr726)


karmi:
Hi @iruimeng, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/749.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

iruimeng:
close

PR #761, opened 2018-04-03
Fix attempting to decode non-str when logging error

ZwodahS:
fix error when node 599, and there is no "body" to decode.
` File "***/venv3/lib/python3.5/site-packages/elasticsearch/connection/base.py", line 103, in log_request_fail
body = body.decode('utf-8', 'ignore')
AttributeError: 'float' object has no attribute 'decode'`


sethmlarson:
This PR looks out of date, if this is still an issue for you let me know here or open an issue and we can discuss. Thanks!

PR #767, opened 2018-04-11
Update connection.rst

daniel878:
Change to correct parameter name


fxdgear:
@daniel878 thanks for the patch. Would you mind signing our CLA? Then I can merge this.

PR #775, opened 2018-04-23
`TransportError` not raised in `helpers.streaming_bulk()` with `max_retries`.

Milly:
When `helpers.streaming_bulk()` is called with `max_retries > 0`, the `TransportError(429)` which raise in the last retry is ignored.

Fixed to re-raise `TransportError` if the last retry fails.


fxdgear:
@Milly thank you, for the patch with a test! very much appreciated.

PR #780, opened 2018-04-30
Fixes double creation of urllib3.HTTPSConnectionPool

IlyaRadinsky:
Fixes double creation of urllib3.HTTPSConnectionPool when ssl_context and use_ssl are used.


IlyaRadinsky:
Done

PR #791, opened 2018-05-16
Drop support for EOL Python 2.6, 3.2 and 3.3

### LINKED ISSUE #788: Drop support for Python 2.6
Python 2.6's EOL was with the [2.6.9 release in October 2013](https://www.python.org/download/releases/2.6.9/).  That's nearly 5 years past the release date. And [2.7 isn't going to last much longer](https://pythonclock.org/).

If issues like #787 are, indeed, from functions not working as expected—whether in our code, or the code of dependencies—we should remove support for Python 2.6 [here](https://github.com/elastic/elasticsearch-py/blob/master/setup.py#L50), to say nothing of support for deprecated versions 3.2, 3.3, and 3.4.

They would also need to be removed [here](https://github.com/elastic/elasticsearch-py/blob/master/.travis.yml#L4), and [here](https://github.com/elastic/elasticsearch-py/blob/master/.travis.yml#L6-L7)

hugovk:
Fixes #788.

Here's the pip installs for elasticsearch from PyPI for April 2018, showing low numbers for 2.6 and even lower for 3.2 and 3.3:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  57.69% |        350,098 |
| 3.6            |  23.78% |        144,329 |
| 3.4            |   9.88% |         59,932 |
| 3.5            |   8.11% |         49,221 |
| 2.6            |   0.48% |          2,916 |
| 3.3            |   0.03% |            180 |
| 3.7            |   0.02% |            131 |
| 3.2            |   0.01% |             32 |
| 3.8            |   0.00% |              1 |
| Total          |         |        606,840 |

Source: `pypinfo --start-date 2018-04-01 --end-date 2018-04-30 --percent --pip --markdown elasticsearch pyversion`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 
 python:
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -237,7 +237,7 @@ an issue) it also just uses ``localhost:9200`` as the address instead of the
 actual address of the host. If the trace logger has not been configured
 already it is set to `propagate=False` so it needs to be activated separately.
 
-.. _logging library: http://docs.python.org/3.3/library/logging.html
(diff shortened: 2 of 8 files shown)


fxdgear:
@hugovk ❤️  I appreciate the work on this. I really do. But the unfortunate aspect to this is that there are still people who depend on py26 specifically those in the redhat community. I would **LOVE** to merge this branch but now is not the right time. I'm sorry 🐼

PR #811, opened 2018-06-25
Check if logger.DEBUG exists before decoding the body. Closes #765

### LINKED ISSUE #765: Body shouldn't be decoded needlessly
Body decoding in https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/base.py#L78-L79 should only be done if we are actually going to use it in the logs by either `log_request_success` or `_log_trace`

fxdgear:


fxdgear:
stale beyond stale. closing.

PR #815, opened 2018-07-05
example requirements.txt add elasticsearch

rdkls:


fxdgear:
@rdkls Thanks for the patch. 

Though I'm wondering why this needs to be added? 

If you're using the examples you should be using master. Not the latest pushed vesion to PyPI cause there's no guarentee that they'll be the same. 

You should have already run `python setup.py install` or `python setup.py develop` locally after checking out the repo. 

If thats not clear I can add some docs to help explain...

rdkls:
You're right I just cloned the repo and jumped straight into running the example; docs are clear (for those who actually read) them thanks Nick :)

PR #827, opened 2018-07-23
Pass `Retry` object to urllib3 instead of `False`

### LINKED ISSUE #826: Excessive logging from urllib3.util.retry
Hello!

I'm using Requests in my project, and have set up retries using urllib3's retry util. To monitor the retries, I've configured that modules logger as it logs useful information.

Unfortunately, this also generates a lot of less useful logs, which are triggered by elasticsearch-py.
`elasticsearch.connection.http_urllib3` [always sets](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/http_urllib3.py#L171) `retries=False` when calling `urllib3.connectionpool.urlopen`. This causes urllib3 to call the `from_int` class method of the `Retry` class, which ultimately causes the following to be logged: `Converted retries value: False -> Retry(total=False, connect=None, read=None, redirect=0, status=None)`

It would be better to instead to call `urlopen` with a `Retry` object with `total=False`, which will be functionally equivalent but suppress the noisy log message.

I'm going to raise a PR for the update for your consideration.

Thanks!

Wynndow:
Closes #826 

By passing a `Retry` object to `urlopen` instead of just `False`, it
prevents `urllib3` from having to create a Retry object with `from_int`.

`from_int` emits an undesirable log message when called with
`retries=False`, and this can be avoided by the method described above.

I'm specifically interested in having this change available for version 5.4 - is this PR correct for doing this, or would it go into the v6 release? Thanks!

### DIFF
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -2,6 +2,7 @@ import time
 import ssl
 import urllib3
 from urllib3.exceptions import ReadTimeoutError, SSLError as UrllibSSLError
+from urllib3.util.retry import Retry
 import warnings
 import gzip
 
@@ -168,7 +169,7 @@ class Urllib3HttpConnection(Connection):
                     # again
                     body = gzip.zlib.compress(body)
 
-            response = self.pool.urlopen(method, url, body, retries=False, headers=request_headers, **kw)
+            response = self.pool.urlopen(method, url, body, retries=Retry(False), headers=request_headers, **kw)
             duration = time.time() - start
             raw_data = response.data.decode('utf-8')
         except Exception as e:


fxdgear:
Thanks for the patch. 

this will by default go in the next 6.x release. But I can cherrypick this commit into the 5's

Wynndow:
Great, thanks very much! If you could cherry pick into the 5's that'd be really helpful.

Cheers,

Chris

Wynndow:
@fxdgear Hello, do you know when this might get cherrypicked into version 5? Thanks!

fxdgear:
cherrypicked and pushed new version to pypi. Sorry about the delay. I had some other pressing things that needed done, and this slipped my mind :(

Wynndow:
Awesome, thanks so much.

PR #870, opened 2018-11-16
better logic to handling unicode and chunking.

### LINKED ISSUE #716: helpers._chunk_actions produces chunks bigger than max_chunk_bytes
Hello! Thanks for the library.

We started to get ocasional errors from Elasticsearch
```
org.jboss.netty.handler.codec.frame.TooLongFrameException: HTTP content length exceeded 104857600 bytes.
```
So I investigated and found an issue in `helpers._chunk_actions`
There are two cases

1. If `max_chunk_bytes` is pretty low, `_chunk_actions` will split actions one by one, but each chunk will be bigger than `max_chunk_bytes`. It's actually used in test https://github.com/elastic/elasticsearch-py/blob/64202273e02ab5f0b99271771d4e8143b3599c3e/test_elasticsearch/test_helpers.py#L35 Maybe it makes sense to raise a warning in this case. 
2. **Most important** `_chunk_actions` count characters instead of bytes to check for chunk size. If user passes actions with unicode values it is possible that output chunk size will exceed `max_chunk_bytes`. This test failes
```python
class TestChunkActions(TestCase):
    def setUp(self):
        super(TestChunkActions, self).setUp()
        self.actions = [({'index': {}}, {'some': u'€€€€€€€€', 'i': i}) for i in range(100)]

    def test_chunks_are_chopped_by_byte_size_properly(self):
        chunks = list(helpers._chunk_actions(self.actions, 100000, 90, JSONSerializer()))
        self.assertEquals(100, len(chunks))
        for chunk in chunks:
            chunk = u''.join(chunk)
            chunk = chunk if isinstance(chunk, str) else chunk.encode('utf-8')
            self.assertLessEqual(len(chunk), 90)
```
`_chunk_actions` should produce 100 chunks (each less than 90 bytes), but produces 50 chunks (each ~114 bytes)

fxdgear:
Chunking was looking at the length of a string but not accounting for the
byte length. closes #716

### DIFF
--- a/elasticsearch/helpers/__init__.py
+++ b/elasticsearch/helpers/__init__.py
@@ -58,11 +58,12 @@ def _chunk_actions(actions, chunk_size, max_chunk_bytes, serializer):
     for action, data in actions:
         raw_data, raw_action = data, action
         action = serializer.dumps(action)
-        cur_size = len(action) + 1
+        # +1 to account for the trailing new line character
+        cur_size = len(action.encode('utf-8')) + 1
 
         if data is not None:
             data = serializer.dumps(data)
-            cur_size += len(data) + 1
+            cur_size += len(data.encode('utf-8')) + 1
 
         # full chunk, send it and start a new one
         if bulk_actions and (size + cur_size > max_chunk_bytes or action_count == chunk_size):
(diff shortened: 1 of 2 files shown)


fxdgear:
Jenkins retest this please.

PR #875, opened 2018-11-26
Fixing flaky test wrt mock and threading.

fxdgear:
Mock is apparetnly not threadsafe, and as a result will sometimes
cause tests to fail when we want to count the number of times a
mock function is called.

I've replaced the mock function with a side_effect, which allows for
locking and incrementing manually.


fxdgear:
Jenkins retest this please

PR #881, opened 2018-12-12
add time/size params in cat api

LoadingZhang:


elasticcla:
Hi @LoadingZhang, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/881.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #894, opened 2019-01-28
BlockingPool must fit thread_count

### LINKED ISSUE #816: elasticsearch.helpers.parallel_bulk() hangs forever on Python 3.7.0
I haven't had a chance to dig into what's new in Python 3.7.0, or how `elasticsearch.helpers.parallel_bulk()` works.  But for some reason, it hangs forever on Python 3.7.0.

`elasticsearch.helpers.streaming_bulk()` works fine on Python 3.7.0, and my app isn't indexing-heavy, so I've just switched to streaming for now.

sirdodger:
queue_size must be greater than or equal to thread_count, or else the main thread can hang trying to acquire a lock to write sentinel values to the queue during teardown.

It is covered by the existing test_chunk_sent_from_different_threads(), which should now work on Python 3.7.2.  If you think a mocked test to test the max() behavior specifically is necessary, let me know.

Closes #816

### DIFF
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -292,7 +292,9 @@ def parallel_bulk(client, actions, thread_count=4, chunk_size=500,
     class BlockingPool(ThreadPool):
         def _setup_queues(self):
             super(BlockingPool, self)._setup_queues()
-            self._inqueue = Queue(queue_size)
+            # The queue must be at least the size of the number of threads to 
+            # prevent hanging when inserting sentinel values during teardown.
+            self._inqueue = Queue(max(queue_size, thread_count))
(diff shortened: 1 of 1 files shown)


fxdgear:
jenkins test this please

sirdodger:
The clients-ci failure looks like a transient network issue; they could use a re-run.

seperman:
Any updates on this? Thanks.

fxdgear:
Jenkins retest this please

fxdgear:
Thanks for the patch, 

the 3.3 failures I'm not worried about since it's EOL'd and no longer supported and the matrix elements have been removed in `master` the 2.7 failure is a known issue wrt to the fact that mock is not thread safe.

PR #898, opened 2019-01-29
Use relative imports in xpack module

### LINKED ISSUE #897: Use relative imports in the xpack module
The `elasticsearch.client.xpack` module uses absolute imports in multiple places e.g. (https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/xpack/__init__.py#L1) which is not consistent with the rest of the library and also it means that the library cannot be vendored without code changes.

beezz:
Closes #897

### DIFF
--- a/elasticsearch/client/xpack/__init__.py
+++ b/elasticsearch/client/xpack/__init__.py
@@ -1,4 +1,4 @@
-from elasticsearch.client.utils import NamespacedClient, query_params
+from ..utils import NamespacedClient, query_params
 
 from .graph import GraphClient
 from .license import LicenseClient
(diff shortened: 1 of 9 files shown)


elasticcla:
Hi @beezz, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-py/pull/898.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

fxdgear:
jenkins test this please

fxdgear:
Thanks for the patch.

Issue #890, opened 2019-01-10
elasticsearch.client.Elasticsearch.scroll causes Validation Failed: 1: scrollId is missing


k3v142:
When I execute this row:
```python
page = es.scroll(scroll_id = scroll_id, scroll = '1m')
```
I get:
```
Traceback (most recent call last):
  File "es.py", line 19, in <module>
    for hit in response:
  File "/home/osboxes/Desktop/test/lib/python3.5/site-packages/elasticsearch_dsl/search.py", line 726, in scan
    **self._params
  File "/home/osboxes/Desktop/test/lib/python3.5/site-packages/elasticsearch/helpers/__init__.py", line 379, in scan
    **scroll_kwargs)
  File "/home/osboxes/Desktop/test/lib/python3.5/site-packages/elasticsearch/client/utils.py", line 76, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/home/osboxes/Desktop/test/lib/python3.5/site-packages/elasticsearch/client/__init__.py", line 1016, in scroll
    params=params, body=body)
  File "/home/osboxes/Desktop/test/lib/python3.5/site-packages/elasticsearch/transport.py", line 318, in perform_request
    status, headers_response, data = connection.perform_request(method, url, params, body, headers=headers, ignore=ignore, timeout=timeout)
  File "/home/osboxes/Desktop/test/lib/python3.5/site-packages/elasticsearch/connection/http_urllib3.py", line 186, in perform_request
    self._raise_error(response.status, raw_data)
  File "/home/osboxes/Desktop/test/lib/python3.5/site-packages/elasticsearch/connection/base.py", line 125, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(status_code, error_message, additional_info)
elasticsearch.exceptions.RequestError: RequestError(400, 'action_request_validation_exception', 'Validation Failed: 1: scrollId is missing;')
```
In this case `elasticsearch-py` puts `scroll_id` to the body ([`elasticsearch-py/elasticsearch/client/__init__.py:998`](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/__init__.py#L998)), but `scroll_id` in `body` also doesn't pass validation even during manual testing, but if I put it to the GET params it work OK. 
Then it tried to modify `elasticsearch.client.Elasticsearch.scroll` from:
```python
if scroll_id in SKIP_IN_PATH and body in SKIP_IN_PATH:
[...]

PR #904, opened 2019-02-12
Fix scroll_id issue

k3v142:
#890 
I am not really sure it is correct changes but it works for me.


fxdgear:
Jenkins test this please

fxdgear:
I'm not certain this is the right fix. 

If you look at the docs for [`scroll`](https://www.elastic.co/guide/en/elasticsearch/reference/6.7/search-request-scroll.html) you can see that `scroll_id` is part of the body.

fxdgear:
@ k3v142 

Thanks for the patch, though when I was upgrading the client to 7.x I found that there's some differences with the ES7 api for scroll. It required a few more changes that altered some tests as well. 

I am sorry to close this on you. BUT if it wasn't for this PR I wouldn't have realized the api changes so quickly.

Issue #906, opened 2019-02-22
Add index name in traceback NotFoundError


mazzma12:
Hi,

Would it be possible to display the index name in the traceback raised by a NotFoundError Exception ? 

```python
raise HTTP_EXCEPTIONS.get(status_code, TransportError)(status_code, error_message, additional_info)
elasticsearch.exceptions.NotFoundError: NotFoundError(404, 'index_not_found_exception', 'no such index')
```
When this error is raised it's hard to see which index is at fault. 
Haven't dived in the code yet but maybe `urllib` returns error codes only?  

Using versions 6.3.1

fxdgear:
theoretically it should be possible. If you want to create a patch I'll happily review it for you.

sethmlarson:
This issue has been fixed. Thanks for the contribution!

PR #909, opened 2019-03-07
Add resource type and resource id in TransportError Traceback

mazzma12:
Hi @fxdgear. 
I modified the `TransportErrorException` class for #906 . I changed the str formatting with % into the `str.join()` method. I found it more flexible to deal with empty strings and extra commas. Let me know if you think it is a problem (I checked the tests and it passed)

### DIFF
--- a/elasticsearch/exceptions.py
+++ b/elasticsearch/exceptions.py
@@ -4,6 +4,7 @@ __all__ = [
     'SSLError', 'ConnectionTimeout', 'AuthenticationException', 'AuthorizationException'
 ]
 
+
 class ImproperlyConfigured(Exception):
     """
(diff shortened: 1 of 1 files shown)


mazzma12:
@fxdgear ?

fxdgear:
Jenkins test this please

fxdgear:
@mazzma12 Could you please signt he CLA? 

Your tests are passing. It just failed cause some python3.3 weirdness (which is ok cause I'm removing this element from the test matrix cause 3.3 is EOL'd)

mazzma12:
@fxdgear Happy to contribute, even as small it could be. Signed !

PR #921, opened 2019-03-29
Document that 'index' may also be a list of strings for most methods

adamchainz:
I discovered whilst researching https://github.com/scoutapp/scout_apm_python/issues/150 that the index parameter for all methods also supports a list of strings, but this isn't documented here. For example `elasticsearch-dsl-py` [always passes a list of strings](https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/search.py#L100).

It's [`_make_path`](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L44) that allows this to happen.

This updates the relevant methods' documentation to indicate that it is supported.

### DIFF
--- a/elasticsearch/client/__init__.py
+++ b/elasticsearch/client/__init__.py
@@ -710,7 +710,8 @@ class Elasticsearch(object):
         Execute a search query and get back search hits that match the query.
         `<http://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html>`_
 
-        :arg index: A comma-separated list of index names to search; use `_all`
+        :arg index: A list of index names to search, or a string containing a
+            comma-separated list of index names to search; use `_all`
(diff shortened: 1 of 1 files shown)


fxdgear:
I appreciate the patch, but these docs are mostly auto generated from the [elastic/elasticsearch](https://github.com/elastic/elasticsearch) repo. This line in particular can be found here: https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/search.json#L11

I'll merge this, but there's no guarantee that it'll end up staying in the long run :(

adamchainz:
Okay. Would you accept a test that ensures these lines are present in the docstrings? I think it's important to keep the docs correct here as the Python layer is adding an extra option, and it broke our product.

PR #924, opened 2019-04-04
Scan refactor

3lnc:
Since `scan` search-type is deprecated since 2.1, there's simpler implementation for `scan` logic.

Also adds tests for scan-related stuff

### DIFF
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -427,33 +427,19 @@ def scan(
     if not preserve_order:
         query = query.copy() if query else {}
         query["sort"] = "_doc"
+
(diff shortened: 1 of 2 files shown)


fxdgear:
jenkins test this please

3lnc:
Side note on this: for next major release I propose to change this one step further — let's make error-handling run before yielding results. This will not affect case with `raise_on_error=False`, but will introduce fast exit route for `raise_on_error=True`. IMO, raising exception as soon as we see problem in response makes more sense that yielding batch of results first.

fxdgear:
i've yet to do the 7.x release, so if  you want to do that now feel free.

PR #938, opened 2019-04-28
Added equality check for Connection class

### LINKED ISSUE #936: PriorityQueue for dead Connection can have the same timestamp
I recently started investigating this issue in aioelasticsearch:
https://github.com/aio-libs/aioelasticsearch/issues/100

And it turns out that potentially if few of the connections getting killed at the same time it could raise an error because PriorityQueue tries to compare two instances of Connection.
In my opinion, two instances of Connection should be considered equal and PriorityQueue then can easily handle such a situation. 

I'd gladly provide a PR for that feature if you're okay with this.

Here's what I did in case of aioelasticsearch:
https://github.com/aio-libs/aioelasticsearch/pull/164

svalee:
This commit fixes #936 

Do we need a test for this, what do you think?

### DIFF
--- a/elasticsearch/connection/base.py
+++ b/elasticsearch/connection/base.py
@@ -46,6 +46,13 @@ class Connection(object):
     def __repr__(self):
         return '<%s: %s>' % (self.__class__.__name__, self.host)
 
+    def __eq__(self, other):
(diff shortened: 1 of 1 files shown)


fxdgear:
Jenkins test this please

PR #928, opened 2019-04-12
Add ignore_throttled to query_params for search

andresp99999:
When trying to query a frozen index in ES6 with the elasticsearch library I got the following exception:

```
....
  File ".../lib/python2.7/site-packages/elasticsearch/client/utils.py", line 73, in _wrapped
    return func(*args, params=params, **kwargs)
TypeError: search() got an unexpected keyword argument 'ignore_throttle
```

It looks like we need to add: ignore_throttled to the allowed query parameters for search (https://www.elastic.co/guide/en/elasticsearch/reference/current/searching_a_frozen_index.html)

This change should be applied for 6.x and 7.x versions of the lib.

PR #942, opened 2019-04-29
Add support for ignore_throttled search request parameter

andresp99999:
Backport to 6.x of change: https://github.com/elastic/elasticsearch-py/pull/928 to allow using frozen indexes.


andresp99999:
A question, any change for 6 should be sent to 6.x or the 6.4.0 tag?

fxdgear:
@andresp99999 I greatly appriciate the help adding this. 

I've got a big PR for 6.x in the works (merging soon). 

Which does this, but also addresses all the other missing features for 6. 

I'll be releasing a new `elasticsearch6` package on pypi shortly after. 

Thanks for bringing this to my attention, which also made me realize there's probably others that are missing as well.

PR #948, opened 2019-05-09
Black

fxdgear:
Lets just format all the code to be "black" this way it's not an issue.


fxdgear:
jenkins test this again please

Failed due to network timeouts.

Issue #971, opened 2019-06-14
Regression in 6.4: Scroll failes with large scroll_id


ChrisPortman:
changes to the scoll method in 6.4 submits the scroll id as part of the URL.  This causes:

elasticsearch.exceptions.RequestError: RequestError(400, 'too_long_frame_exception', 'An HTTP line is larger than 4096 bytes.')

When there are a large number of shards involved creating a large scroll id.

https://github.com/elastic/elasticsearch-py/blob/99effab913c29ce341b3199f042bcb45cf8291a2/elasticsearch/client/__init__.py#L1341

mwsmith2:
I struggled with this same problem today.  I found that I could resolve the problem by making some minor changes to the scroll method, using POST instead of GET and stuffing the `scroll_id` into the body.

https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/__init__.py#L1300

```
    @query_params("scroll", "rest_total_hits_as_int", "scroll_id")
    def scroll(self, body=None, params=None):
        """
        Scroll a search request created by specifying the scroll parameter.
        `<http://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html>`_

        :arg scroll_id: The scroll ID
        :arg body: The scroll ID if not passed by URL or query parameter.
        :arg scroll: Specify how long a consistent view of the index should be
            maintained for scrolled search
        :arg rest_total_hits_as_int: This parameter is used to restore the total hits as a number
            in the response. This param is added version 6.x to handle mixed cluster queries where nodes
            are in multiple versions (7.0 and 6.latest)
        """
        _body = f'"scroll_id": {body}'
        return self.transport.perform_request(
            "POST", "/_search/scroll", params=params, body=_body
        )
```

fxdgear:
I'm sorry you're running into issues, but unfortunately (or fortunately, because it's a config change and not a code change/new release) this error is not to do with the Elasticsearch-py client, but Elasticsearch itself. 

The error you're getting is coming back directly from Elasticsearch. 

For reference please see: https://github.com/elastic/elasticsearch/issues/3210

In order to fix this you need to adjust some of the parameters in Elasticsearch itself.
For this particular error you need to change [`http.max_initial_line_length`](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html).

I hope this helps

ChrisPortman:
[...]

PR #974, opened 2019-06-25
Use HTTP POST for elasticsearch.scroll().

gmazzola:
Addresses Issue #971 

This pull request POSTs a JSON payload to the Elasticsearch scroll endpoint. We generate the POST body sanely by prioritizing the `body` argument first, and if it is absent, we generate a valid `body` using the `scroll_id` and `scroll` arguments.

This PR does not change the method signature. I tested it against an Elasticsearch 6.2.4 cluster and also via the existing unittests. I have also signed the contributor agreement. :)

### DIFF
--- a/elasticsearch/client/__init__.py
+++ b/elasticsearch/client/__init__.py
@@ -1296,8 +1296,8 @@ class Elasticsearch(object):
             "GET", _make_path(index, doc_type, id, "_explain"), params=params, body=body
         )
 
-    @query_params("scroll", "rest_total_hits_as_int", "scroll_id")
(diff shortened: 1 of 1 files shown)


elasticcla:
Hi @gmazzola, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-py/pull/974.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

gmazzola:
Will re-open this pull request using my personal git credentials. Sorry!

PR #975, opened 2019-06-25
Use HTTP POST for elasticsearch.scroll().

gmazzola:
Addresses Issue #971

This pull request POSTs a JSON payload to the Elasticsearch scroll endpoint. We generate the POST body sanely by prioritizing the body argument first, and if it is absent, we generate a valid body using the scroll_id and scroll arguments.

This PR does not change the method signature. I tested it against an Elasticsearch 6.2.4 cluster and also via the existing unittests.

This PR is complimentary to the one proposed by jtyoung. I'm patching the `scroll` method that directly calls `/_search/scroll`, whereas he is patching the `scan` method so that it forces a HTTP POST in `scroll`.

### DIFF
--- a/elasticsearch/client/__init__.py
+++ b/elasticsearch/client/__init__.py
@@ -1296,8 +1296,8 @@ class Elasticsearch(object):
             "GET", _make_path(index, doc_type, id, "_explain"), params=params, body=body
         )
 
-    @query_params("rest_total_hits_as_int")
-    def scroll(self, scroll=None, scroll_id=None, body=None, params=None):
+    @query_params("scroll", "rest_total_hits_as_int", "scroll_id")
+    def scroll(self, body=None, params=None):
         """
         Scroll a search request created by specifying the scroll parameter.
         `<http://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html>`_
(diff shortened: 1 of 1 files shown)


fxdgear:
Jenkins test this please

Winterflower:
Hi @gmazzola - thanks very much for your PR! Just going through the Jenkins output it seems that there are a few failures in the tests. Seems most of them have to do with 

```
18:57:26 FAIL: test_elasticsearch.test_server.test_common:TestScroll11Clear.test_from_yaml_0 (scroll/11_clear.yml): Clear scroll
18:57:26 ----------------------------------------------------------------------
18:57:26 Traceback (most recent call last):
[omitted]
18:57:26     self.assertIsInstance(exception, TransportError)
18:57:26 AssertionError: ValueError('You need to supply (scroll and scroll_id) or body.') is not an instance of <class 'elasticsearch.exceptions.TransportError'>
```

which looks like it's coming from the fact that you raise a value error in line 1316 in your PR. The test code appears to be excepting an elasticsearch specific TransportError exception. I haven't been in this part of the codebase too frequently so I can't comment on whether or not this a good thing, but happy to hear your thoughts.

elasticsearch/client/__init__.py
```diff
@@ -1311,8 +1311,20 @@ def scroll(self, body=None, params=None):
             are in multiple versions (7.0 and 6.latest)
         """
 
+        if scroll in SKIP_IN_PATH and body in SKIP_IN_PATH or \
+                scroll_id in SKIP_IN_PATH and body in SKIP_IN_PATH:
+            raise ValueError("You need to supply (scroll and scroll_id) or body.")
```
Winterflower: It looks like the CI tests are failing due to this exception being of the wrong type. I can dig into this a bit further to understand what's going on.

fxdgear:
Jenkins test this please

elasticsearch/client/__init__.py
gmazzola: Hi @Winterflower . Thanks for being receptive to this PR. :)

The CI failure was my fault -- I didn't fully understand the `query_params` implementation, and my safety check on lines 1314 and below effectively re-implemented that existing logic.

The PR is much simpler now: just a one-line change in total. Thanks for your help!

sethmlarson:
This has been resolved by #1145.

Issue #969, opened 2019-06-11
Version 7.0.1: Getting ConnectionError(a float is required) when passing request_timeout in the client calls


jyothikp:
I am using the version 7.0.1 and getting this error when passing the request_timeout in client calls:
es.indices.delete_template(
                name=template_name, 
                master_timeout="60s",
                request_timeout= 90
            )
`Traceback (most recent call last):
    \n  File "<ipython-input-13-d6617379fbe8>", line 4, in <module>    \n    request_timeout=template_timeouts.get(\'request_timeout_secs\', 90)    \n  File "build/bdist.linux-x86_64/egg/elasticsearch/client/utils.py", line 84, in _wrapped    \n    return func(*args, params=params, **kwargs)    \n  File "build/bdist.linux-x86_64/egg/elasticsearch/client/indices.py", line 607, in delete_template    \n    "DELETE", _make_path("_template", name), params=params    \n  File "build/bdist.linux-x86_64/egg/elasticsearch/transport.py", line 353, in perform_request    \n    timeout=timeout,    \n  File "build/bdist.linux-x86_64/egg/elasticsearch/connection/http_urllib3.py", line 232, in perform_request    \n    raise ConnectionError("N/A", str(e), e)    \nConnectionError: ConnectionError(a float is required) caused by: TypeError(a float is required)    \n`

I am making a delete_template call here. I am suspecting it started coming after this change:
https://github.com/elastic/elasticsearch-py/pull/937, as it was working in version 7.0.0. With this change request_timeout will go as normal query_params and _escape will be called on this value which will convert it to a string. While urllib library is expecting this parameter as a float value. Please let me know if any extra information required around the issue.

tj8000rpm:
I have a same issue at ``update_aliases``. It was working in version 6.3.

Here is my code and my stack traces.

**code**
```python
from elasticsearch import Elasticsearch
es = Elasticsearch('172.17.0.1:9200')

update_alias={
    "actions" : [
        { "remove" : { "index" : "hoge1", "alias" : "hoge" } },
        { "add" :    { "index" : "hoge2", "alias" : "hoge" } }
    ]
}
es.indices.update_aliases(body=update_alias, request_timeout=30)
```

**traceback**
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/urllib3/util/timeout.py", line 130, in _validate_timeout
    if value <= 0:
TypeError: '<=' not supported between instances of 'str' and 'int'

[...]

PR #976, opened 2019-06-25
fix invalid type of request_timeout (#969)

tj8000rpm:
Some elasticsearch indices object methods(such as delete_template and update_aliases) could not handle request_timeout argument correctly (It was expected pass as int, but it was actually as str).

Fixed the processing order in decorator query_params because it  could not process correctly special arguments.

It seems likely the original section of "don't treat ignore and request_timeout as other params to avoid escaping" does nothing.

And there are no test about decorator query_params, so added few test cases for this issue.

### DIFF
--- a/elasticsearch/client/utils.py
+++ b/elasticsearch/client/utils.py
@@ -71,16 +71,16 @@ def query_params(*es_query_params):
             params = {}
             if "params" in kwargs:
                 params = kwargs.pop("params").copy()
-            for p in es_query_params + GLOBAL_PARAMS:
-                if p in kwargs:
-                    v = kwargs.pop(p)
-                    if v is not None:
-                        params[p] = _escape(v)
 
             # don't treat ignore and request_timeout as other params to avoid escaping
             for p in ("ignore", "request_timeout"):
                 if p in kwargs:
                     params[p] = kwargs.pop(p)
+            for p in es_query_params + GLOBAL_PARAMS:
+                if p in kwargs:
+                    v = kwargs.pop(p)
+                    if v is not None:
+                        params[p] = _escape(v)
             return func(*args, params=params, **kwargs)
 
         return _wrapped
--- a/test_elasticsearch/test_client/test_utils.py
+++ b/test_elasticsearch/test_client/test_utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from elasticsearch.client.utils import _make_path, _escape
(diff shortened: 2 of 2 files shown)


Winterflower:
Jenkins test this please

tj8000rpm:
I fixed my test case problems of fail in case of python 2.7, can you jenkins test run again please?

fxdgear:
Jenkins test this please

tj8000rpm:
Thank you for supporting, and kicking a test. @fxdgear and @Winterflower .
I'm also wondering why decorators don't work with Python 2.7 test code ....  But my change seems work well for both python 2 and 3.

However clients-ci was failed but it was successfully end in my local environment test using the command shown below(and attach the build log).
[build.log.txt](https://github.com/elastic/elasticsearch-py/files/3326463/build.log.txt)

```
ELASTICSEARCH_VERSION=7.0.1 PYTHON_VERSION=2.7 make run_tests
```

Is this like a temporary failure?

Winterflower:
Hi @tj8000rpm - it does look like it might be sporadic
The failure seems to be happening in a different method now


```
15:46:19 ======================================================================
15:46:19 FAIL: test_all_chunks_sent (test_elasticsearch.test_helpers.TestParallelBulk)
15:46:19 ----------------------------------------------------------------------
15:46:19 Traceback (most recent call last):
15:46:19   File "/usr/local/lib/python2.7/site-packages/mock/mock.py", line 1330, in patched
15:46:19     return func(*args, **keywargs)
15:46:19   File "/code/elasticsearch-py/test_elasticsearch/test_helpers.py", line 38, in test_all_chunks_sent
15:46:19     self.assertEquals(50, _process_bulk_chunk.call_count)
15:46:19 AssertionError: 50 != 49
15:46:19 
```

I wonder if I run the CI again, if it will pass.

Winterflower:
Jenkins test this please

est:
This is a **** bug. Why do someone put @query_params("request_timeout",) in the first place??

It's hard to debug such obscure error. 

Which version do I need to reverse to in order to use under 2.7?

sethmlarson:
`request_timeout` and `ignore` aren't a part of the REST API spec anywhere as far as I can tell and so shouldn't appear in `query_params()`, am I mistaken with that?

sethmlarson:
This PR isn't needed any more, it's not valid to have `query_params("request_timeout")` and if that's found anywhere it should be fixed. It looks like that issue is fixed in the 7.x and 6.x branches.

Issue #943, opened 2019-04-30
urllib3 1.25 breaks ssl verification when verify_certs=False


jrobison-sb:
Our code:

```
Elasticsearch(os.environ['ELASTICSEARCH_URL'],
                             verify_certs=False, timeout=60)
```

When our underlying pip3 package of `urllib3` is at 1.24.2, the above works great.

When we use `urllib3` 1.25, it breaks with the following error:
```
ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:720)
```

Considering that urllib3 [changed the way that SSL verification works by default](https://github.com/urllib3/urllib3/blob/master/CHANGES.rst) in 1.25, I suspect that either Elasticsearch isn't correctly passing `verify_certs=False` to urllib3, or urllib3 isn't honoring it, or something...

I don't know for sure if this is an elasticsearch-py problem or a urllib3 problem, but since my goal is to use Elasticsearch, I figured I should start here. Thanks.

OS: Debian Stretch, in a docker container
Python: 3.5.3
Elasticsearch python package: 7.0.0
Elasticsearch server: 6.2, using AWS's managed Elasticsearch service

vtermanis:
Shameless bump - hopefully addressed by #987

jolyj:
I had the problem here, the only fix was to downgrade to 1.24.2 to make urllib3 ignore certificate verification

xNinjaKittyx:
I had this problem with requests library. It completely ignored my verify=False parameter in requests.

I was on 1.25.4, but had to revert to 1.25.3 for it to work properly again.

mlcdf:
> I had the problem here, the only fix was to downgrade to 1.24.2 to make urllib3 ignore certificate verification

Thank for the workaround!

vtermanis:
@HonzaKral - is [the fix](https://github.com/elastic/elasticsearch-py/commit/4c531c5cdbd5ba8b59be16c2fa15d5cf81d0a92f) going to be backported to the ES6 branch of this package?

mccarthyp-snet:
I downgraded urllib3 to 1.24.2 and I'm still getting `sslv3 alert bad certificate`.

Python: 3.8.0
Python client for Elasticsearch: 7.6.0
Elasticsearch: 6.2.3

sanand0:
[...]

PR #987, opened 2019-07-26
Fix verify_certs=False behaviour for urllib3>=1.25

vtermanis:
(Addresses #943)

Since [urllib3 1.25](https://github.com/urllib3/urllib3/blob/1.25.3/CHANGES.rst#125-2019-04-22) SSL certificate verification is enabled by default (https://github.com/urllib3/urllib3/pull/1507). This currently makes it impossible to use `verify_certs=False` via elasticsearch (unless one specifies one's own `SSLContext`).

This change explicitly sets `cert_reqs=CERT_NONE` when using `verify_certs=False`.

### DIFF
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -184,6 +184,7 @@ class Urllib3HttpConnection(Connection):
                     }
                 )
             else:
+                kw.update({"cert_reqs": "CERT_NONE"})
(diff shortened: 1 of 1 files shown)


vtermanis:
FYI - I've just signed the CLA (with my github usename but different email address than in commit).

HonzaKral:
This was already resolved in 4c531c5
closing the PR

PR #991, opened 2019-08-07
Send User-Agent header identifying the Client

philkra:
This PR adds to every request the `user-agent` header with the string `elasticsearch-py/x.y.z`, with `x.y.z` being the Client release version, in order to identify a Client on the Elasticsearch side.

### DIFF
--- a/elasticsearch/transport.py
+++ b/elasticsearch/transport.py
@@ -1,5 +1,6 @@
 import time
 from itertools import chain
+from platform import python_version
 
 from .connection import Urllib3HttpConnection
 from .connection_pool import ConnectionPool, DummyConnectionPool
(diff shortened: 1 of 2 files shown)


elasticsearch/transport.py
```diff
@@ -335,6 +337,10 @@ def perform_request(self, method, url, headers=None, params=None, body=None):
             if isinstance(ignore, int):
                 ignore = (ignore,)
 
+        if headers is None:
+            headers = {}
+        headers["user-agent"] = "elasticsearch-py/%s (Python %s)" % (__versionstr__, python_version())
```
HonzaKral: this will always override the value in `headers` even if the user supplied their own, is that really what we want?

philkra: That's a good question! I'll check in with other the other clients. What are your 2 cents on this? Did you came across a setup where `user-agent` was actively populated?

philkra: Added PR #992 to address this.

PR #992, opened 2019-08-12
Ensure a custom User-Agent header is not overwritten

philkra:
The PR is an enhancement to the [comment](https://github.com/elastic/elasticsearch-py/pull/991/files/4a355cdbc526e586672fe1d506e52ff92bb32dfe#r312094064) of PR#991 and ensures that a custom `user-agent` header is not overwritten with the default value.

### DIFF
--- a/elasticsearch/transport.py
+++ b/elasticsearch/transport.py
@@ -336,12 +336,6 @@ class Transport(object):
             ignore = params.pop("ignore", ())
             if isinstance(ignore, int):
                 ignore = (ignore,)
-
-        if headers is None:
-            headers = {}
-        if "user-agent" not in headers:
-            headers["user-agent"] = "elasticsearch-py/%s (Python %s)" % (__versionstr__, python_version())
-
         for attempt in range(self.max_retries + 1):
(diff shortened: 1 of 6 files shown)


elasticsearch/transport.py
```diff
@@ -339,7 +339,8 @@ def perform_request(self, method, url, headers=None, params=None, body=None):
 
         if headers is None:
             headers = {}
-        headers["user-agent"] = "elasticsearch-py/%s (Python %s)" % (__versionstr__, python_version())
+        if "user-agent" not in headers:
+            headers["user-agent"] = "elasticsearch-py/%s (Python %s)" % (__versionstr__, python_version())
```
HonzaKral: Could we perhaps not do this at every request? I think this should follow the same exact logic as setting `content-type` at https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/http_urllib3.py#L129 - done only once when instantiating the client and kept there as the default. Also all the headers will then be set in a single place.

philkra:
my pleasure :)

PR #999, opened 2019-08-21
Fix _broken_ Bulk Helper test for py2.7

philkra:
This PR tries to isolate and fix the failing test: `test_elasticsearch.test_helpers.TestParallelBulk`. It has to be noted that this test only fails on python `v2.7`

### DIFF
--- a/test_elasticsearch/test_helpers.py
+++ b/test_elasticsearch/test_helpers.py
@@ -32,10 +32,12 @@ class TestParallelBulk(TestCase):
         side_effect=mock_process_bulk_chunk,
     )
     def test_all_chunks_sent(self, _process_bulk_chunk):
-        actions = ({"x": i} for i in range(100))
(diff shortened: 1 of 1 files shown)


HonzaKral:
What failure is this trying to fix? Aside from changing the size from `2` to `4` it doesn't change any logic. Not to mention that the logic of counting the number of calls required would break in `2.7` where `/` on ints doesn't have to be int

HonzaKral:
Oh, I see - it is trying to fix the intermittent failures we gen on testing parallel bulk...

The problem there is that `Mock.call_count` is not thread safe so it can lead to wrong results. We need to change the logic used to checking the calls made to something that can be incremented atomically like appending to a list or something.

Closing this as it is not valid resolution for the issue, will open a ticket to track the problem

philkra:
I figured this would not be the solution and some thread safety culprit 😄 thanks for opening the correct issue.

PR #1010, opened 2019-09-04
API key Authentication

philkra:
This PR adds a simplified usage of Elasticsearch's API Key authentication.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -247,7 +247,7 @@ Please note this authentication method has been introduced with release of Elast
     # or you pass the base 64 encoded token
     es = Elasticsearch(
         ['node-1', 'node-2', 'node-3'],
-        api_key=('base64encoded tuple'),
+        api_key='base64encoded tuple',
     )
 
 .. _API Key: https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -230,6 +230,28 @@ description of the options.
 
 .. _certifi: http://certifiio.readthedocs.io/en/latest/
 
+APIKey Authentication
+~~~~~~~~~~~~~~~~~~~~~~
+
+You can configure the client to use Elasticsearch's `API Key`_ for connecting to your cluster.
+Please note this authentication method has been introduced with release of Elasticsearch ``6.7.0``.
+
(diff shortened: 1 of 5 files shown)


docs/index.rst
```diff
@@ -230,6 +230,28 @@ description of the options.
...
+    # you can use the api key tuple
+    es = Elasticsearch(
+        ['node-1', 'node-2', 'node-3'],
+        api_key=('id', 'api_key'),
+    )
+
+    # or you pass the base 64 encoded token
+    es = Elasticsearch(
+        ['node-1', 'node-2', 'node-3'],
+        api_key=('base64encoded tuple'),
```
HonzaKral: can we remove the `(` and `)` here? it is a bit confusing as it seems to suggest it is still a tuple when it's just a string

elasticsearch/connection/base.py
```diff
@@ -9,6 +10,7 @@
 
 from ..exceptions import TransportError, HTTP_EXCEPTIONS
 from .. import __versionstr__
+from ..compat import string_types
```
HonzaKral: unused import it seems

elasticsearch/connection/http_requests.py
```diff
@@ -80,6 +82,8 @@ def __init__(
             elif isinstance(http_auth, string_types):
                 http_auth = tuple(http_auth.split(":", 1))
             self.session.auth = http_auth
+        if api_key is not None:
+            self.session.headers.setdefault('authorization', self._get_api_key_header_val(api_key))
```
HonzaKral: I don't think we should be using `setdefault` here as it might be confusing when the `api_key` parameter will be ignored if `authorization` is already in `headers`.

Either override the value (since it is an explicit request of the user to pass `api_key`) or raise an exception if there already is an `authorization` header present.

docs/index.rst
philkra: ++, good call!

elasticsearch/connection/base.py
philkra: yes, forgot to remove that one, thanks

elasticsearch/connection/http_requests.py
philkra: I opted to override the `authorization` header if `api_key` is set.

PR #1021, opened 2019-09-27
Clear all scroll_ids for one single scan()

XD-DENG:
## Background

According to [scroll documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-scroll): 
"The initial search request and each subsequent scroll request each return a _scroll_id. While **the _scroll_id may change between requests,** it doesn’t always change — in any case, only the most recently received _scroll_id should be used."

This means that there may be multiple scrolls to be created within one single scan, while current implementation only clears the last scroll.

For example:
- Step-1: initial search with scroll, returns 1st batch of results and `scroll_id_1`
- Step-2: use `scroll_id_1` to call `/scroll`, returns 2nd batch of results and `scroll_id_2`
- Step-3, use `scroll_id_2` to call `/scroll`, returns 3rd batch (the last batch) of results and `scroll_id_3`.
- Step-4, further request to `/scroll` will not give me any valid `[hits]`, so I will stop.

Currently implementation only clear scroll of id `scroll_id_3`.

## My Question (inputs needed)

The help I need from Elasticsearch experts is: will scroll of `scroll_id_1` be cleared automatically when new scroll id `scroll_id_2` is created? The same, will scroll of `scroll_id_2` be cleared automatically when new scroll id `scroll_id_3` is created?

- If the answer is `no`, we should explicitly clear all `scroll_id_1`, `scroll_id_2` and `scroll_id_3`, and this PR should be considered;
- If the answer is `yes`, current implementation would suffice (we only need to clear the latest scroll_id) and please feel free to close this PR.

## Why Should We Care

"_Search context are automatically removed when the scroll timeout has been exceeded. However keeping scrolls open has a cost, as discussed in the previous section so scrolls should be explicitly cleared as soon as the scroll is not being used anymore_" ([reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#_clear_scroll_api))

### DIFF
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -430,11 +430,14 @@ def scan(
         query = query.copy() if query else {}
         query["sort"] = "_doc"
 
+    scroll_ids = set()
     # initial search
     resp = client.search(
         body=query, scroll=scroll, size=size, request_timeout=request_timeout, **kwargs
     )
     scroll_id = resp.get("_scroll_id")
+    if scroll_id:
+        scroll_ids.add(scroll_id)
 
     try:
         while scroll_id and resp["hits"]["hits"]:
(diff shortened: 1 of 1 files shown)


XD-DENG:
Hi @philkra , do you mind taking a look at this PR? As well as https://github.com/elastic/elasticsearch-py/pull/1025 . Thanks.

HonzaKral:
> This means that there may be multiple scrolls to be created within one single scan, while current implementation only clears the last scroll.

This is not really the case. If the scroll id changes the previous one is invalid and doesn't have to be cleaned up. Only the last scroll id ever requires clearing.

XD-DENG:
Thanks @HonzaKral . Good learning for me.

PR #1034, opened 2019-10-09
Include simplejson in extras

eric-weaver:
elasticsearch-py has support [here](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/serializer.py#L2) for [simplejson](https://github.com/simplejson/simplejson) if it's installed but does not currently include this dependency as an optional extra in the setup file.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,5 +59,6 @@ setup(
     extras_require={
         "develop": tests_require + ["sphinx<1.7", "sphinx_rtd_theme"],
         "requests": ["requests>=2.4.0, <3.0.0"],
+        "simplejson": ["simplejson"],
     },
 )


sethmlarson:
Simplejson will be automatically detected if the user installs the dependency and doesn't change the functionality of the module in any way except a modest speed increase.

eric-weaver:
@sethmlarson The same can be said for requests lib but that's included in the extras field. This is one of the [recommended](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) ways to list optional dependencies. 

I only raise this next point because I found this simplejson dependency in one of my services and removing it did cause breaks to tests. It took quite some time to track down where this package was being used since it was not included in any of the normal expected locations. At least on the 2.x branch of this library simplejson also has the side effect of properly serializing json data that has unicode characters that are not handle automatically by pythons default json lib.

sethmlarson:
I interpret `extras_require` as a way to add extra functionality to the module versus "optional sub-dependencies". The `elasticsearch[requests]` extra adds `RequestsHttpConnection` whereas no new functionality is locked behind whether `simplejson` is installed or not.

I'm interested in knowing the JSON you found that serialized improperly so I can test against other branches, could you provide an example to reproduce the issue?

PR #1036, opened 2019-10-13
Modify the `allow_no_indices` docstring

### LINKED ISSUE #988: es.indices.exists allow_no_indices default is misleading
Hi.

When using es.indices.exists, the default value for allow_no_indices is documents to be False. It seems like the default value is actually True. See my exam below:
`
from elasticsearch6 import Elasticsearch
es = Elasticsearch([{"host" : "localhost"}])
es.indices.create("is_bug-a")
es.indices.create("is_bug-b")

## Says allow_no_indices default = False
es.indices.exists("is_bug_*") # Returns True

## Setting to false gives different result
es.indices.exists("is_bug_*", allow_no_indices=False) # Returns False

## Setting to True gives same result
es.indices.exists("is_bug_*", allow_no_indices=True) # Return True

`

elasticsearch6 version -> 6.4.2

tzabal:
The docstring of the `allow_no_indices` parameter in the methods `get`, `delete`, and `exists`, of the `indices.IndicesClient` was misleading by reporting an incorrect default value (report false, when it was actually true).

This PR modifies the docstring of this parameter in the three aforementioned methods, by replacing it with the docstring of this parameter from the other methods of the codebase. The new one does not report a default value, because the elasticsearch-py library does not actually set a default in case of absence when performing the HTTP request.

Closes #988.

### DIFF
--- a/elasticsearch/client/indices.py
+++ b/elasticsearch/client/indices.py
@@ -122,8 +122,9 @@ class IndicesClient(NamespacedClient):
         `<http://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-index.html>`_
 
         :arg index: A comma-separated list of index names
-        :arg allow_no_indices: Ignore if a wildcard expression resolves to no
-            concrete indices (default: false)
+        :arg allow_no_indices: Whether to ignore if a wildcard indices
+            expression resolves into no concrete indices. (This includes `_all`
+            string or when no indices have been specified)
         :arg expand_wildcards: Whether wildcard expressions should get expanded
             to open or closed indices (default: open), default 'open', valid
             choices are: 'open', 'closed', 'none', 'all'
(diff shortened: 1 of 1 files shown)


Winterflower:
Thank you very much for your PR @tzabal !
I will take a look at the docstring and provide some feedback shortly. In the meantime, let's run the build process.

Winterflower:
Jenkins test this please

HonzaKral:
Thank you very much for the patch. Unfortunately this is not something that lives in this repository as the code is generated from the official rest specifications stored in elasticsearch repo (0). Any modification to the description of parameters should then be done dirrectly at the source.

I apologize for the inconvenience

0 - https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/indices.delete.json#L39

PR #1038, opened 2019-10-17
HTTP 405 when node_id is defined

nfsec:
_make_path encodes "/" - so the request looks like: GET /_cluster%2Fstats%2Fnodes/node_id and elastic returns a 405 error.


HonzaKral:
Thank you very much for the patch, @nfsec !

Issue #958, opened 2019-05-17
Sniffing adds invalid connections in 7.x


mo-getter:
When transport sniffing is enabled, invalid host names are added to the connection pool, which causes "Name or service not known" errors. This is caused by the reintroduction of `<hostname>/<ip>:<port>` format, which is the default as of 7.x+. The old syntax--which the sniffer parsing still expects--is now [deprecated](https://github.com/elastic/elasticsearch/blob/e1fdd00420b226b33adba3675ac0ff5d6daf954c/server/src/main/java/org/elasticsearch/http/HttpInfo.java#L89).

Related to #925.

valentinomiazzo:
A workaround is to register a callback when you create the connection and do the fix there.
For example:

```
def _get_host_info(node_info, host):
        if node_info.get('roles', []) == ['master']:
            host = None
        if '/' in host.get('host'):
            host = {'host': host.get('host').rsplit('/', 1)[0], 'port': host.get('port')}
        return host

es = Elasticsearch(
                ...
                host_info_callback=_get_host_info,
            )
```

LoadingZhang:
Thx @valentinomiazzo ,
`if '/' in host.get('host')` should change to `elif '/' in host.get('host')`, else get error on master node.

HonzaKral:
Resolved by #1039

Issue #925, opened 2019-04-04
Add back support for hostname/ip:port parsing


fxdgear:
Previously Elasticsearch would return publish_address as `hostname/ip:port` if `http.publish_host` was set. This is coming back in future version of ES.

yogevyuval:
We are experiencing this too, any plans on fixing this?

HonzaKral:
Resolved by #1039

PR #1039, opened 2019-10-18
Support http.publish_host addresses

vEpiphyte:
Add support for ES instances where `http.publish_host` is set and the published address is in the `domain/ip:port` format.

- [x] Add unit test

### DIFF
--- a/elasticsearch/transport.py
+++ b/elasticsearch/transport.py
@@ -245,8 +245,16 @@ class Transport(object):
         if not address or ":" not in address:
             return None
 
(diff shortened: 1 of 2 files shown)


vEpiphyte:
This should also resolve #958

vEpiphyte:
This should also resolve #925

vEpiphyte:
Added a unit test to cover this behavior change.

HonzaKral:
Thank you very much for the patch! Much appreciated!

messense:
Can we backport this patch to 6.x?

Issue #1037, opened 2019-10-16
Issue with elasticsearch 7.4 instances and ids with spaces 


chrisbennight:
**Overview**:  
Issue is an interaction between existing url path encoding behaviour and changes made in 7.4 https://www.elastic.co/guide/en/elasticsearch/reference/7.x/breaking-changes-7.4.html which no longer interpret a + in a url path as a space.

**Details**: 
Id's with spaces in them are encoded with python's  quote_plus method:  https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L52  

This method (docs here:  https://docs.python.org/3/library/urllib.parse.html ) encodes spaces as the `+` character instead of `%20`.   When an id field contains a space previously (es < 7.4) these `+` symbols would be translated back to a space.

Since this behaviour has changed in es this now causes spaces in id's to be turned into the `+` character, which is then directly stored in elasticsearch.  

Note that since bulk methods don't have the id in the URL field they continue to work correctly - so indexing the same data with the single and bulk methods results in different id's.  

**Suggested Fix:**
I don't believe the use of the python `urllib.parse.quote_plus` is appropriate;  I believe just using `urllib.parse.quote`  here:  https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L52  will fix the issue.    The only difference between the two methods (see docs for both method:  https://docs.python.org/3/library/urllib.parse.html ) is the handling of spaces.   the `urllib.parse.quote` method encodes spaces with %20 instead of +,  which I believe should work as expected in es 7.4  (and I think should work correctly with earlier versions - though haven't confirmed).

chrisbennight:
PR Above, and CLA paperwork signed

HonzaKral:
This has been fixed by #1040, thank you again for the help!

PR #1040, opened 2019-10-18
Fix for 7.4 space encoding change

chrisbennight:
Ref: #1037 

Elastic-py currently encodes spaces in the url as `+` (as opposed to `%20`).   Elasticsearch in 7.4 no longer decode `+` as a space, so this breaks parameters (id's in my test) that have spaces.   

(Note the bulk API works correctly with ID's with spaces, so indexing is currently also divergent between the bulk and single api)

This PR uses the python quote method instead of quote_plus which is the same except for encoding spaces as `%20` instead of `+`.

### DIFF
--- a/elasticsearch/client/utils.py
+++ b/elasticsearch/client/utils.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals
 import weakref
 from datetime import date, datetime
 from functools import wraps
-from ..compat import string_types, quote_plus, PY2
+from ..compat import string_types, quote, PY2
 
 # parts of URL to be omitted
 SKIP_IN_PATH = (None, "", b"", [], ())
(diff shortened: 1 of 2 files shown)


elasticcla:
Hi @chrisbennight, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-py/pull/1040.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

elasticcla:
Hi @chrisbennight, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-py/pull/1040.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

chrisbennight:
CLA is signed, and email address matches now;  not sure how to re-trigger the CLA check

HonzaKral:
Unfortunately this does not work with python 2. @chrisbennight would you have some time to have a look into the Python 2 failures?

Thank you!

chrisbennight:
Mea culpa, oversight on my part;  works fine on py2 now (well, py2.7.x anyway, which I think matches the support matrix)

HonzaKral:
Thank you for the quick turnaround, @chrisbennight !

PR #1047, opened 2019-10-26
Adding gzip compression support to requests connection

kkoziara:
Currently gzip compression and `http_compress` parameter handling is only implemented for `Urllib3HttpConnection`, so when using alternative `RequestsHttpConnection` the parameter is ignored and the requests are not compressed, which might be unexpected for the users.
The PR adds gzip compression support to `RequestsHttpConnection` class. Compression is enabled by setting `http_compress` constructor parameter to `True` exactly as in the `Urllib3HttpConnection`.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -4,6 +4,7 @@ Changelog
 =========
 7.1.0 (dev)
 -----------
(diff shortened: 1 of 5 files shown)


sethmlarson:
Closing in favor of #1124.

PR #1048, opened 2019-10-29
Closes #1044. Fixing bug with mark_dead in connection pool

### LINKED ISSUE #1044: mark_dead() in ConnectionPool always removes the first connection in the pool
`ConnectionPool`'s `mark_dead()` method is supposed to remove the dead `connection` from the `connections` list:

`self.connections.remove(connection)`

However the equality comparison implementation `__eq__` in the `Connection` class is incorrect. It always returns `True` as long as both objects are `Connection` instances:

```
def __eq__(self, other):
        if not isinstance(other, Connection):
            raise TypeError(
                "Unsupported equality check for %s and %s" % (self, other)
            )
        return True
```

This defect manifests itself by removing all good connections until the connections is empty. One dead connection will bring down the whole connection pool.

Proposed solution: removing `__eq__` from `Connection`. Thoughts?

alexk307:


HonzaKral:
This is great, Could I just ask you to sign the [CLA](https://www.elastic.co/contributor-agreement/) so that we can merge it?

Thank you!

alexk307:
Just signed! Thanks

Issue #1026, opened 2019-10-02
fix typing of index creation timeout parameter


tomkralidis:
Hi: we use this awesome package as part of the [pygeoapi](https://pygeoapi.io) project.  Our builds started to [break](https://travis-ci.org/geopython/pygeoapi/jobs/592275226) with version 7.0.5 of the package.  The issue seems to be bound to [setting request timeouts](https://github.com/geopython/pygeoapi/blob/master/tests/load_es_data.py#L74).

Note that this works fine using 7.0.4.

HonzaKral:
This has been fixed in `master` by https://github.com/elastic/elasticsearch-py/pull/1051

PR #1049, opened 2019-10-29
Fixes `request_timeout` param. Closes #1035, #1026

### LINKED ISSUE #1035: delete request_timeout args error
elasticsearch-7.2 :
I before use elasticsearch-5.6 version use the .indices.delete function, use request_timeout args because delete often timeout, now I update to 7.2 version ,this args can't use it , always report this error:   "TypeError: a float is required" , 
try trace source code, find this function will change int or float to string type , i find thie code from 5.0 never change ,but not error
so this is bug or design in this way:
https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L12

if not bug ,how do I use it? 
thanks!

alexk307:
The urllib3 dependency was floating in [setup.py](https://github.com/elastic/elasticsearch-py/blob/master/setup.py#L14) such that the newest version of urllib3 would always be installed with this library. If urllib3 introduces breaking changes then our library is vulnerable to those changes as well. I froze the dep version in place so that it would remain at a stable place (we can change this if a newer version is desired)

I'm not quite sure if the urllib3 library introduced a breaking change where the `request_timeout` parameter could no longer be a string, or our library internally introduced this issue. It's hard to tell because the unit tests are not actually using urllib3, but instead a [DummyConnection](https://github.com/elastic/elasticsearch-py/blob/master/test_elasticsearch/test_transport.py#L13) This attempts to set the param by trying to convert it to a float.

I have tested this locally with the examples in the #1026

```python
es = Elasticsearch()
es.indices.create(index='test', body={}, request_timeout=90)
``` 

Before the patch this would break as `request_timeout` was being converted to a `str` [here](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L41). My patch looks for the `request_timeout`.

### DIFF
--- a/elasticsearch/connection/base.py
+++ b/elasticsearch/connection/base.py
@@ -65,7 +65,7 @@ class Connection(object):
             raise TypeError(
                 "Unsupported equality check for %s and %s" % (self, other)
             )
-        return True
+        return self.__hash__() == other.__hash__()
 
     def __hash__(self):
         return id(self)
--- a/elasticsearch/connection_pool.py
+++ b/elasticsearch/connection_pool.py
@@ -150,6 +150,7 @@ class ConnectionPool(object):
         try:
             self.connections.remove(connection)
         except ValueError:
+            logger.info("Attempted to remove %r, but it does not exist in the connection pool.", connection)
             # connection not alive or another thread marked it already, ignore
             return
         else:
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -442,17 +442,18 @@ def scan(
                 yield hit
 
             # check if we have any errors
-            if resp["_shards"]["successful"] < resp["_shards"]["total"]:
+            if (resp["_shards"]["successful"] + resp["_shards"]["skipped"]) < resp["_shards"]["total"]:
                 logger.warning(
-                    "Scroll request has only succeeded on %d shards out of %d.",
+                    "Scroll request has only succeeded on %d (+%d skipped) shards out of %d.",
                     resp["_shards"]["successful"],
+                    resp["_shards"]["skipped"],
                     resp["_shards"]["total"],
                 )
                 if raise_on_error:
                     raise ScanError(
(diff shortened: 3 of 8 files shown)


elasticcla:
Hi @alexk307, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-py/pull/1049.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

alexk307:
Apologies for the weird commit history, I was trying to get my verified email to show up (finally). I can rebase or squash if this is too messy.

HonzaKral:
I am not sure this patch is needed - you see in https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L80-L83 that `_escape` should never be called on `request_timeout` parameter, it is only called on the declared parameters or the API methods as defined here: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L74

Either it is a non issue or we are trying to solve the wrong thing with this PR.

alexk307:
The method I am testing on is the create index function which accepts the `request_timeout` parameter in the decorator here: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/indices.py#L84

That will call the `_escape` method here: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L74

The `pop` to remove the request_timeout param is done below the `_escape`call here: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L81. Maybe we should re-order them such that it pops out those parameters before it tries to escape them?

HonzaKral:
Thank you for the pointer, in that case the proper fix for the error is to remove the `request_timeout` from that list and from any other decorator where it is used as it really should not be defined anywhere.

alexk307:
Closing in favor of a new PR

PR #1050, opened 2019-10-30
declare python 3.7 support

rooterkyberian:


HonzaKral:
Thank you for the patch, much appreciated!

PR #1051, opened 2019-10-30
Removes request_timeout params from indices handler.  Closes #1035, #1026

### LINKED ISSUE #1035: delete request_timeout args error

alexk307:
Follow up from the closed discussion here: https://github.com/elastic/elasticsearch-py/pull/1049

### DIFF
--- a/elasticsearch/client/indices.py
+++ b/elasticsearch/client/indices.py
@@ -81,7 +81,6 @@ class IndicesClient(NamespacedClient):
     @query_params(
         "master_timeout",
         "timeout",
-        "request_timeout",
         "wait_for_active_shards",
         "include_type_name",
     )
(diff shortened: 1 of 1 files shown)


alexk307:
There are still references to `request_timeout` in the docs such as [here](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/indices.py#L97).

The param itself will still work with respect to the transport layer, wondering if we should remove it from the docs or just keep it?

HonzaKral:
It should be removed from the docstrings as well as those only list API specific params. There is a separate section in the docs for global params like `ignore` and `request_timeout`

PR #1056, opened 2019-11-07
Align order of arguments to put_mapping() in docs with code

iandanforth-alation:
The current docs appear to have the arguments out of order, let's make this line up with the code.


HonzaKral:
Thank you for the patch, @iandanforth-alation, I appreciate you taking the time to submit it!

PR #1062, opened 2019-11-21
Support both deprecated and new bulk parameters

### LINKED ISSUE #1020: Bulk on 7.0.4 does not support `version` and `version_type` fields
As part of upgrading our cluster to Elasticsearch v7.1, we updated the elasticsearch library to v7.0.4. Unfortunately, we encountered a regression in the `bulk` function when changing our `_version` and `_version_type` action properties to `version` and `version_type`, respectively, as suggested [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#_camel_case_and_underscore_parameters_deprecated_in_6_x_have_been_removed).

The hardcoded whitelist referred to in this [old issue](https://github.com/elastic/elasticsearch-py/issues/564) is to blame. We have worked around this for now by overriding the `expand_action_callback` parameter on `bulk` but this feels like something the library should support out of the box.

thomasthiebaud:
Fixes #1020

### DIFF
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -27,22 +27,26 @@ def expand_action(data):
     op_type = data.pop("_op_type", "index")
     action = {op_type: {}}
     for key in (
+        "_id",
         "_index",
         "_parent",
         "_percolate",
+        "_retry_on_conflict",
         "_routing",
         "_timestamp",
-        "routing",
         "_type",
         "_version",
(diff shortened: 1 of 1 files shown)


thomasthiebaud:
I approved the CLA, but it does not seems to be visible here 🤔

marshallmain:
This looks good to me.  Thanks for the contribution!

PR #1072, opened 2019-12-04
Added `bytes` params for nodes function

shlem952:
Added `bytes` param for nodes function
https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-nodes.html


sethmlarson:
Closing this PR as it's been solved by auto-generating the API functions.

PR #1073, opened 2019-12-09
Updated requirement specifier.

toanant:
Updated as per pypi's [requirement-specifiers](https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers) doc, also defined at - [PEP 508](https://www.python.org/dev/peps/pep-0508/)
Keeping it simple will avoid causes for confusion like - https://www.zdnet.com/article/two-malicious-python-libraries-removed-from-pypi/


chason:
This looks correct to me.

PR #1076, opened 2019-12-10
ssl_show_warn option improvements

### LINKED ISSUE #1075: ssl_show_warn option improvements
* No ssl_show_warn accepted when a ssl_context is provided
* Disable InsecureRequestWarning from urllib3's connectionpool with ssl_show_warn option

alvarolmedo:
* No ssl_show_warn accepted when a ssl_context is provided
* Disable InsecureRequestWarning from urllib3's connectionpool with ssl_show_warn option

(Closes #1075)

### DIFF
--- a/elasticsearch/connection/http_requests.py
+++ b/elasticsearch/connection/http_requests.py
@@ -103,6 +103,10 @@ class RequestsHttpConnection(Connection):
                 )
             self.session.verify = ca_certs
 
+        if not ssl_show_warn:
+            requests.packages.urllib3.disable_warnings()
+
+        self.ssl_show_warn = ssl_show_warn
         if self.use_ssl and not verify_certs and ssl_show_warn:
             warnings.warn(
                 "Connecting to %s using SSL with verify_certs=False is insecure."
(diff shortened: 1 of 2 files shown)


elasticsearch/connection/http_requests.py
```diff
@@ -118,6 +119,8 @@ def perform_request(
 
         start = time.time()
         request = requests.Request(method=method, headers=headers, url=url, data=body)
+        if not self.ssl_show_warn:
+            requests.packages.urllib3.disable_warnings()
```
HonzaKral: Hi, thank you for the patch, could you please move this to `__init__`, or is there a reason why it is here? This way we run it with every request which is not ideal. Thank you!

alvarolmedo: Make sense, moved to __init__ and commits amended.

HonzaKral:
Thank you very much, @alvarolmedo , for the patch and the quick reaction!

alvarolmedo:
You are welcome :)
Nice to see my code in the next release. it will be very useful for us.

PR #1081, opened 2019-12-20
Fix params=None in client.search()

Godron629:
Closes https://github.com/elastic/elasticsearch-py/issues/1082

* Default arg is None yet params is iterated over immedately in body
* Add check for None and redefine as {}


HonzaKral:
this is a non issue as the default value is handled by the decorator

PR #1086, opened 2020-01-06
add single test run instructions to docs

marshallmain:
As a non-expert with python setuptools I had to spend some time figuring out how to run a single test instead of the full suite.  This change provides a sample in the docs that shows how to do this for future maintainers.

### DIFF
--- a/test_elasticsearch/README.rst
+++ b/test_elasticsearch/README.rst
@@ -39,6 +39,9 @@ Alternatively, if you wish to control what you are doing you have several additi
  * you can just run your favorite runner in the ``test_elasticsearch`` directory
    (verified to work with nose and py.test) and bypass the fetch logic entirely.
 
+ * to run a specific test, you can use ``python3 setup.py test -s <test_name>``, for example
+   ``python3 setup.py test -s test_elasticsearch.test_helpers.TestParallelBulk.test_all_chunks_sent``
+
 To install all test dependencies you can also run ``pip install -e .[develop]``.
 
 Run Elasticsearch in a Container


test_elasticsearch/README.rst
```diff
@@ -39,6 +39,9 @@ Alternatively, if you wish to control what you are doing you have several additi
  * you can just run your favorite runner in the ``test_elasticsearch`` directory
    (verified to work with nose and py.test) and bypass the fetch logic entirely.
 
+ * to run a specific test, you can use ``python3 setup.py test -s <test_name>``, for example
+   ``python3 setup.py test -s test_elasticsearch.test_helpers.TestParallelBulk.test_all_chunks_sent``
```
HonzaKral: it might be better to specify how to work with the `run_tests.py` which allows more option and is faster as it skips the initialization of the project which is only needed once (and not needed if you do `pip install -e '.[develop]'`). You can run individual files and/or tests, only previously failed tests etc (see https://nose.readthedocs.io/en/latest/man.html for details)

marshallmain: i haven't been able to run run_tests.py on its own, when I try i get 
`Traceback (most recent call last):
  File "test_elasticsearch/run_tests.py", line 85, in <module>
    run_all(sys.argv)
  File "test_elasticsearch/run_tests.py", line 65, in run_all
    fetch_es_repo()
  File "test_elasticsearch/run_tests.py", line 38, in fetch_es_repo
    from test_elasticsearch.test_server import get_client
ImportError: No module named test_elasticsearch.test_server`
However, it works for me when called through setup.py.  Do you know why this might happen?

HonzaKral: How have you installed the library, did you run the `pip` command?

marshallmain: ah running `pip install -e .[develop]` fixed it, I had switched dev machines and forgot to re-run it.

PR #1087, opened 2020-01-07
Use side effect call count instead of mock call count

marshallmain:
Improves upon https://github.com/elastic/elasticsearch-py/pull/875

The flaky test failures prior to this PR can be reliably reproduced with Python 2.7.17 by modifying `test_all_chunks_sent` to create 100000 documents instead of 100 and using 1000 threads in `parallel_bulk` instead of the default of 4.  Local tests under these conditions typically had `_process_bulk_chunk.call_count` ~= 49990 instead of the correct value of 50000.  

The flaky test failures were not reproducible using Python 3.7.5.

This PR fixes the issue in Python 2 by using the mock_process_bulk_chunk.call_count instead, which is updated with a lock and therefore thread safe.


HonzaKral:
Thank you for the fix, much appreciated!

PR #1089, opened 2020-01-09
Adds Python client suppport for Elasticsearch 7.5.1

viglia:
Python client for Elasticsearch 7.5.1

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -3,6 +3,10 @@
 Changelog
 =========
 
+7.5.1 (2020-01-09)
+------------------
+  * Update client to support ES 7.5.1 APIs
+
 7.1.0 (2019-11-14)
 ------------------
   * Fix sniffing with ``http.publish_host``
--- a/elasticsearch/__init__.py
+++ b/elasticsearch/__init__.py
@@ -1,7 +1,7 @@
 # flake8: noqa
 from __future__ import absolute_import
 
-VERSION = (7, 1, 0)
+VERSION = (7, 5, 1)
 __version__ = VERSION
 __versionstr__ = ".".join(map(str, VERSION))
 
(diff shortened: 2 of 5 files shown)


viglia:
@HonzaKral , thanks for the feedback. 

When you say "please do not exactly copy the ES version" do you mean in the `setup.py` and `__init__.py` or that I shouldn't generate from tag linking to patch version.
Should I generate from 7.5.0 instead of 7.5.1 ?

HonzaKral:
yes, `7.5.0` or even `7.2.0` which is the next version in the line of python versions.

viglia:
@philkra please can you clarify and give me a feedback? I've been told to skip [7.2...7.4] and jump to 7.5.1.

HonzaKral:
If you wish to stick to the version, go ahead, but I don't like committing to something that is not sustainable - if we have to do a bugfix release afte `7.5.1` we will have to release `7.5.2` when then Elasticsearch releases `7.5.2` we will be out of sync and there is nothing we can do. At that point I prefer to be explicit about the (possible) divergence. But up to @philkra of course

philkra:
@viglia , please use `7.5.0`. @HonzaKral explained our motives perfectly, we "only" follow up to the `major`.`minor` of Elasticsearch, giving us room to provide fixes in patch versions if necessary. Thanks @HonzaKral and @viglia for tackling `7.5.0`.

viglia:
@HonzaKral @philkra ok, perfect. Then if everyone is fine with this I will close this PR and regenerate pointing to ES 7.5.0 and create a new PR

HonzaKral:
You should keep using `7.5.1` version of elasticsearch, just the version of `elasticsearch-py` should be `7.5.0`

philkra:
thank you @HonzaKral . That's the way to go @viglia .

PR #1114, opened 2020-02-25
Migrate CI to Jenkins on 7.x branch

sethmlarson:
Based mostly on `elasticsearch-ruby`'s configuration but adapted for Python.

Trying on the 7.x branch first as the test suite doesn't pass against the 8.0 Docker image on `master`. Will migrate `master` to Jenkins once I 8.0-ify that branch.

Also using this to test if Jenkins will pick up this PR, is there any way that can be tested before we merge this?


elasticsearch/serializer.py
```diff
@@ -37,7 +50,10 @@ def default(self, data):
 
     def loads(self, s):
         try:
-            return json.loads(s)
+            obj = json.loads(s)
+            if self._is_repl:
+                return _PrettyJSONDict(obj)
```
HonzaKral: This will fail for APIs that return a `list`, for example `es.cat.nodes(format='json')`

test_elasticsearch/test_server/test_common.py
```diff
@@ -108,10 +108,20 @@ def tearDown(self):
...
+            try:
+                xinfo = self.client.xpack.info()
+                XPACK_FEATURES = set(
+                    f for f in xinfo["features"] if xinfo["features"][f]["enabled"]
+                )
+            except RequestError as e:
+                # We receive a 'RequestError' here when using 'oss' because the request
+                # for xpack.info() looks like this: 'GET /_xpack' which errors on the
+                # oss container due to having a similar route to 'GET /<index>'.
+                if "invalid_index_name_exception" not in str(e):
```
HonzaKral: you should be able to check `e.error` directly instead of `str(e)`

test_elasticsearch/run_tests.py
```diff
@@ -21,10 +21,10 @@ def fetch_es_repo():
 
     # no repo
     if not exists(repo_path) or not exists(join(repo_path, ".git")):
-        print("No elasticsearch repo found...")
-        # set YAML DIR to empty to skip yaml tests
-        environ["TEST_ES_YAML_DIR"] = ""
-        return
```
HonzaKral: this option was useful to avoid running the rest tests as they take a long time, this will mean that a typo in a setting will cause the entire ES repo to be cloned which is a very expensive process

.ci/run-elasticsearch.sh
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env bash
+#
+# Launch one or more Elasticsearch nodes via the Docker image,
+# to form a cluster suitable for running the REST API tests.
+#
+# Export the ELASTICSEARCH_VERSION variable, eg. 'elasticsearch:8.0.0-SNAPSHOT'.
+
+# Version 1.0.2
```
HonzaKral: this file should not live in this repo if it is shared by multiple clients, instead it should be in the shared repository somewhere to avoid duplication. It is also unnecessarily complex for the needs of the tests. To run the server for x-pack tests all you need is:

```
docker run \
    --publish 9200:9200 \
    --env "xpack.security.authc.token.enabled=true" \
    --env "xpack.security.authc.api_key.enabled=true" \
    --env "xpack.security.enabled=true"  \
    --env "ELASTIC_PASSWORD=changeme" \
    --env "xpack.license.self_generated.type=trial" \
    --env "discovery.type=single-node" \
    --env "path.repo=/tmp" \
    --env "repositories.url.allowed_urls=http://*" \
    --env "node.attr.testattr=test"  \
    --env "node.name=test"  \
    docker.elastic.co/elasticsearch/elasticsearch:$VERSION
```
(this does not include SSL which can bea easily added).

For non x-pack tests it is
```
docker run --detach --rm --publish 9200:9200 --env "discovery.type=single-node" -e path.repo=/tmp -e "repositories.url.allowed_urls=http://*" -e node.attr.testattr=test  -e node.name=test  docker.elastic.co/elasticsearch/elasticsearch:$VERSION
```

not sure why this entire infrastructure is needed and what it adds for the complexity.

elasticsearch/serializer.py
sethmlarson: Oh this wasn't meant to be in this PR, only meant for my own usage working through the "Getting Started" documentation. :) Will revert these changes.

test_elasticsearch/run_tests.py
sethmlarson: That's true, I'll add back a way to disable running API tests. Will also apply that flag onto the `oss` Jenkins CI config.

HonzaKral: the rest tests that are running now are for the `oss` release

.ci/run-elasticsearch.sh
Mpdreamz: For now this does need to live in this repository as well. The added complexity is in part to support running in different modes. e.g all the clients can run the yaml tests using:

```
$ ELASTICSEARCH_VERSION=7.x-SNAPSHOT TEST_SUITE=xpack ./.ci/run-tests 
```

Or choose to run `run-elasticsearch` separately and (using DETACH=true or not) during development. 

After we get everyone in this harness we'll explore our options to move parts of this into a shared space.

.ci/Dockerfile
```diff
@@ -0,0 +1,8 @@
+ARG PYTHON_VERSION=3.8
+FROM python:${PYTHON_VERSION}
+
+WORKDIR /code/elasticsearch-py
+
+COPY . .
```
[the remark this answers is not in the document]
Mpdreamz: A https://github.com/elastic/eland/blob/master/.dockerignore file could further help speed up the spin up of the container.

sethmlarson:
Now that the jobs are showing up in Jenkins, going to cycle this PR to see if we can get a build.

sethmlarson:
Now that the jobs are showing up in Jenkins, going to cycle this PR to see if we can get a build.

.ci/Dockerfile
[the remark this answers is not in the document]
sethmlarson: Problem is installing `.` requires many files currently, could split it out into a `dev-requirements.txt` though for sure. I'll add a .dockerignore file. :)

Issue #1115, opened 2020-02-25
A wrong warning is shown when creating a secured elasticsearch ssl (since 7.5.0)


yogevyuval:
**Elasticsearch version** not relevant:

**`elasticsearch-py` version 7.5.1**: 

**When creating a secured elasticsearch ssl connection a warning is shown**:
```
http_urllib3.py:148: UserWarning: When using `ssl_context`, all other SSL related kwargs are ignored
  "When using `ssl_context`, all other SSL related kwargs are ignored"
```

This happens because the new `ssh_show_warn` parameter defaults to `True` but also is in the or statement that shows the above error. 

The warning is wrong, because we are only setting the ssl_context parameter and nothing else.

sethmlarson:
Yep this looks incorrect, I'll make a PR to fix the issue.

PR #1116, opened 2020-02-25
Fix wrong ssl warning

yogevyuval:
Resolves: https://github.com/elastic/elasticsearch-py/issues/1115

The ssl_show_warn defaults to True, making this warning to be printed even in situations where no extra ssl parameter has been set.

The fix is to check for the extra parameter **and** that ssl_show_warn is True

### DIFF
--- a/elasticsearch/connection/http_urllib3.py
+++ b/elasticsearch/connection/http_urllib3.py
@@ -142,7 +142,7 @@ class Urllib3HttpConnection(Connection):
             or client_cert
             or client_key
             or ssl_version
-            or ssl_show_warn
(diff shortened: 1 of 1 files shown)


sethmlarson:
Closed in favor of #1117. FYI if you'd like to contribute to this library in the future you should sign the CLA.

yogevyuval:
@sethmlarson 
Thanks! I actually clicked the link and signed the agreement, not sure what else I should do to have it updated here.

sethmlarson:
Oh you signed the CLA, I missed the `cla-checker-service`'s comment that you signed! :)

PR #1122, opened 2020-02-27
Move alternative examples from 'example/' to 'docs/examples/'

sethmlarson:
Manually implemented some alternative examples for Elasticsearch docs, ran Black formatting on them so when there's a generator for the examples they won't be modified.

Deleted the old folder in favor of having under `docs/examples/`. Excluded that folder from Sphinx so RTD won't be affected.


philkra:
you need to open a PR in [conf.yaml](https://github.com/elastic/docs/blob/master/conf.yaml#L266) to change the path from `example/docs/asciidoc` to `docs/examples` and remove this (now) obsolete [item](https://github.com/elastic/docs/blob/master/conf.yaml#L268-L272), to get the build "green"  [job#48](https://elasticsearch-ci.elastic.co/job/elastic+docs+elasticsearch-py+pull-request/48/console)

PR #1124, opened 2020-03-02
Add HTTP compression to RequestsHttpConnection

sethmlarson:
- Adds `http_compress` support to `RequestsHttpConnection`
- Moves the gzip compression function into the base `Connection` class.
- Only runs compression when we're actually handed a body. Won't add the `Content-Encoding: gzip` header unless we run compression to not mess with Elastic Cloud proxies.
- Added a few unit tests specifically for headers


elasticsearch/connection/http_urllib3.py
```diff
@@ -230,17 +229,12 @@ def perform_request(
...
+            request_headers.update(headers or ())
+
             if self.http_compress and body:
-                try:
-                    body = gzip.compress(body)
-                except AttributeError:
-                    # oops, Python2.7 doesn't have `gzip.compress` let's try
-                    # again
-                    body = gzip.zlib.compress(body)
+                body = self._gzip_compress(body)
```
HonzaKral: this looks like it might interfere with the calls to `log_request_success` and `log_request_fail` below. Not a new bug if true, but probably worth fixing

PR #1126, opened 2020-03-03
Refactor cloud_id, api_key, headers, and http_compress

sethmlarson:
Moves the logic for `cloud_id`, `api_key`, `headers`, and `http_compress` into `Connection` so it's easier to ensure all clients behave the same.

This change feels a little hacky since there's an "assumption" that the super-class of `Connection` will define a `headers` parameter and if it doesn't then all the headers that are "set" within `Connection.__init__()` will be thrown away and replaced with whatever the super-class is doing, but that should be fine? I'm assuming most users either sub-class `Urllib3HttpConnection` and `RequestsHttpConnection` directly anyways.

### DIFF
--- a/elasticsearch/connection/base.py
+++ b/elasticsearch/connection/base.py
@@ -1,5 +1,4 @@
 import logging
-import base64
 import binascii
 import gzip
 import io
@@ -48,7 +47,7 @@ class Connection(object):
         url_prefix="",
         timeout=10,
         headers=None,
-        http_compress=False,
+        http_compress=None,
         cloud_id=None,
         api_key=None,
         **kwargs
@@ -66,7 +65,8 @@ class Connection(object):
             host = "%s.%s" % (es_uuid, parent_dn)
             port = 9243
             use_ssl = True
-            http_compress = True
+            if http_compress is None:
+                http_compress = True
 
         # Work-around if the implementing class doesn't
         # define the headers property before calling super().__init__()
@@ -91,7 +91,7 @@ class Connection(object):
             scheme = "https"
             use_ssl = True
         self.use_ssl = use_ssl
-        self.http_compress = http_compress
+        self.http_compress = http_compress or False
 
         self.hostname = host
         self.port = port
@@ -243,5 +243,5 @@ class Connection(object):
         """
         if isinstance(api_key, (tuple, list)):
             s = "{0}:{1}".format(api_key[0], api_key[1]).encode('utf-8')
-            return "ApiKey " + base64.b64encode(s).decode('utf-8')
+            return "ApiKey " + binascii.b2a_base64(s).rstrip(b"\r\n").decode('utf-8')
         return "ApiKey " + api_key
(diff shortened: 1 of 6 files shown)


elasticsearch/connection/base.py
```diff
@@ -38,20 +47,54 @@ def __init__(
...
+                parent_dn, es_uuid, _ = (
+                    binascii.a2b_base64(cloud_id.encode("utf-8")).decode("utf-8").split("$")
+                )
+            except ValueError:
+                raise ImproperlyConfigured("'cloud_id' is not properly formatted")
+
+            host = "%s.%s" % (es_uuid, parent_dn)
+            port = 9243
+            use_ssl = True
+            http_compress = True
```
HonzaKral: this doesn't seem ideal, it is (afaict) undocumented that we force compression on cloud connections and it is not always called for. `cloud_id` can also be used for local ECE installations where we might not want `http_compress`.

Maybe we can just set `http_compress` to `True` if it hasn't been set or just document that, when connecting to public cloud, it is a good idea to enable `http_compress`?

elasticsearch/connection/base.py
```diff
@@ -38,20 +47,54 @@ def __init__(
...
+
+        # Work-around if the implementing class doesn't
+        # define the headers property before calling super().__init__()
```
HonzaKral: I love this, very defensive, I like your pessimism!

elasticsearch/connection/base.py
```diff
@@ -38,20 +47,54 @@ def __init__(
...
-        :arg port: port to use (integer, default: 9200)
-        :arg url_prefix: optional url prefix for elasticsearch
-        :arg timeout: default timeout in seconds (float, default: 10)
-        """
+
+        if cloud_id:
+            try:
+                _, cloud_id = cloud_id.split(":")
+                parent_dn, es_uuid, _ = (
+                    binascii.a2b_base64(cloud_id.encode("utf-8")).decode("utf-8").split("$")
```
HonzaKral: we are using the `base64` wrapper in other places, maybe worth unifying on the low level `binascii`?

elasticsearch/connection/base.py
sethmlarson: This was added in a recent PR by me due to [this team issue](https://github.com/elastic/clients-team/issues/13) which recommended enabling compression, I'll add logic that doesn't override if the user sets an explicit value. I'll also add a tidbit in the cloud ID docs.

elasticsearch/connection/base.py
```diff
@@ -200,5 +243,5 @@ def _get_api_key_header_val(self, api_key):
         """
         if isinstance(api_key, (tuple, list)):
             s = "{0}:{1}".format(api_key[0], api_key[1]).encode('utf-8')
-            return "ApiKey " + base64.b64encode(s).decode('utf-8')
+            return "ApiKey " + binascii.b2a_base64(s).rstrip(b"\r\n").decode('utf-8')
```
sethmlarson: It's a bummer that Python 2 doesn't have a `newline` kwarg for `b2a_base64()` so we need to call `rstrip()` on the result.

sethmlarson:
Travis PR failed on downloading a package, all other Travis jobs passed. Merging this.

PR #1128, opened 2020-03-04
Update Cloud ID default port, parsing of Cloud ID

sethmlarson:
This started as a simpler PR until I discovered our Cloud ID parsing code doesn't match the Cloud ID spec. The basics of this PR are that port `9243` is no longer the standard and we should instead default to no port which results in port `443` being used implicitly by the HTTP libraries due to the scheme being `https`. We respect users wishes to override the default to be another port if needed.

If no `cloud_id` is set the port will default to `9200` as before.

I've also normalized the insecure connection warning messages between the Urllib3HttpConnection and RequestsHttpConnection to be the full scheme+host+port rather than just the host or including the URL path prefix as only (scheme, host, port) are relevant to TLS anyways.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -230,6 +230,27 @@ description of the options.
 
 .. _certifi: http://certifiio.readthedocs.io/en/latest/
 
+Connecting via Cloud ID
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Cloud ID is an easy way to configure your client to work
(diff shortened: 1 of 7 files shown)


sethmlarson:
Once everything is clarified I'll add a section on how to configure the client to use `cloud_id` along with all the configuration that occurs when using `cloud_id` as currently there's no section for that in our docs.

PR #1135, opened 2020-03-06
Add support for X-Opaque-Id

sethmlarson:
Adds `opaque_id` as a parameter similar to `request_timeout` and `ignore` which is transformed to a header at the `Transport` level. Applies to all endpoints through `query_params` decorator.

Will be backported to 7.x and 6.x branches.


elasticsearch/transport.py
```diff
@@ -59,6 +59,7 @@ def __init__(
         retry_on_status=(502, 503, 504),
         retry_on_timeout=False,
         send_get_body_as="GET",
+        opaque_id=None,
```
HonzaKral: I wonder if it wouldn't be easier to directly support `headers` here instead of just one of the use cases. Another use case that comes to mind is the `run_as` functionality which could also benefit from this - https://www.elastic.co/guide/en/elasticsearch/reference/current/run-as-privilege.html

What do you think?

That way we could expose the `headers` to the end user just like `params` and allow for some of the global args to be passed through there instead of params in the `query_params`

sethmlarson: Even with the addition of `headers` to the transport constructor I think we'd still want these individually configurable arguments for Opaque ID like we do for Cloud ID, API Key, basic auth, etc.  I don't think having `headers` also in the Transport constructor is a bad idea on its own, however having the "kwargs" setup that we currently have it's unfortunate to have headers in both the Transport and Connection constructors. :/

Also is the `headers` parameter for `Transport.perform_request()` used anywhere? I couldn't find any usage within the library itself.

HonzaKral: > Even with the addition of headers to the transport constructor I think we'd still want these individually configurable arguments for Opaque ID like we do for Cloud ID, API Key, basic auth, etc.

100%, exactly as we have named kwargs for any query parameters but further down the line (immediately in the API) we simplify it by folding it into `params`, what I'd suggest for headers is the same treatment

> Also is the headers parameter for Transport.perform_request() used anywhere? I couldn't find any usage within the library itself.

It is not used anywhere, it is future-proof idea for when we started using headers in elasticsearch (`opaque_id`, `run_as`, ...) but was never exposed all the way through to the APIs

> I don't think having headers also in the Transport constructor is a bad idea on its own, however having the "kwargs" setup that we currently have it's unfortunate to have headers in both the Transport and Connection constructors. :/

I treat it as only important for the connection, through `Transport` can make use of the values as well. I agree that the ambiguity is not ideal of course

sethmlarson: Ah I get you, I'll give that a spin and see how it looks. Probably simpler than this implementation.

sethmlarson:
Won't be backporting this feature to 6.x due to the difficulty of backporting without an API generator.

PR #1138, opened 2020-03-09
Add linting and docs to GitHub Actions CI

sethmlarson:
Adds linting and building of RTD documents on PRs. GitHub Actions is way better than Travis as we get a lot more concurrent jobs and each job takes ~30s instead of minimum 8-10 minutes. In the future will disable Travis completely for this repository once I move off Travis on all active branches.

Will mark the PR versions of each commit status are required on protected branches.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,55 @@
+---
+name: CI
+
+on: [push, pull_request]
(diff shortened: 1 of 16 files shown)


HonzaKral:
LGTM, thanks for the cleanup!

PR #1143, opened 2020-03-10
Restore default metric='_all' to cluster.state()

### LINKED ISSUE #1141: cluster.state(index='name')['routing_table'] not found in client 7.5.1
Using the most recent release of the Elasticsearch Python client (7.5.1 at this time), the cluster state endpoint is not defaulting to show `_all` keys.

### 7.5.1

```
client.cluster.state(index=idx).keys()
dict_keys(['cluster_name', 'cluster_uuid'])
```

### 7.1.1:

```
client.cluster.state(index=idx).keys()
dict_keys(['cluster_name', 'cluster_uuid', 'version', 'state_uuid', 'master_node', 'blocks', 'nodes', 'metadata', 'routing_table', 'routing_nodes'])
```

This omission, if intended, would seem to be something to change in a major release version, rather than a minor release.

This was detected and reported [here](https://github.com/elastic/support-dev-help/issues/10455#issue-558049925). This compels me to pin Curator at client version `elasticsearch==7.1.0` for now.

sethmlarson:
It looks like this little piece of logic was lost in the API auto-generation commit. I went through that commit and I didn't see any other special logic that was lost so we should be good now?
Should probably release a 7.5.2 with the past few commits and fixes.

Closes #1141

### DIFF
--- a/elasticsearch/client/cluster.py
+++ b/elasticsearch/client/cluster.py
@@ -102,6 +102,9 @@ class ClusterClient(NamespacedClient):
         :arg wait_for_timeout: The maximum time to wait for
             wait_for_metadata_version before timing out
         """
+        if index and not metric:
+            metric = "_all"
+
         return self.transport.perform_request(
             "GET", _make_path("_cluster", "state", metric, index), params=params
         )
--- a/elasticsearch/client/cluster.py
+++ b/elasticsearch/client/cluster.py
@@ -102,7 +102,7 @@ class ClusterClient(NamespacedClient):
         :arg wait_for_timeout: The maximum time to wait for
             wait_for_metadata_version before timing out
         """
-        if index and not metric:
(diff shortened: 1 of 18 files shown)


elasticsearch/client/cluster.py
```diff
@@ -102,6 +102,9 @@ def state(self, metric=None, index=None, params=None):
         :arg wait_for_timeout: The maximum time to wait for
             wait_for_metadata_version before timing out
         """
+        if index and not metric:
```
HonzaKral: I think just for consistency we might want to check against `SKIP_IN_PATH` instead of just the `bool` value. It amounts to the same thing in this case pretty much but would probably be better

sethmlarson: Yeah agreed.

PR #1164, opened 2020-03-19
Upgrade API to ES 8.0.0

sethmlarson:
This PR also migrates off of Travis on to Jenkins like the 7.x branch.


sethmlarson:
Yeah that'd be ideal, it's tough cuz the previous CI doesn't actually test x-pack. Would have to do git juggling locally and rely on everything being right in both PRs to get the desired final outcome. :(

Will be updating to the latest in the ci folder for sure, especially the newly generated certs.

PR #1179, opened 2020-03-24
Surface deprecation warnings from Elasticsearch

sethmlarson:


sethmlarson:
Looks like we're failing a handful of index templates tests on CI, if that's the only failure then I'll merge this and update the API for v8.0

PR #1180, opened 2020-03-25
Support serializing numpy and pandas types

### LINKED ISSUE #1178: Support serializing numpy and pandas types when available
Created this issue to resolve this: https://github.com/elastic/eland/issues/142

I'm thinking we can try importing `numpy` and `pandas` and if we detect either of those modules drop in to a separate section in `JSONSerializer.default()` that handles all the types automatically. I figure it's easier for us to do this once in the client than force everyone using numpy+elasticsearch to maintain this snippet themselves.

sethmlarson:
This PR attempts to import `numpy` and `pandas` and if either library is found adds to the list of types that the default `JSONSerializer` supports. Numpy adds the integers, floats, boolean, `ndarray`, and datetime. Pandas adds support for `Series`, `Timestamp`, and `NA` -> `None`. Am I missing any important types that can be safely serialized to JSON?

Notably I left out `DataFrame` and `numpy.nan`. NaN is already handled by JSON and doesn't have semantics for Elasticsearch (at least I don't think it does?) and `DateFrame` seemed a bit too heavy to support natively? Better for users to call `DataFrame.to_json()` themselves?

Also wanted to confirm my thinking that it is appropriate to support `Series` and `ndarray`? Or is that also too presumptive of what a user wants from the library?

Closes #1178 
Closes elastic/eland#142

### DIFF
--- a/elasticsearch/serializer.py
+++ b/elasticsearch/serializer.py
@@ -2,12 +2,48 @@ try:
     import simplejson as json
 except ImportError:
     import json
+
 import uuid
 from datetime import date, datetime
 from decimal import Decimal
 
 from .exceptions import SerializationError, ImproperlyConfigured
 from .compat import string_types
+
+INTEGER_TYPES = ()
+FLOAT_TYPES = (Decimal,)
+TIME_TYPES = (date, datetime)
+
+try:
+    import numpy as np
+
+    INTEGER_TYPES += (
+        np.int_,
+        np.intc,
+        np.int8,
+        np.int16,
+        np.int32,
+        np.int64,
+        np.uint8,
+        np.uint16,
+        np.uint32,
+        np.uint64,
+    )
+    FLOAT_TYPES += (
+        np.float_,
+        np.float16,
+        np.float32,
+        np.float64,
+    )
+except ImportError:
+    np = None
+
+try:
+    import pandas as pd
+
+    TIME_TYPES += (pd.Timestamp,)
+except ImportError:
+    pd = None
 
 
 class TextSerializer(object):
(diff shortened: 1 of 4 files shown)


elasticsearch/serializer.py
```diff
@@ -27,12 +63,29 @@ class JSONSerializer(object):
...
+        # Special cases for numpy and pandas types
+        elif np:
+            if isinstance(data, np.bool_):
+                return bool(data)
+            elif isinstance(data, np.datetime64):
+                return data.item().isoformat()
+            elif isinstance(data, np.ndarray):
+                return data.tolist()
+        if pd:
+            if isinstance(data, pd.Series):
```
Winterflower: I'm definitely not the expert here (so I might asking the wrong question here), but I've been recently working a bit with `pandas.dtypes` and I'm wondering, how will this serialiser handle the types of the elements inside a `pd.Series` or a `pd.DataFrame`. Usually these elements are various numerical types, so probably handled by the Numpy converters, but sometimes a user can set these dtypes to pandas specific stuff like `category`. For ex. I've recently done something like this in my Jupyter notebook

```
# finally we will correct the mappings on the remaining columns 

mappings = {'carat': 'float64',
           'cut': 'category',
           'color': 'category',
           'depth':'float64',
           'table':'float64',
           'price': 'float64',
           'x':'float64',
           'y':'float64',
           'z':'float64'}

df_cleaned = df_cleaned.astype(mappings)
```

How would the serialisation handle `category` or some of the other pandas specific dtypes in a Series (here is a list of some more exotic dtypes https://pandas.pydata.org/pandas-docs/stable/getting_started/basics.html#dtypes)?

sethmlarson: Category is an interesting one, we'd lose the "categorical" aspect of the value if we serialize to a string but maybe that's fine? Maintaining the categorical aspect would require config on the mapping but as long as that's done it'd be a solution, so maybe we do `category` -> `str`?

PR #1199, opened 2020-04-10
Update docstring

shotamakino:
escaped underscore -- from_ param mislabeled as from in rendered markdown


sethmlarson:
@shotamakino So this won't persist the change if I were to run the API generator again. Almost all code within the `elasticsearch/client/` directory is generated automatically.

You should try looking at this PR for what to do to fix issues like this: https://github.com/elastic/elasticsearch-py/pull/1177

sethmlarson:
Closing due to inactivity.

PR #1201, opened 2020-04-17
[7.x] Concatenate groups of strings in _make_path

sethmlarson:
`_make_path` doesn't have to work as hard now with chains of strings.


sethmlarson:
See explanation in https://github.com/elastic/elasticsearch-py/pull/1202 for close.

PR #1202, opened 2020-04-17
Concatenate groups of strings in _make_path

sethmlarson:
Also adds SearchableSnapshots client


sethmlarson:
This would actually break an assumption that slashes are percent-encoded within the individual path segments. Going to close, should probably add a test case for this behavior.

PR #1203, opened 2020-04-17
Add support for async I/O

sethmlarson:
- [x] Revises the generation of API from the JSON spec:
  - Generates the `elasticsearch/_async/client/...` code using normal workflow
  - Uses [`unasync`](https://github.com/python-trio/unasync) to translate async code to sync code in `elasticsearch/client/...`
- [x] On Python 3.6+ import the `AsyncElasticsearch`, `AsyncTransport`, and `AIOHttpConnection` into the `elasticsearch` namespace.
- [x] Add async helpers on Python 3.6+
  - [x] Use unasync to generate these?
- [x] Async unit tests
  - [x] Move to pytest
  - [x] Might be able to use unasync here too (nope)
  - [ ] On CI use a random gen seed, maybe on date, that "stripes" the test suite on Jenkins between all the connection classes
- [x] Trim down the AsyncTransport class, doesn't need to override much
- [x] AsyncTransport sniffing via async
  - May need to be different than the sync variant because most users will create `AsyncElasticsearch` in the global scope, outside an event loop.
  - Instead we can have sniff_on_start=True mean that sniffing happens in the background on the first `perform_request()` or call to  `__aenter__()`
  - Can attempt to grab a currently running loop in constructor but if we don't find one fallback on the above method?
- [x] Create documentation for `AsyncElasticsearch` and how to use
- [ ] Point to the new implementation on the `elasticsearch-async` repository
- [x] Add `aiohttp` as an extra `elasticsearch[async]`?


elasticsearch/_async/http_aiohttp.py
```diff
@@ -0,0 +1,189 @@
...
+        if params:
+            url_path = "%s?%s" % (url, urlencode(params or {}))
+        url = self.host + url_path
+
+        timeout = aiohttp.ClientTimeout(total=timeout)
+        req_headers = self.headers.copy()
+        if headers:
+            req_headers.update(headers)
+
+        loop = get_running_loop()
```
danielmitterdorfer, line 144: I think this can cause a performance penalty and I wonder whether it is worth caching the eventloop in a field (based on your PR description I understand you want to retrieve it lazily). The problem is that [the CPython implementation](https://github.com/python/cpython/blob/df8913f7c48d267efd662e8ffd9496595115eee8/Lib/asyncio/events.py#L693-L702) calls `getpid` on every call which in turn will [issue the `getpid` system call](http://man7.org/linux/man-pages/man2/getpid.2.html) for glibc > 2.25 (also [Python does not cache the pid](https://github.com/python/cpython/blob/2b5603140c09766a7d4e8243a70d7144f684f6f9/Modules/posixmodule.c#L6926-L6939)).

elasticsearch/_async/transport.py
```diff
@@ -0,0 +1,153 @@
...
+                # or as source parameter
+                elif self.send_get_body_as == "source":
+                    if params is None:
+                        params = {}
+                    params["source"] = body
+                    body = None
+
+        if body is not None:
+            try:
+                body = body.encode("utf-8")
```
vEpiphyte, line 98: This should have the errors=surrogatepass flag since the real world is rather untidy. Without it, this introduces known issues.

https://github.com/elastic/elasticsearch-py-async/issues/62
https://github.com/elastic/elasticsearch-py-async/pull/69/files

setup.py
```diff
@@ -10,7 +10,12 @@
 with open(join(dirname(__file__), "README")) as f:
     long_description = f.read().strip()
 
-install_requires = ["urllib3>=1.21.1", "certifi"]
+install_requires = [
+    "urllib3>=1.21.1",
+    "certifi",
+    # Async is supported on Python 3.6+
+    "aiohttp; python_version>='3.6'",
```
vEpiphyte, line 17: I'd suggest adding a version range for aiohttp like ``aiohttp<4.0,>=3.5.4`` to avoid getting a potentially unknown package and having an API break occur.

sethmlarson: Good idea, I'm planning on adding an extra for async via `elasticsearch[async]` so will fix then!

elasticsearch/_async/transport.py
sethmlarson: Agreed, will take from the sync version so it's identical.

elasticsearch/_async/http_aiohttp.py
sethmlarson: I'll cache this along with the "sniff on first async action", good point!

elasticsearch/_async/http_aiohttp.py
```diff
@@ -0,0 +1,189 @@
...
+        self, method, url, params=None, body=None, timeout=None, ignore=(), headers=None
+    ):
+        url_path = url
+        if params:
+            url_path = "%s?%s" % (url, urlencode(params or {}))
+        url = self.host + url_path
+
+        timeout = aiohttp.ClientTimeout(total=timeout)
+        req_headers = self.headers.copy()
+        if headers:
```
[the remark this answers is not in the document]
sethmlarson: Probably! I think we can trust aiohttp to not modify things, they're probably copying themselves in some way as well :)

elasticsearch/_async/http_aiohttp.py
```diff
@@ -0,0 +1,189 @@
...
+
+    async def perform_request(
+        self, method, url, params=None, body=None, timeout=None, ignore=(), headers=None
+    ):
+        url_path = url
+        if params:
+            url_path = "%s?%s" % (url, urlencode(params or {}))
+        url = self.host + url_path
+
+        timeout = aiohttp.ClientTimeout(total=timeout)
```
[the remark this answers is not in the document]
sethmlarson: Yes will need to grab all defaults from the sync Transport and make them identical

elasticsearch/_async/http_aiohttp.py
```diff
@@ -0,0 +1,189 @@
...
+    async def close(self):
+        await self.session.close()
+
+    async def perform_request(
+        self, method, url, params=None, body=None, timeout=None, ignore=(), headers=None
+    ):
+        url_path = url
+        if params:
+            url_path = "%s?%s" % (url, urlencode(params or {}))
+        url = self.host + url_path
```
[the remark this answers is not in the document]
sethmlarson: Definitely will make this improvement, need to look at Rally's async conn implementation for other things to steal ;)

elasticsearch/_async/http_aiohttp.py
```diff
@@ -0,0 +1,189 @@
...
+            else:
+                raise ImproperlyConfigured("ca_certs parameter is not a path")
+
+        self.headers.setdefault("connection", "keep-alive")
+        self.session = aiohttp.ClientSession(
+            auth=http_auth,
+            headers=self.headers,
+            auto_decompress=True,
+            connector=aiohttp.TCPConnector(
+                limit=maxsize,
```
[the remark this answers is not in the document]
sethmlarson: `maxsize` is actually from the `Urllib3HttpConnection`, I can use a better name here.

elasticsearch/_async/http_aiohttp.py
```diff
@@ -0,0 +1,189 @@
...
+                verify_ssl=verify_certs,
+                use_dns_cache=True,
+                ssl_context=ssl_context,
+                keepalive_timeout=10,
```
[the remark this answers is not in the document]
sethmlarson: Was just going off the things I've seen before mentioning 7-10s as good default for keepalive, I may as well keep it the default for now.

sethmlarson:
Closed in favor of #1230, thanks for all the early review comments!

Issue #1196, opened 2020-04-08
Error in documentation for es.search function


bfishh:
In this doc:
https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch
for "serach" function offset parameter is named as "from" but in code it is "from_".
The name "from" is invalid and causes invalid syntax error.

sethmlarson:
@bfishh Great catch, this looks to be an issue in our API generator. Would you be up to make this change?

bfishh:
>Would you be up to make this change?
Not this time, sorry.

bartier:
@sethmlarson Hi! Can I work on this? If yes, could you provide more information about how you would handle this issue?

`from` is a reserved keyword in Python, so I guess the documentation should be updated, right?

https://github.com/elastic/elasticsearch-py/blob/7dec9e3dbaabb2b7cc5cdb1a49370c89de72fa12/elasticsearch/client/__init__.py#L702-L704

I would propose something like this in the  docs:
![image](https://user-images.githubusercontent.com/18057391/79622008-e7a15000-80eb-11ea-9301-019141acaec1.png)


Reference link to actual content: https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.Elasticsearch.search

The actual docs is wrong because `_` needs a back slash `\` to be rendered:

Actual code:
https://github.com/elastic/elasticsearch-py/blob/7dec9e3dbaabb2b7cc5cdb1a49370c89de72fa12/elasticsearch/client/__init__.py#L656

Code that would work:
```
        :arg from\_: Starting offset (default: 0)
                     ``from`` is a reserved keyword in Python. In order to avoid a syntax error this function uses ``from_``
```

Is this a valid solution? If yes, can I send a PR? :)

sethmlarson:
@bartier That would be great! You'll have to make the change within the `utils/` directory as all of the code including docstrings under `elasticsearch/clients/`  is generated from `utils/generate_api.py`

You can take a look at the commit history of that file for examples :)

Let me know if you run into problems or have questions, I'm here to help! :D

bartier:
@sethmlarson Thanks for answering me :D

I sent my a PR (first one here!) that attempts to solve this issue. Could you check it if it is a valid solution?

PR #1206, opened 2020-04-20
Fix docs

bartier:
This PR attempts to close issue #1196

As discussed in the issue, `from` is a reserved Python keyword, and parameters are passed with `from_` to avoid a syntax error. See a example below:

https://github.com/elastic/elasticsearch-py/blob/7dec9e3dbaabb2b7cc5cdb1a49370c89de72fa12/elasticsearch/client/__init__.py#L702-L704

Docs were not rendering the `_` underscore character and it would causes a syntax error if you use `from` instead of `from_`.

To correcly output underscore character it just needs a `\` before it.

Resume of changes:
- Change inside the utils/ directory as all of the code including docstrings under elasticsearch/clients/ is generated from utils/generate_api.py

Hope it helps :)

### DIFF
--- a/elasticsearch/client/__init__.py
+++ b/elasticsearch/client/__init__.py
@@ -653,7 +653,7 @@ class Elasticsearch(object):
         :arg expand_wildcards: Whether to expand wildcard expression to
             concrete indices that are open, closed or both.  Valid choices: open,
             closed, hidden, none, all  Default: open
-        :arg from_: Starting offset (default: 0)
+        :arg from\\_: Starting offset (default: 0)
         :arg ignore_unavailable: Whether specified concrete indices
             should be ignored when unavailable (missing or closed)
         :arg lenient: Specify whether format-based query failures (such
@@ -1408,7 +1408,7 @@ class Elasticsearch(object):
             closed, hidden, none, all  Default: open
         :arg explain: Specify whether to return detailed information
             about score computation as part of a hit
-        :arg from_: Starting offset (default: 0)
+        :arg from\\_: Starting offset (default: 0)
         :arg ignore_throttled: Whether specified concrete, expanded or
             aliased indices should be ignored when throttled
         :arg ignore_unavailable: Whether specified concrete indices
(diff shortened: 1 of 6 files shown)


bartier:
@sethmlarson Thanks for your reply. I  ran successfuly `utils/generate_api.py` after cloning elasticsearch.

In my understanding, changing only `SUBSTITUTIONS = {"type": "doc_type", "from": "from\_"}` affects places that should not be changed. (real code for example, instead of documentation, right?)

I could see some changes with my Git UI (GitKraken) after running the script (image below)

I don't understand specifically why the first diff with `included in the analysis...` appeared. Is that a change that should be there? 

The two diffs about `from` I know it is my change taking effect.
image

In essence, I need to make it work only with `:arg from_:` string to replace `from` word, right?

Thanks for your time and patience. [noob alert here haha]

sethmlarson:
@bartier Don't worry about the "included with analysis" you shouldn't have to commit any of the changes not related to `from\_`. Your assessment of which things should change and which should not are correct, we only want docstrings to change. :)

And no worries, I'm here to help you through the process, I really appreciate you taking on this issue :D

bartier:
@sethmlarson Hi! Is my [last commit](https://github.com/elastic/elasticsearch-py/pull/1206/commits/71a93192767329f1c5d3a1a48b415c9d37055422) a valid solution?

I've changed `utils/templates/base` like you said and committed the results of the script related to 'from'.

A lint error in CI appeared and I'm not sure how should I fix it :/ 

BTW, this PR commit history is kind of messy. Will my PR be closed automatically by GitHub if I squash all commits to generate only one (and facilitate review code process)? I ask that because my fail PR #1205 looks that this happened.

sethmlarson:
The lint error can be seen here: https://github.com/elastic/elasticsearch-py/pull/1206/checks?check_run_id=603994279 It is caused by the `\` within the docstrings, maybe you can change it to `\\`? Either that or use `r"""` for docstring beginnings.

No worries on the commit history, I'll squash all commits to one when merging.

bartier:
@sethmlarson I escaped `\` to fix the lint error.

Can you check it again? :)

bartier:
@sethmlarson Now I guess everything is good. I fixed my mistakes with commit history I guess (and PR was not closed automatically haha).

bartier:
@sethmlarson Is there anything I need to do to keep this going on or something I maybe have done wrong? :)

Thanks

PR #1208, opened 2020-04-20
fix params is None, avoid raising AttributeError

zhucebuliaopx:
fix params is None, avoid raising AttributeError:'NoneType' object has no attribute 'copy'


sethmlarson:
Hey thank you for noticing this issue, I fixed it already in https://github.com/elastic/elasticsearch-py/pull/1209. Closing this for now.

PR #612, opened 2017-06-30
Fixes non UTF-8 surrogateescapes

tylerjharden:
Surrogate escapes in Unicode (non UTF-8 encoding) will be properly escaped with backslashes when encountered, versus breaking the transport layer.

This addresses my issue here: https://github.com/elastic/elasticsearch-py/issues/611

PR #1212, opened 2020-04-20
Missing surrogatepass in urllib3 handler

vEpiphyte:
A coworker of mine encountered an issue when retrieving a document, when using the scrollapis, causing a decoding issue in the urllib3 connection class.

This is similar to #612 and #629 .


vEpiphyte:
Coworker has confirmed for me that this has resolved the issue he was seeing.

sethmlarson:
jenkins test this please

vEpiphyte:
@sethmlarson I'd be happy to add it to `RequestsHttpConnection`. I was initially a bit hesitant because of the use of `response.text` which does automatic decoding vs. decoding the raw response content; but it should be trivial to add.

PR #1232, opened 2020-05-14
Add AIOHttpConnection and [async] extra

sethmlarson:
Adds the `AIOHttpConnection` class which uses `aiohttp` as an HTTP client.


sethmlarson:
(CI failure is a failed `docker pull`, all other tests passed)

elasticsearch/_async/http_aiohttp.py
```diff
@@ -0,0 +1,298 @@
...
+                data=body,
+                headers=req_headers,
+                timeout=timeout,
+                fingerprint=self.ssl_assert_fingerprint,
+            ) as response:
+                if is_head:  # We actually called 'GET' so throw away the data.
+                    await response.release()
+                    raw_data = ""
+                else:
+                    raw_data = (await response.read()).decode("utf-8", "surrogatepass")
```
vEpiphyte, line 240: :heart:

danielmitterdorfer, line 240: An alternative to this would be to call:

```python
raw_data = await response.text()
```

and define a custom aiohttp client response:

```python
class ESClientResponse(aiohttp.ClientResponse):
    async def text(self, encoding=None, errors="strict"):
        if self._body is None:
            await self.read()

        return self._body.decode("utf-8", "surrogatepass")
```

This needs to be passed with `response_class=ESClientResponse` when creating the `aiohttp.ClientSession` below. The reason I'm suggesting this is that we want to work with raw bytes in Rally and if the client would delegate to the response class, it would be easier to swap.

elasticsearch/_async/http_aiohttp.py
```diff
@@ -0,0 +1,298 @@
...
+        return response.status, response.headers, raw_data
+
+    async def close(self):
+        """
+        Explicitly closes connection
+        """
+        if self.session:
+            await self.session.close()
+
+    async def _create_aiohttp_session(self):
```
danielmitterdorfer, line 283: That's great for very target customization of the session!

PR #1235, opened 2020-05-14
Add AsyncTransport

sethmlarson:
This PR adds `AsyncTransport` to the `AIOHttpConnection` branch as it depends on the async http connection being available to test.


sethmlarson:
CI appears to be failing due to the random order of tasks returning in the sniff task, will see if I can make the task order consistent.

elasticsearch/transport.py
```diff
@@ -405,3 +391,38 @@ def close(self):
...
+                # or as source parameter
+                elif self.send_get_body_as == "source":
+                    if params is None:
+                        params = {}
+                    params["source"] = body
+                    body = None
+
+        if body is not None:
+            try:
+                body = body.encode("utf-8", "surrogatepass")
```
vEpiphyte, line 415: :heart:

elasticsearch/_async/transport.py
```diff
@@ -0,0 +1,327 @@
...
+                if data:
+                    data = self.deserializer.loads(data, headers.get("content-type"))
+                return data
+
+    async def close(self):
+        """
+        Explicitly closes connections
+        """
+        if self.sniffing_task:
+            self.sniffing_task.cancel()
```
danielmitterdorfer, line 324: According to the [docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel), this is only requesting a cancellation so I think this should be followed by

```python
await self.sniffing_task
```

PR #923, opened 2019-04-02
Update starting example in docs to v7

elyase:
Should fix #922

PR #1268, opened 2020-05-31
Fix example usage

elyase:


docs/index.rst
```diff
@@ -67,26 +67,26 @@ Example Usage
 
 ::
 
+    from locasticsearch import Locasticsearch
     from datetime import datetime
-    from elasticsearch import Elasticsearch
-    es = Elasticsearch()
+
+    es = Locasticsearch()
```
sethmlarson, line 73: What is locasticsearch? Typo?

docs/index.rst
```diff
@@ -67,26 +67,26 @@ Example Usage
...
+        "text": "Elasticsearch: cool. bonsai cool.",
+        "timestamp": datetime(2010, 10, 10, 10, 10, 10),
     }
-    res = es.index(index="test-index", id=1, body=doc)
-    print(res['result'])
+    res = es.index(index="test-index", doc_type="tweet", id=1, body=doc)
 
-    res = es.get(index="test-index", id=1)
-    print(res['_source'])
+    res = es.get(index="test-index", doc_type="tweet", id=1)
```
sethmlarson, line 82: With 7.x we shouldn't use `doc_types` in examples, why are we using this here?

docs/index.rst
elyase: it is a new POC library I am experimenting with. But definitely out of place here :-)

docs/index.rst
elyase: I was also confused. Stripped them away now

sethmlarson:
@elyase I'm not seeing what is improved with this diff, can you explain the rationale?

elyase:
Well this is somewhat embarrassing but I just realized that I was trying to fix something that is already fixed. To make things worse I just noticed I myself submitted the [same pull request](https://github.com/elastic/elasticsearch-py/pull/923) last year 🤦‍♂️.

I think I must have opened an old version of the documentation, is the only explanation I can think of. So sorry for the waste of time and lets pray it doesn't happen again next year :-)

sethmlarson:
@elyase No problem! You just wanted to make sure that issue was **definitely fixed** :wink: Thanks for your contributions.

PR #1292, opened 2020-06-23
Update .rst doc files, solve some issues regarding document generation

### LINKED ISSUE #1287: Migrate all rST code blocks from :: to code-block: python
Our documentation has a mix of `::` and ` .. code-block: python`, we should migrate to all `code-block: python` so it's less confusing and more apparent when a code block is defined.

mesejo:
- Update documentation code blocks
- Remove duplicated reference in changelog
- Remove unused class DeprecationClient from docs
- Remove static file configuration in conf.py

Closes #1287


docs/api.rst
```diff
@@ -30,7 +30,8 @@ elasticsearch returns a 2XX response. Otherwise an instance of
 raised. You can see other exception and error states in :ref:`exceptions`. If
 you do not wish an exception to be raised you can always pass in an ``ignore``
 parameter with either a single status code that should be ignored or a list of
-them::
+them
+.. code-block:: python
```
sethmlarson, line 34: Typically there's a blank line between text and a directive.

docs/xpack.rst
```diff
@@ -202,15 +202,3 @@ saves it to a destination index.
 .. autoclass:: TransformClient
    :members:
 
-
```
sethmlarson, line 205: Let's handle this outside this PR so I can more easily backport to 7.x

.gitignore
```diff
@@ -15,3 +15,5 @@ coverage.xml
 nosetests.xml
 junit-*.xml
 build
+.idea/
```
sethmlarson, line 18: Can we actually change this to GitHubs Python gitignore template? Covers a lot more bases! Then add `test_elasticsearch/local.py` and `.idea` to it.

docs/api.rst
mesejo: Done

docs/xpack.rst
mesejo: Done

PR #1313, opened 2020-07-03
fix #1299 by catching only HttpError

### LINKED ISSUE #1299: perform_request hijacked user defined exception
**`elasticsearch-py` version (`master`)**: 

https://github.com/elastic/elasticsearch-py/blob/f40390187a79ccd773a09aa1f671562d42cd3938/elasticsearch/connection/http_urllib3.py#L237

this try exception hijacked all exceptions, and raised with a `SSLError` or `ConnectionTimeout` or `ConnectionError`.All user defined exceptions with other type will be hijacked.So that, users can not catch the exception in up level code.

**Steps to reproduce**:

```python
class MyTimeout(Exception):
    pass

class timeout(object):
    """
    To be used in a ``with`` block and timeout its content.
    """

    def __init__(self, seconds=1, error_message='Timeout'):
        self.seconds = seconds
        self.error_message = error_message

    def handle_timeout(self, signum, frame):
        self.log.error("Process timed out")
        raise MyTimeout(self.error_message)

    def __enter__(self):
        try:
            signal.signal(signal.SIGALRM, self.handle_timeout)
            signal.alarm(self.seconds)
        except ValueError as e:
            self.log.warning("timeout can't be used in the current context")
            self.log.exception(e)

    def __exit__(self, type, value, traceback):
        try:
            signal.alarm(0)
        except ValueError as e:
            self.log.warning("timeout can't be used in the current context")
            self.log.exception(e)

def long_time_request():
    # DO A LONG TIME QUERY，LONG THAN 5 SECONDS
    pass

def do_query():
    try:
        with timeout(5): # 5 SECONDS TIMEOUT
            long_time_request()
    except Exception as e:
        assert isinstance(e, MyTimeout) # can not get MyTimeout
    
if __name__ == "__main__":
    do_query()
```

lsyarn:
Catch only HTTPError when deal with request_posts. So that, other exceptions will not be hijacked.


lsyarn:
Signed Contributor Agreement

sethmlarson:
Thanks for contributing this. I'm closing this for now, see issue for explanation.

Issue #1259, opened 2020-05-27
Bulk Helper: Actions params whitelist is missing `if_seq_no` and `if_primary_term`


USSRLivesOn:
**Elasticsearch version** (`bin/elasticsearch --version`):
7.4.2

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**:
7.5.1, but as far as I can tell, bug is still in `master`.

**Description of the problem including expected versus actual behavior**:
The `bulk` helper should respect the [`if_seq_no` and `if_primary_term` parameters](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/docs-bulk.html#bulk-optimistic-concurrency-control) in the action, but it does not. As far as I can tell, they're just ignored entirely because they're not included in the [`expand_action` whitelist](https://github.com/elastic/elasticsearch-py/blob/cd85aa2ba31b2745ffa71d2edf6902400639bafc/elasticsearch/helpers/actions.py#L34-L49). They should be included in the whitelist.

**Steps to reproduce**:
Use the `bulk` helper with the `if_seq_no` and `if_primary_term` params. Notice that they don't actually get sent to the cluster.

----
Separately, it'd be really nice if the `bulk` helper didn't just silently ignore unrecognized attributes. If you're whitelisting the valid params, then I'd expect at least a warning logged that an unrecognized param was encountered. An error would be even better.

sethmlarson:
@USSRLivesOn This sounds like a good improvement, would you be willing to create the patch adding `if_seq_no` and `if_primary_term` to the list?

I also like the idea of having a warning for unrecognized parameters.

sethmlarson:
This issue has been resolved in #1387

PR #1316, opened 2020-07-07
Fix for issue #1259

saket1205:
Signed the contributor agreement


saket1205:
check is unable to recognize my signed contributor agreement

PR #1317, opened 2020-07-07
Fix for issue #1259

saket1205:
signed contributor agreement and added params for bulk apis


sethmlarson:
Thanks for this, could you add a test case or two to verify this functionality works as expected?

sethmlarson:
Closed in favor of #1387

Issue #1261, opened 2020-05-28
log unsuccessful shards in failed scrolls


tommyzli:
**Elasticsearch version** (`bin/elasticsearch --version`): 7.6.1

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**: 7.5.1

**Description of the problem including expected versus actual behavior**:

The scan() helper function only logs the number of successful vs failed shards. It would be helpful to also log the shards that failed, so I can quickly jump onto the node and grab the appropriate server logs. That data is a part of the response, but gets thrown away by the client.

**Steps to reproduce**:
A call to `scan(client, query, raise_on_error=True)` fails and throws
`ScanError("Scroll request has only succeeded on 9 (+0 skiped) shards out of 10.")`

Proposed error:
`ScanError("Scroll request has only succeeded on 9 (+0 skipped) shards out of 10. First failure: node 'foo', shard 'bar', reason 'reason'")`

bartier:
@tommyzli Unfortunately I did not find the information of shards/nodes that were not successful to answer using [scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-request-body.html#request-body-search-scroll). I may be forgetting something, but only the successful/total shards counts is presented in the raw response.

The only exception where I could see  scroll API send information about nodes that do not return successfully is when the _scroll_id request has answered initially with all shards successfully (for example 3/3) and when consuming the scroll API  shards become unavailable because a node is unreachable (2/3 primary shards available in example bewlow). Is that what you are referring to?


1) Request _scroll_id with all shards available (3/3 in example)
```
POST /twitter/_search?scroll=1m&pretty HTTP/1.1
{
    "size": 3,
    "query": {
        "match_all" : {}
    }
}
# Response 3/3 shards
{
  "_scroll_id" : "FGluY2x1ZGVfY29udGV4dF91dWlkDnF1ZXJ5VGhlbkZldGNoAxRjQTM2SVhNQk1hMC15cHkyd2o4egAAAAAAAAAYFkZzZ2tJN2JrVEMtc1RUbGcxcWl2TmcUelRyNklYTUJQQUNTcE1WendnVXcAAAAAAAAA_BYzYXp1WXJ1LVRwV1JSd004dlV2YmNRFFlldjZJWE1CY3ZEeDdNb253aDR4AAAAAAAAAAgWdnNSSTJPSVpRd0dJbUxvN3RZX3I2QQ==",
  "took" : 10,
[...]

PR #1318, opened 2020-07-07
Log scroll response failures

bartier:
As raised by issue #1261, when the endpoint POST /_search/scroll answers with failures indications they are not provided in the exception or log to help troubleshoot the root cause.

This PR intends to solve this adding failures indications like the example below:
image

A reason limit (5) is set to avoid the exception end up massive as pointed by @tommyzli

### DIFF
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -528,23 +528,30 @@ def scan(
                 resp_failures = resp["_shards"].get("failures")
                 failures = []
                 if resp_failures is not None:
-                    for resp_failure in resp_failures:
-                        failure = "Failure type[%s] received with reason: %s" % (
-                            resp_failure["reason"]["type"],
-                            resp_failure["reason"]["reason"],
+                    for i in range(5):
+                        if i < len(resp_failures):
+                            failure = "Failure type[%s] received with reason: %s" % (
+                                resp_failures[i]["reason"]["type"],
+                                resp_failures[i]["reason"]["reason"],
+                            )
+                            failures.append(failure)
+                            logger.warning(failure)
+
+                    if len(resp_failures) > 5:
+                        omitted_failures = len(resp_failures) - 5
+                        failures.append(
+                            "...and %d more omitted failures" % omitted_failures
                         )
-                        failures.append(failure)
-                        logger.warning(failure)
                 if raise_on_error:
                     raise ScanError(
                         scroll_id,
-                        "Scroll request has only succeeded on %d (+%d skiped) shards out of %d.\n"
+                        "Scroll request has only succeeded on %d (+%d skipped) shards out of %d.\n"
                         "%s"
                         % (
                             resp["_shards"]["successful"],
                             resp["_shards"]["skipped"],
                             resp["_shards"]["total"],
-                            "".join(failures),
+                            "\n".join(failures),
                         ),
                     )
             resp = client.scroll(
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -525,12 +525,14 @@ def scan(
                     resp["_shards"]["total"],
                 )
 
-                resp_failures = resp['_shards'].get('failures')
+                resp_failures = resp["_shards"].get("failures")
                 failures = []
                 if resp_failures is not None:
                     for resp_failure in resp_failures:
-                        failure = "Failure type[%s] received with reason: %s" % \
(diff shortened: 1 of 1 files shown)


elasticsearch/helpers/actions.py
```diff
@@ -524,14 +524,27 @@ def scan(
...
+
+                resp_failures = resp["_shards"].get("failures")
+                failures = []
+                if resp_failures is not None:
+                    for resp_failure in resp_failures:
+                        failure = "Failure type[%s] received with reason: %s" % (
+                            resp_failure["reason"]["type"],
+                            resp_failure["reason"]["reason"],
+                        )
+                        failures.append(failure)
```
tommyzli, line 536: If you have a large number of shards, this line can end up massive. I think just logging the first failure is a reasonable tradeoff for visibility

elasticsearch/helpers/actions.py
```diff
@@ -524,14 +524,27 @@ def scan(
...
+                        logger.warning(failure)
                 if raise_on_error:
                     raise ScanError(
                         scroll_id,
-                        "Scroll request has only succeeded on %d (+%d skiped) shards out of %d."
+                        "Scroll request has only succeeded on %d (+%d skiped) shards out of %d.\n"
```
tommyzli, line 541: nit: `skiped` -> `skipped`

tommyzli (commented):
Thanks for picking up my issue!

elasticsearch/helpers/actions.py
bartier: @tommyzli What do you think about a reasonable limit (5 failures for example)? The exception would look like the example below:
line1
line2
line3
line4
line5
...and 23 more omitted failures

tommyzli: @bartier I like that!

bartier:
@sethmlarson Hi! If possible, could you do a code review at this PR? :)

I guess me and @tommyzli  agree on the proposal code changes already.

sethmlarson (changes requested):
Could you write some unit tests for this functionality specifically a <=5 errors case and >5 errors case.

bartier:
@sethmlarson Sure. Unfortunately I'm struggling to run the test module where I want to add my tests. I made the following steps to run and it shows a error like 'AttributeError: 'TestXXXXX' object has no attribute 'client'. I guess somehow I am not able to start ES but I'm not sure how I should make this works. If I could make this work I could continue to work doing the tests.

Any idea if I'm doing something wrong?

Notes:
I have installed docker/docker-compose and get_client from `test_elasticsearch/test_server/__init__.py` is raising error trying to import `test_elasticsearch.local` and also the fallback `test.get_test_client(**kwargs)` raises an error.
image

```
$ git clone https://github.com/bartier/elasticsearch-py.git
$ cd elasticsearch-py && git checkout log-scroll-failures
$ python3.8 -m venv venv
$ source venv/bin/activate
$ python setup.py test -s test_elasticsearch.test_server # error happens here

# Below output was truncated
======================================================================
ERROR: test_all_documents_are_read (test_elasticsearch.test_server.test_helpers.TestScan)
----------------------------------------------------------------------
  File "/home/vitor/github/elasticsearch-py/test_elasticsearch/test_server/test_helpers.py", line 362, in test_all_documents_are_read
    self.client.bulk(bulk, refresh=True)
AttributeError: 'TestScan' object has no attribute 'client'

======================================================================
ERROR: test_clear_scroll (test_elasticsearch.test_server.test_helpers.TestScan)
----------------------------------------------------------------------
  File "/home/vitor/github/elasticsearch-py/test_elasticsearch/test_server/test_helpers.py", line 481, in test_clear_scroll
    self.client.bulk(bulk, refresh=True)
AttributeError: 'TestScan' object has no attribute 'client'

======================================================================  
```

ajosh0504:
Hi! @sethmlarson @bartier Has there been any progress on this? I am running into The same error.

bartier:
> Hi! @sethmlarson @bartier Has there been any progress on this? I am running into The same error.

Unfortunately it has not been progress yet. I guess maintainers should have a look at this to give a hint at least how we could run the tests, so we can make unit tests and go ahead with PRs.

Sparkycz:
> > Hi! @sethmlarson @bartier Has there been any progress on this? I am running into The same error.
> 
> Unfortunately it has not been progress yet. I guess maintainers should have a look at this to give a hint at least how we could run the tests, so we can make unit tests and go ahead with PRs.

Try to run tests over `py.test` command. e.g. `$ py.test test_elasticsearch/test_server/test_helpers.py`

PR #1324, opened 2020-07-20
Update APIs and docs to Elasticsearch 7.9

sethmlarson:


sethmlarson:
Only failure is 3.7 on GHA for unrelated install reasons and transient Docker pull failures. Merging this.

PR #1279, opened 2020-06-16
Catch sniff errors during sniff on fail if retrying the request

l1nd3r0th:
If the request timed out, it seems likely that sniffing right after might also fail. And when it does, it used to interrupt the retry logic. This fix will make it proceed to retry the original request as configured.

Here's a sample of what that would look like in the wild (from es-py 7.0.5 on py2.7):
```
[application code calling .save() on a doc instance]

/srv/app/scs/env/local/lib/python2.7/site-packages/elasticsearch_dsl/document.pyc in save(self, using, index, validate, skip_empty, **kwargs)
    454             index=self._get_index(index),
    455             body=self.to_dict(skip_empty=skip_empty),
--> 456             **doc_meta
    457         )
    458         # update meta information from ES
         
/srv/app/scs/env/local/lib/python2.7/site-packages/elasticsearch/client/utils.pyc in _wrapped(*args, **kwargs)
     82                 if p in kwargs:
     83                     params[p] = kwargs.pop(p)
---> 84             return func(*args, params=params, **kwargs)
     85  
     86         return _wrapped
         
/srv/app/scs/env/local/lib/python2.7/site-packages/elasticsearch/client/__init__.pyc in index(self, index, body, doc_type, id, params)
    362                 raise ValueError("Empty value passed for a required argument.")
    363         return self.transport.perform_request(
--> 364             "POST", _make_path(index, doc_type, id), params=params, body=body
    365         )
    366  
         
/srv/app/scs/env/local/lib/python2.7/site-packages/elasticsearch/transport.pyc in perform_request(self, method, url, headers, params, body)
    365                 if retry:
    366                     # only mark as dead if we are retrying
--> 367                     self.mark_dead(connection)
    368                     # raise exception on last retry
    369                     if attempt == self.max_retries:
         
/srv/app/scs/env/local/lib/python2.7/site-packages/elasticsearch/transport.pyc in mark_dead(self, connection)
    284         self.connection_pool.mark_dead(connection)
    285         if self.sniff_on_connection_fail:
--> 286             self.sniff_hosts()
    287  
    288     def perform_request(self, method, url, headers=None, params=None, body=None):
         
/srv/app/scs/env/local/lib/python2.7/site-packages/elasticsearch/transport.pyc in sniff_hosts(self, initial)
    261             (``sniff_on_start``), ignore the ``sniff_timeout`` if ``True``
    262         """
--> 263         node_info = self._get_sniff_data(initial)
    264  
[...]

PR #1326, opened 2020-07-20
Don't raise sniffing error when retrying request

sethmlarson:
When an HTTP request fails and we decide to retry the request, `mark_dead()` can cause a sniff attempt to start if `sniff_on_connection_fail=True`. We don't want to interrupt a retry if a sniff request fails, instead we'd rather retry or raise the original exception that occurred.

Related: #1279


sethmlarson:
Only failure is 3.7 on GHA for unrelated install reasons. Merging this.

PR #1352, opened 2020-08-20
Have AIOHttpConnection honour the 'url_prefix' argument. Fixes #1351

### LINKED ISSUE #1351: Async connection ignores url_prefix parameter
**Elasticsearch version** (`bin/elasticsearch --version`): 7.8.1

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**:  7.8.1

I'm using AsyncElasticsearch. If I try to query a server with a url_prefix in the address, I get a 404 Not Found in the logs. The logs also mention the URL that the library tries to connect to, and the url prefix is missing from the URL. Expected behaviour is of course a reply to the query.

**Steps to reproduce**:

```
client = AsyncElasticsearch('https://es.example.com/prefix')
resp = await client.search(index='index', body={ 'query': { 'match': { 'field': 'value' } } } )

```
The issue seems to be in elasticsearch/_async/http_aiohttp.py, in perform_request(). It builds a URL using yarl without taking the url_prefix into account. This is still the case in the master branch

tjeerdbakker38:


tjeerdbakker38:
Although it seems a bit overkill for such a trivial contribution, I signed the contributor agreement anyway

sethmlarson:
@tjeerdbakker38 Could you create a unit test for this functionality?

tjeerdbakker38:
I could create a very simple check if the url_prefix attribute of the connection is set, but I'm not up to spinning up a custom server with a url prefix for this test case...

sethmlarson:
No problem, I'll make the fix and create test cases. Thanks for raising the issue!

sethmlarson:
Closed in favor of https://github.com/elastic/elasticsearch-py/pull/1357

PR #1368, opened 2020-09-01
Need to propagate headers from the scan API to the clear_scroll API

ashokc:
When headers (say an oauth2 token for authentication) are passed to the helper scan API, that header/token should be passed down to the clear_scroll API. Else, an authentication exception will ensue. The user can avoid this error by explicitly setting 'clear_scroll' to False - it is True by default of course.  The error message

AuthenticationException(401, 'security_exception',
{ 
  "error": {
    "root_cause": [
      {
        "type": "security_exception",
        "reason": "missing authentication credentials for REST request [/_search/scroll]",
        "header": {
          "WWW-Authenticate": [
            "Bearer realm=\"security\"",
            "ApiKey"
          ]
        }
      }
    ],
    "type": "security_exception",
    "reason": "missing authentication credentials for REST request [/_search/scroll]",
    "header": {
      "WWW-Authenticate": [
        "Bearer realm=\"security\"",
        "ApiKey"
      ]
    }
  },
  "status": 401
}

The fix here is to simply check if the incoming 'kwargs' to the scan API have headers, and pass it down. Simple

### DIFF
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -535,7 +535,7 @@ def scan(
                         ),
                     )
             resp = client.scroll(
-                body={"scroll_id": scroll_id, "scroll": scroll}, **scroll_kwargs
+                body={"scroll_id": scroll_id, "scroll": scroll}, headers=kwargs.get('headers', None), **scroll_kwargs
             )
             scroll_id = resp.get("_scroll_id")
 
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -535,13 +535,20 @@ def scan(
                         ),
                     )
             resp = client.scroll(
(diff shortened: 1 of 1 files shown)


ashokc:
If the headers are only in kwargs (but not in scroll_kwargs, which does get passed to the scroll API) - they need to be propagated to the scroll API as well

sethmlarson:
Thanks for this change, could you run `nox -s blacken` locally to fix the linting issues and add a test case to ensure this behavior is working as expected? I'd recommend checking out `test_elasticsearch/test_server/test_helpers.py` for some examples :)

ashokc:
> Thanks for this change, could you run `nox -s blacken` locally to fix the linting issues and add a test case to ensure this behavior is working as expected? I'd recommend checking out `test_elasticsearch/test_server/test_helpers.py` for some examples :)

Fixed the lint issue with nox. Will take a look at writing a test case in a bit.

PR #1392, opened 2020-10-03
Fix #1391: add option to use cache on ElasticSearch::count

### LINKED ISSUE #1391: .count, when using `request_cache` as a parameter results in an Exception being raised, yet results are present
**Elasticsearch version** (`bin/elasticsearch --version`): 7.9.0

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**: 7.9.1

When using `request_cache=False`, I would expect count to hit the Elasticsearch endpoint; instead an Exception is raised

**Steps to reproduce**:
Set `request_cache=True`, and request to get count, and the following will result in an error.

```
es.count(index=self._index, body=d, **self._params)["count"]
---
/usr/local/lib/python3.6/dist-packages/elasticsearch/client/utils.py in _wrapped(*args, **kwargs)
    150                 if p in kwargs:
    151                     params[p] = kwargs.pop(p)
--> 152             return func(*args, params=params, headers=headers, **kwargs)
    153 
    154         return _wrapped

TypeError: count() got an unexpected keyword argument 'request_cache'
```

**Linked ticket:  https://github.com/elastic/elasticsearch-dsl-py/issues/1429**

NavyaZaveri:
Fixes #1391

### DIFF
--- a/elasticsearch/client/__init__.py
+++ b/elasticsearch/client/__init__.py
@@ -54,7 +54,6 @@ from .ssl import SslClient
 from .transform import TransformClient
 from .watcher import WatcherClient
 
-
 logger = logging.getLogger("elasticsearch")
 
 
@@ -485,6 +484,7 @@ class Elasticsearch(object):
         "min_score",
         "preference",
         "q",
+        "request_cache",
         "routing",
         "terminate_after",
     )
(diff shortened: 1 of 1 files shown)


NavyaZaveri:
Ah, I've just realized that the changes have to first be made on the native elasticsearch layer (i.e, the [rest api spec](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec/src/main/resources/rest-api-spec/api)), and then codegen'd for the py client.  Closing this PR as it probably isn't the right way to go about fixing the issue

Issue #1416, opened 2020-10-28
ID with whitespaces stored incorrectly


ghost:
**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**: 6.8.1

**Description of the problem including expected versus actual behavior**:

When indexing a document with an ID with an whitespace, the ID does end up stored with the `+` character instead of a whitespace. This breaks the consistency of the client API. Looking at the example used below, `count` and `search` operations do not come back with the expected result.

Digging a bit deeper I could find that the library doesn't seem to be compliant with the https://tools.ietf.org/html/rfc3986#section-2.1 standard when it comes to escaping the ID (part of the URI path component). It is using `+` to escape, which is wrong, whitespaces are escaped using `%20`.

**Steps to reproduce**:

Run this test. Setup env vars: ELASTIC_ADDRESS, ELASTIC_USER and ELASTIC_PASS

```python
import os

from elasticsearch import Elasticsearch


def test_example():
    under_test = Elasticsearch(os.environ.get("ELASTIC_ADDRESS"), http_auth=(
        os.environ.get("ELASTIC_USER"),
        os.environ.get("ELASTIC_PASS")
    ))

    index = 'wrong-escaping'
    example_id = "One Example"

    under_test.indices.create(index=index, ignore=400)
    under_test.index(index=index, id=example_id, body={"field": "value"}, doc_type="doc")

    result = under_test.get(index=index, id=example_id, doc_type="doc")
    assert result['_id'] == example_id

    search_result = under_test.search(index=index, q="_id:{}".format(example_id), doc_type="doc")
    assert search_result['hits']['total']['value'] == 1

    count_result = under_test.count(index=index, q="_id:{}".format(example_id), doc_type="doc")
    assert count_result == 1
```

sethmlarson:
Yep, agree that this is an issue. Are you up for creating a PR against the `6.x` branch with a test case?

sethmlarson:
Closed in https://github.com/elastic/elasticsearch-py/pull/1535

PR #1417, opened 2020-10-28
#1416 Using quote instead of quote_plus to escape path params

joaomelohanzo:


sethmlarson:
Closing in favor of https://github.com/elastic/elasticsearch-py/pull/1535, thanks for this contribution.

PR #1434, opened 2020-11-13
Extend the serializer of elasticsearch-py by adding a CSVSerializer c…

danielpine:
…lass to support CSV when using es.sql.query(body,format='csv'') .


sethmlarson:
Is the email you used for the CLA the same as the one on the commits?

PR #1438, opened 2020-11-14
Added a CSVSerializer class in serializer.py to support CSV mimetype

danielpine:
when using es.sql.query(body,format='csv') ,
and added test cases for csv.


mrbaguvix (approved):

sethmlarson:
Just to fill you in on my plans, instead of a `CSVSerializer` I'm planning on making the `TextSerializer` accept `text/*` and adding support for wildcards in the `content_type` field of serializers.

danielpine:
> Just to fill you in on my plans, instead of a `CSVSerializer` I'm planning on making the `TextSerializer` accept `text/*` and adding support for wildcards in the `content_type` field of serializers.

It's more wise

danielpine:
If you use 'text/*' as the key, you need to convert the mimetype startwith text/  to 'text/*'

PR #1443, opened 2020-11-18
Fix typo in collections.abc import path

Tenzer:
The module to import from it called "collections.abc", not "collections.abs".

This has worked due to the fallback of importing from "collections" instead, in
case the import from "collections.abs" failed. The fallback on importing from
"collections" will however stop working with Python 3.9 which then makes this a
breaking bug.


sethmlarson (approved):
Thanks for fixing this, LGTM

PR #1444, opened 2020-11-19
Fix typo: collections.abs -> collections.abc

earwig:
`collections.abs` does not exist, and surely `collections.abc` was meant instead. (Introduced in #1387.)

This causes a warning in recent versions of Python 3.


earwig:
I should have noticed #1443. Whoops!

PR #1494, opened 2020-12-27
Fix #1492

### LINKED ISSUE #1492: [feature request] support change http_compress level.
**Describe the feature**:  

The http_compress argument for the Elasticsearch class  uses **_COMPRESS_LEVEL_BEST** by default.  I want to change it according to the need.

yumendy:
Add http_compress_level argument for the Connection class.


sethmlarson:
Thanks for this suggestion, however I think I'm going to decline support of this option. I'm going to recommend using a custom HTTPConnection for this use-case.

PR #1509, opened 2021-01-13
fix link to async.html

gianfa:


sethmlarson:
This actually needs fixing on the `7.x` and `7.11` branches, set them both to be `async.html`

Issue #1265, opened 2020-05-29
Implement the 'contains', 'transform_and_set', and 'arbitrary_key' features


sethmlarson:
We're missing three features for the REST API spec tests: `contains`, `transform_and_set`, and `arbitrary_key` for parity with upstream. We should implement these features so we can run all REST API spec tests.

There is a description of each feature here: https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/test/README.asciidoc

V1NAY8:
@sethmlarson Any inputs on this change? I am thinking of trying this. :)

sethmlarson:
@V1NAY8 That'd be awesome! Each of the features is documented within [this document](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/test/README.asciidoc). I'd recommend creating a PR for each one individually.

The easiest is likely `contains`. Look in `test_elasticsearch/test_server/test_rest_api_spec.py`, this is where the code should go. You would add a `run_contains` function to `YamlRunner` and then implement the functionality there. Does this help? :)

V1NAY8:
Yep, It helps 😃

PR #1511, opened 2021-01-14
Implement run_contains for REST API Tests

V1NAY8:
Partially does #1265 

- Implemented `run_contains` method.
- There is only one example of `yml` file `elasticsearch\rest-api-spec\src\main\resources\rest-api-spec\test\cluster.voting_config_exclusions\10_basic.yml`
- Also added `.vscode` folder to `.gitignore`. If that is not required. I'll remove it.
@sethmlarson Please review this. Let me know if any changes are required. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,6 +140,9 @@ cython_debug/
 # Pycharm project settings
 .idea
 
+# VSCode project settings
+/.vscode
+
 # elasticsearch files
 test_elasticsearch/cover
 test_elasticsearch/local.py
--- a/.gitignore
+++ b/.gitignore
@@ -140,9 +140,6 @@ cython_debug/
 # Pycharm project settings
 .idea
 
-# VSCode project settings
-/.vscode
-
 # elasticsearch files
 test_elasticsearch/cover
 test_elasticsearch/local.py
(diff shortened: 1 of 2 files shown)


sethmlarson:
jenkins test this please

.gitignore
```diff
@@ -140,6 +140,9 @@ cython_debug/
 # Pycharm project settings
 .idea
 
+# VSCode project settings
+/.vscode
```
sethmlarson, line 144: Let's remove this, instead you can add your editor files into a [global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c).

test_elasticsearch/test_server/test_rest_api_spec.py
```diff
@@ -275,6 +276,14 @@ def run_match(self, action):
             else:
                 assert expected == value, "%r does not match %r" % (value, expected)
 
+    def run_contains(self, action):
+        for path, expected in action.items():
+            value = self._lookup(path)  # list[dict[str,str]] is returned
+            expected = self._resolve(expected)  # dict[str, str]
+
+            if expected not in value:
+                raise AssertionError(f"{expected} does not match {value}")
```
sethmlarson, line 285: The assertation message here could be better, something like `"%s is not contained by %s" % (expected, value)`

Also we can't use f-strings yet for this project, we still support Python <3.6

sethmlarson (changes requested):
Thanks for this! Couple of changes for you, if you don't get to them before I do I can manually make commits too :)

V1NAY8:
Yes, made the changes.

sethmlarson (approved):
LGTM, thank you!

PR #1519, opened 2021-01-26
Allow for multiple test YAML dirs by default

sethmlarson:
This allows testing OSS and X-Pack against an X-Pack instance in CI.

Looked at git history and it seems that previously we weren't executing X-Pack integration tests and this change enables that by default.


sethmlarson:
Closing this as will require a different approach now that specs+yaml tests are delivered via the artifacts API.

Issue #1027, opened 2019-10-03
bulk() and streaming_bulk() helpers don't apply the `ignore` argument to op results


jboning:
The documentation for `ignore` indicates:

> An API call is considered successful (and will return a response) if elasticsearch returns a 2XX response. Otherwise an instance of TransportError (or a more specific subclass) will be raised. You can see other exception and error states in Exceptions. If you do not wish an exception to be raised you can always pass in an ignore parameter with either a single status code that should be ignored or a list of them

The `bulk()` and `streaming_bulk()` helpers parameters accept the `ignore` argument and pass it through to the underlying API call, but don't otherwise consider it while processing results. It would be more useful (and more consistent with how `ignore` works with other client methods) to apply the value to individual op results, treating ops with ignored statuses as successful.

jboning:
If this is desirable, I'd be willing to assemble a PR!

HonzaKral:
The `bulk` and `streaming_bulk` helpers have different mechanisms on how to control the flow in case of errors (`raise_on_error` and `raise_on_exception`) so they don't use the `ignore` parameter.

I understand this is confusing and perhaps we could document it but overall I do believe it to be the correct behavior.

Thank you for raising the issue!

arusahni:
I'd love to have some client-side filter for the status codes - for example, I may want to ignore version conflict errors (409), but not others. In order to reduce confusion (and boiler-plate across various code-bases), would a `dont_raise_on_status=` kwarg be more palatable?

HonzaKral:
@arusahni that is a great idea, thank you! Just not sure about the best naming. But definitely doable and quite simple to do.

Reopening the issue to keep track of this.

Sparkycz:
@sethmlarson Guess you forgot to close the issue...

PR #1553, opened 2021-03-28
Add `ignore` kwarg to bulk/streaming_bulk to ignore error status codes

Sparkycz:
Solves https://github.com/elastic/elasticsearch-py/issues/1027

### DIFF
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -161,7 +161,7 @@ def _chunk_actions(actions, chunk_size, max_chunk_bytes, serializer):
         yield ret
 
 
-def _process_bulk_chunk_success(resp, bulk_data, ignore, raise_on_error=True):
+def _process_bulk_chunk_success(resp, bulk_data, ignore_status, raise_on_error=True):
     # if raise on error is set, we need to collect errors per chunk before raising them
     errors = []
 
@@ -172,7 +172,7 @@ def _process_bulk_chunk_success(resp, bulk_data, ignore, raise_on_error=True):
         status_code = item.get("status", 500)
 
         ok = 200 <= status_code < 300
-        if not ok and raise_on_error and status_code not in ignore:
+        if not ok and raise_on_error and status_code not in ignore_status:
             # include original document source
             if len(data) > 1:
                 item["data"] = data[1]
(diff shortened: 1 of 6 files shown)


elasticsearch/helpers/actions.py
```diff
@@ -221,6 +223,7 @@ def _process_bulk_chunk(
     bulk_data,
     raise_on_exception=True,
     raise_on_error=True,
+    ignore=(),
```
sethmlarson, line 226: Instead of `ignore` let's use `ignore_status`, this is what the parameter will be called on APIs in 8.x onwards and will be deprecated in 7.x.

elasticsearch/helpers/actions.pyi
```diff
@@ -61,6 +61,7 @@ def streaming_bulk(
     initial_backoff: Union[float, int] = ...,
     max_backoff: Union[float, int] = ...,
     yield_ok: bool = ...,
+    ignore: Collection[int] = ...,
```
sethmlarson, line 64: Should be `Optional[Union[int, Collection[int]]]`

sethmlarson (changes requested):
Thanks for this! This change will also need to be made to `elasticsearch/_async/helpers.py`

sethmlarson:
jenkins test this please

test_elasticsearch/test_async/test_server/test_helpers.py
```diff
@@ -332,6 +332,41 @@ async def test_error_is_raised(self, async_client):
         with pytest.raises(helpers.BulkIndexError):
             await helpers.async_bulk(async_client, [{"a": 42}, {"a": "c"}], index="i")
 
+    def test_ignore_error_if_raised(self, async_client):
+        # ignore the status code 400 in tuple
+        helpers.async_bulk(
```
sethmlarson, line 337: All of the calls to `async_bulk` need to be `await`-ed and run within an `async def test_...`

sethmlarson (changes requested):
One last thing, then this is looking good.

test_elasticsearch/test_async/test_server/test_helpers.py
Sparkycz: Thanks for the great review.

Sparkycz:
Hello @sethmlarson, please, what is the PR waiting for? Should I change the destination branch or what to close the PR? :)

sethmlarson (approved):
LGTM

PR #1573, opened 2021-04-15
Fix API doc reference to Cluster GET Settings page

DJRickyB:
Hi, was using the API doc and noticed this one link was out-of-place


sethmlarson:
Thanks for this, unfortunately this fix needs to be made to the elastic/elasticsearch repository as these URLs are auto-generated from [these files](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/cluster.get_settings.json). Could you make the change there and then eventually we'll receive the fix.

PR #1589, opened 2021-05-03
Fix link to JSONSerializer in docs

akiomik:


mrbaguvix (approved):

sethmlarson (approved):
Since this URL is likely to become stale in the future due to normal changes to the code I just removed it.

PR #1592, opened 2021-05-15
Need to propagate headers that the scan API receives, to the scroll & clear_scroll calls it subsequently makes

ashokc:
When headers (say an oauth2 token for authentication, or an api-key) are passed to the helper scan API, that header token/api-key should be passed down to the scroll & clear_scroll calls. Otherwise an authentication error ensues. The error message one gets is like: 

AuthenticationException(401, 'security_exception',

{
  "error": {
    "root_cause": [
      {
        "type": "security_exception",
        "reason": "missing authentication credentials for REST request [/_search/scroll]",
        "header": {
          "WWW-Authenticate": [
            "Bearer realm=\"security\"",
            "ApiKey"
          ]
        }
      }
    ],
    "type": "security_exception",
    "reason": "missing authentication credentials for REST request [/_search/scroll]",
    "header": {
      "WWW-Authenticate": [
        "Bearer realm=\"security\"",
        "ApiKey"
      ]
    }
  },
  "status": 401
}

The fix is simply to pull the supplied headers if any in kwargs (to scan api) and add supply them to the scroll & clear_scroll calls


sethmlarson:
By the way I am planning on implementing this, I appreciate the PR but might do a slightly different implementation. I'll take a look next week

mrbaguvix (approved):

ashokc:
Great thanks, Seth.

Issue #1602, opened 2021-06-04
When index method is called in recursion it kills python interpreter 


hmilkovi:
Hi all,

Versions:
- elasticsearch==7.13.0
- Python 3.8.6

Expected:
- Raise any kind of error

Behavior:
- Fatal Python error: Cannot recover from stack overflow.
- Doesn't respect sys.getrecursionlimit() and gets called ~1029 in test case 1000 is the recursion limit

To reproduce:
```
import os
import sys
import inspect
import elasticsearch

es_client = elasticsearch.Elasticsearch(
    [os.environ["ELASTIC_API"]],
    verify_certs=False,
)
print(sys.getrecursionlimit())
def rec():
    print(len(inspect.stack()))
    es_client.index(index='testing', body={'test': 'test'}, request_timeout=1)
    rec()

rec()
````

**Comment by sethmlarson** [comment]
I'm not sure there's much we can do here, the function itself is infinitely recursive so of course you're going to hit the recursion limit eventually? What are we supposed to fix here?

**Comment by hmilkovi** [comment]
Hi @sethmlarson, one thing is exception (hiting the limit) `RecursionError` is not the case, the case is core dump and whole interpreter trashes (going over the limit and then core dump).

Use case where we in recursion to index 2000 documents -> won't work.

Also issue is similar to this one that was solved: https://bugs.python.org/issue36272

I have fixed it in my code with:
```
if len(inspect.stack()) >= sys.getrecursionlimit():
            raise RecursionError
```

The issue is that https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/http_urllib3.py#L256  RecursionError it doesn't get propagated it's needed to be.




In python recursion and even change it's limit like:
`sys.setrecursionlimit(Newlimit)`
and here it get's interesting as it'd doesn't core dump until it exceeds Pythons recursion limit.


Tested cases:
1. When left default limit (1000) it crashes on 1029
2. When I set limit to 1500 it crashes on 1529
3. When I set limit to 2 it doesn't crash interpreter and returns expected RecursionError
4. When I set it to 100 it crashes on 129

Please consider my PR: https://github.com/elastic/elasticsearch-py/pull/1607

```
Recursion limit before: 1000
Recursion limit After: 100
2
3
....
126
127
128
129
Fatal Python error: Cannot recover from stack overflow.
Python runtime state: initialized

Current thread 0x00007f0848578740 (most recent call first):
[...]

PR #1607, opened 2021-06-04
Handle RecursionError exception in perform_request (issue 1602)

hmilkovi:
Propagate RecursionError if it happends on lower levels so when used like:
```
import sys
import inspect
import elasticsearch

es_client = elasticsearch.Elasticsearch("http://localhost:9200/")
print(f"Recursion limit: {sys.getrecursionlimit()}")

def rec():
    print(len(inspect.stack()))
    es_client.index(index='testing', body={'test': 'test'}, request_timeout=1)
    rec()

rec()
```
doesn't crash python interpreter with core dump.

Personal opinion is that this is really needed when we you web servers like uwsgi, gunicorn as
then the whole web server dies and doesn't return http 500 status code.

Related issue: https://github.com/elastic/elasticsearch-py/issues/1602

### DIFF
--- a/elasticsearch/_async/http_aiohttp.py
+++ b/elasticsearch/_async/http_aiohttp.py
@@ -303,7 +303,8 @@ class AIOHttpConnection(AsyncConnection):
         # We want to reraise a cancellation.
         except asyncio.CancelledError:
             raise
-
+        except RecursionError:
+            raise
         except Exception as e:
             self.log_request_fail(
                 method,
@@ -319,8 +320,6 @@ class AIOHttpConnection(AsyncConnection):
                 e, (asyncio.TimeoutError, aiohttp_exceptions.ServerTimeoutError)
             ):
                 raise ConnectionTimeout("TIMEOUT", str(e), e)
-            if isinstance(e, RecursionError):
-                raise RecursionError("RECURSION", str(e), e)
             raise ConnectionError("N/A", str(e), e)
 
         # raise warnings if any from the 'Warnings' header.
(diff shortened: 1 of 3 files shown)


sethmlarson:
Gotcha, so the problem really is that we're swallowing the `RecursionError` out of `except Exception:`? If that's the case let's do something simpler, instead the very first branch under `except Exception:` check for `RecursionError` and reraise without modifying the exception.

hmilkovi:
Yes that's the case :) . Does accepting CLA take some time to propagate or as I have signed it?

sethmlarson:
@hmilkovi Did you sign the CLA with the same email address as the one used in your Git commits?

elasticsearch/_async/http_aiohttp.py
```diff
@@ -303,7 +303,8 @@ async def perform_request(
         # We want to reraise a cancellation.
         except asyncio.CancelledError:
             raise
-
+        except RecursionError:
```
sethmlarson, line 306: Let's use the tuple here ie:

```python
except (asyncio.CancelledError, RecursionError):
    raise
```

hmilkovi: Yes are right, done.

hmilkovi:
CLA check (read up in docs that I need to comment again)

hmilkovi:
I have changed email on my last commit. Do I need to change it on all?
I want to contribute as private entity with private email.

Sorry for bothering you with that.

sethmlarson:
Yeah creating a new PR is probably best

hmilkovi:
Closing this one for https://github.com/elastic/elasticsearch-py/pull/1609 .
Now CLA is fine :)

PR #1609, opened 2021-06-06
Handle RecursionError exception in perform_request

hmilkovi:
Propagate RecursionError if it happends on lower levels so when used like:
```
import sys
import inspect
import elasticsearch

es_client = elasticsearch.Elasticsearch("http://localhost:9200/")
print(f"Recursion limit: {sys.getrecursionlimit()}")

def rec():
    print(len(inspect.stack()))
    es_client.index(index='testing', body={'test': 'test'}, request_timeout=1)
    rec()

rec()
```
So it doesn't crash python interpreter with core dump and raises as expected RecursionError.

Related issue: #1602


sethmlarson (approved):

sethmlarson:
Thanks much!!

PR #1623, opened 2021-06-28
Verify connection to Elasticsearch

sethmlarson:
Follows this logic for verifying a connection to Elasticsearch:

Before the first API request:
1. Make an API request to `/`, inspect the response:
2. If call to `/` fails with `401` or `403` pass the check and show a warning (message will be linked later). This happens if the  `monitor` permission missing for user. The subsequent checks must be ignored.
2. If there's no `version` field or if the `version` field value is <6.0.0 raise an error.
3. If there is a `version` field and it's >= 6.0.0 and <7.0.0:
    - If there is no `tagline` field or if the `tagline` field value isn't `You know, for Search` raise an error.
4. If there is a `version` field and it's >=7.0.0 and <7.14.0:
    - If there is no `tagline` field or if the `tagline` field value isn't `You know, for Search` raise an error.
    - If there is no `build_flavor` field or if the `build_flavor` field value isn't `default` raise an error.
5. If there is a `version` field and it's >=7.14.0:
    - If there is no `X-Elastic-Product` HTTP header in the response or if the `X-Elastic-Product` HTTP header value isn't `Elasticsearch` raise an error.


lnielsen:
It's disappointing to see this :-( You're forcing us as bystanders in a battle to choose sides. We develop an Open Source product that could likely easily work with both Elasticsearch or OpenSearch and then the users can choose for themselves if they want Elasticsearch or OpenSearch. Now, instead, we likely have to make choice for all our users if we want OpenSearch or Elasticsearch. This and other behaviours from Elastic really does not give me any confidence in Elastic and what you might do in the future. And don't blame it all on Amazon - you've already changed the server license, you didn't have to make this move :-(

philkra:
Amazon OpenSearch is a different product. And while there is some shared history, there are already many differences that cause real confusion and issues.

ruurd:
> Amazon OpenSearch is a different product. And while there is some shared history, there are already many differences that cause real confusion and issues.

In fact, ElasticSearch beyond 7.10 is a different product. Why are people fucking over an open source client to cater to the ES traitors that changed licensing on everyone?

PR #1638, opened 2021-07-14
Allow integration tests to be run from IDE

sethmlarson:


sethmlarson:
We're finally passing! :tada:

PR #1645, opened 2021-07-16
Support data stream re-indexing 

### LINKED ISSUE #1613: Support using the reindex helper to target a datastream
**Describe the feature**: The `elasticsearch.helpers.reindex` function currently does not support re-indexing into a data stream. If you run something like the following code:

```python
helpers.reindex(
    client=source_client,
    source_index=source_index,
    target_client=target_client,
    target_index=target_data_stream,        
    )
```

it will return a `BulkIndexError` with the error `{'type': 'illegal_argument_exception', 'reason': 'only write ops with an op_type of create are allowed in data streams'}`. Per the Elastic API documentation for `reindex` under the [query parameters](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html#docs-reindex-api-query-params): `To reindex to a data stream destination, this argument must be create.`

Would be great if this option was available in the helper!

V1NAY8:
Closes #1613 

@sethmlarson Take a look :)

### DIFF
--- a/elasticsearch/_async/helpers.py
+++ b/elasticsearch/_async/helpers.py
@@ -437,7 +437,7 @@ async def async_reindex(
     :arg scroll: Specify how long a consistent view of the index should be
         maintained for scrolled search
     :arg op_type: Explicit operation type. Defaults to '_index'. Data streams must
-        be set to 'create'. If not specified will auto-detect if target_index is a
+        be set to 'create'. If not specified, will auto-detect if target_index is a
         data stream.
     :arg scan_kwargs: additional kwargs to be passed to
         :func:`~elasticsearch.helpers.async_scan`
--- a/elasticsearch/_async/helpers.py
+++ b/elasticsearch/_async/helpers.py
@@ -457,7 +457,7 @@ async def async_reindex(
     kwargs = {"stats_only": True}
     kwargs.update(bulk_kwargs)
 
-    is_data_stream: bool = False
+    is_data_stream = False
     try:
         # Verify if the target_index is data stream or index
         data_streams = await target_client.indices.get_data_stream(target_index)
--- a/elasticsearch/_async/helpers.py
+++ b/elasticsearch/_async/helpers.py
@@ -406,9 +406,9 @@ async def async_reindex(
     target_client=None,
     chunk_size=500,
     scroll="5m",
+    op_type=None,
     scan_kwargs={},
     bulk_kwargs={},
-    op_type=None,
 ):
 
     """
@@ -436,11 +436,13 @@ async def async_reindex(
     :arg chunk_size: number of docs in one chunk sent to es (default: 500)
     :arg scroll: Specify how long a consistent view of the index should be
         maintained for scrolled search
+    :arg op_type: Explicit operation type. Defaults to '_index'. Data streams must
+        be set to 'create'. If not specified will auto-detect if target_index is a
+        data stream.
     :arg scan_kwargs: additional kwargs to be passed to
         :func:`~elasticsearch.helpers.async_scan`
     :arg bulk_kwargs: additional kwargs to be passed to
         :func:`~elasticsearch.helpers.async_bulk`
-    :arg op_type: Explicit operation type.
     """
     target_client = client if target_client is None else target_client
     docs = async_scan(
@@ -462,19 +464,20 @@ async def async_reindex(
     is_data_stream = False
     try:
         # Verify if the target_index is data stream or index
-        data_streams = await target_client.indices.get_data_stream(target_index)
-    except NotFoundError:
+        data_streams = await target_client.indices.get_data_stream(
+            target_index, expand_wildcards="all"
+        )
+        is_data_stream = any(
+            data_stream["name"] == target_index
+            for data_stream in data_streams["data_streams"]
+        )
+    except (TransportError, KeyError, NotFoundError):
         # If its not data stream, might be index
         pass
-    else:
-        for stream in data_streams["data_streams"]:
-            if target_index == stream["name"]:
-                is_data_stream = True
-                break
 
     if is_data_stream:
-        if op_type is not None and op_type != "create":
-            raise ValueError("Data Stream should have op_type as create")
+        if op_type not in (None, "create"):
+            raise ValueError("Data streams must have 'op_type' set to 'create'")
         else:
             op_type = "create"
 
--- a/elasticsearch/_async/helpers.py
+++ b/elasticsearch/_async/helpers.py
@@ -19,7 +19,7 @@ import asyncio
 import logging
 
 from ..compat import map
-from ..exceptions import TransportError
+from ..exceptions import NotFoundError, TransportError
 from ..helpers.actions import (
     _ActionChunker,
     _process_bulk_chunk_error,
@@ -408,6 +408,7 @@ async def async_reindex(
     scroll="5m",
     scan_kwargs={},
     bulk_kwargs={},
+    op_type=None,
 ):
 
     """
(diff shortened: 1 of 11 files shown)


elasticsearch/helpers/actions.py
```diff
@@ -659,18 +659,33 @@ def reindex(
...
+            if is_data_stream:
+                h["_op_type"] = "create"
             if "fields" in h:
                 h.update(h.pop("fields"))
             yield h
 
     kwargs = {"stats_only": True}
     kwargs.update(bulk_kwargs)
+
+    is_data_stream = False
```
sethmlarson, line 674: Instead of auto-detecting if `target_index` is a data_stream, instead we can provide a new parameter `op_type` where if it's set will set `_op_type` to that value and then document that `op_type="create"` is needed for Data Streams.

What do you think of this approach?

elasticsearch/_async/helpers.py
```diff
@@ -439,24 +440,47 @@ async def async_reindex(
...
+
+    is_data_stream = False
+    try:
+        # Verify if the target_index is data stream or index
+        data_streams = await target_client.indices.get_data_stream(target_index)
+    except NotFoundError:
+        # If its not data stream, might be index
+        pass
+    else:
+        for stream in data_streams["data_streams"]:
```
sethmlarson, line 470: Let's move everything in the `else:` block into the `try` block so we don't have to think about the `data_streams` name outside the `try-except`.

elasticsearch/_async/helpers.py
```diff
@@ -439,24 +440,47 @@ async def async_reindex(
...
+            if target_index == stream["name"]:
+                is_data_stream = True
```
sethmlarson, line 472: `is_data_stream = any(data_stream["name"] == target_index for data_stream in data_streams["data_streams"])`

elasticsearch/_async/helpers.py
```diff
@@ -439,24 +440,47 @@ async def async_reindex(
...
             yield h
 
     kwargs = {"stats_only": True}
     kwargs.update(bulk_kwargs)
+
+    is_data_stream = False
+    try:
+        # Verify if the target_index is data stream or index
+        data_streams = await target_client.indices.get_data_stream(target_index)
+    except NotFoundError:
```
sethmlarson, line 466: Let's be super careful here and also catch `TransportError` and `KeyError`?

`except (TransportError, KeyError):`

elasticsearch/_async/helpers.py
```diff
@@ -408,6 +408,7 @@ async def async_reindex(
     scroll="5m",
     scan_kwargs={},
     bulk_kwargs={},
+    op_type=None,
```
sethmlarson, line 411: Move this parameter to be above `scan_kwargs` both here and in docstring

elasticsearch/_async/helpers.py
```diff
@@ -439,24 +440,47 @@ async def async_reindex(
         :func:`~elasticsearch.helpers.async_scan`
     :arg bulk_kwargs: additional kwargs to be passed to
         :func:`~elasticsearch.helpers.async_bulk`
+    :arg op_type: Explicit operation type.
```
sethmlarson, line 443: Mention that the default is `index` but `create` is required for data streams and will auto-detect if `target_index` is a data stream.

elasticsearch/_async/helpers.py
```diff
@@ -439,24 +440,47 @@ async def async_reindex(
...
                 h.update(h.pop("fields"))
             yield h
 
     kwargs = {"stats_only": True}
     kwargs.update(bulk_kwargs)
+
+    is_data_stream = False
+    try:
+        # Verify if the target_index is data stream or index
+        data_streams = await target_client.indices.get_data_stream(target_index)
```
sethmlarson, line 465: Need to add `expand_wildcards="all"` so hidden data streams are shown too.

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -754,3 +757,62 @@ def test_children_are_reindexed_correctly(self):
...
+        )
+    sync_client.bulk(bulk, refresh=True)
+    sync_client.indices.put_index_template(
+        "my-index-template",
+        body={
+            "index_patterns": ["py-*-*"],
+            "data_stream": {},
+        },
+    )
+    sync_client.indices.create_data_stream("py-test-stream")
```
sethmlarson, line 784: Use keyword arguments for all APIs

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -754,3 +757,62 @@ def test_children_are_reindexed_correctly(self):
...
+
+
+class TestDataStreamReindex(object):
+    @pytest.mark.parametrize("op_type", [None, "create"])
+    def test_reindex_index_datastream(
+        self, op_type, sync_client, reindex_data_stream_setup
+    ):
+        helpers.reindex(
+            sync_client,
+            "test_index_stream",
```
sethmlarson, line 795: Use keywords here too (except `client`)

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -754,3 +757,62 @@ def test_children_are_reindexed_correctly(self):
...
+            "py-test-stream",
+            query={"query": {"bool": {"filter": {"term": {"type": "answers"}}}}},
+            op_type=op_type,
+        )
+        sync_client.indices.refresh()
+        assert sync_client.indices.exists("py-test-stream")
```
sethmlarson, line 801: keyword argument

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -754,3 +757,62 @@ def test_children_are_reindexed_correctly(self):
...
+
+    def test_reindex_index_datastream_op_type_index(
+        self, sync_client, reindex_data_stream_setup
+    ):
+        with pytest.raises(
+            ValueError, match="Data Stream should have op_type as create"
+        ):
+            helpers.reindex(
+                sync_client,
+                "test_index_stream",
```
sethmlarson, line 814: see above

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -754,3 +757,62 @@ def test_children_are_reindexed_correctly(self):
             },
             q,
         )
+
+
+@pytest.fixture(scope="function")
+def reindex_data_stream_setup(sync_client):
```
sethmlarson, line 763: This is awesome! :tada:

test_elasticsearch/test_async/test_server/test_helpers.py
```diff
@@ -884,3 +885,67 @@ async def test_children_are_reindexed_correctly(
             "_version": 1,
             "found": True,
         } == q
+
+
+@pytest.fixture(scope="function")
```
sethmlarson, line 890: Apply all the comments from the sync tests here as well.

elasticsearch/helpers/test.py
```diff
@@ -73,6 +73,9 @@ def teardown_method(self, _):
         if self.es_version() >= (7, 7):
             expand_wildcards.append("hidden")
 
+        self.client.indices.delete_data_stream(
```
sethmlarson, line 76: :+1:

elasticsearch/helpers/actions.pyi
```diff
@@ -107,4 +107,5 @@ def reindex(
     scroll: str = ...,
     scan_kwargs: Optional[Mapping[str, Any]] = ...,
     bulk_kwargs: Optional[Mapping[str, Any]] = ...,
+    op_type: bool = ...,
```
sethmlarson, line 110: `str`

sethmlarson (changes requested):
Thanks for this, I've added a bunch of comments for you:

sethmlarson (approved):
LGTM, thank you!!

Issue #1647, opened 2021-07-17
Group similar tests and convert tests to use pytest framework


V1NAY8:

V1NAY8:
Assign this to me

dannysepler:
Just released this and looking for folks to test it! https://github.com/dannysepler/pytestify

It's totally okay if it doesn't work out for you 👍

sethmlarson:
Going to close this as refactor work for the test suite isn't planned at this time.

PR #1656, opened 2021-08-01
WIP => Refactor Tests for test_client directory

V1NAY8:
Converts tests  to use pytest framework and optimize them.  

Still WIP

related_task : "Task"  =  #1647


sethmlarson:
jenkins test this please

sethmlarson:
jenkins test this please

test_elasticsearch/test_client/common.py
```diff
@@ -36,15 +36,8 @@ def perform_request(self, method, url, params=None, headers=None, body=None):
...
-
-    def assert_call_count_equals(self, count):
-        assert count == self.client.transport.call_count
-
-    def assert_url_called(self, method, url, count=1):
-        assert (method, url) in self.client.transport.calls
-        calls = self.client.transport.calls[(method, url)]
-        assert count == len(calls)
-        return calls
+def assert_helper(client, method, url, count=1):
```
sethmlarson, line 39: I think I prefer having a `DummyTransportTestCase` over the `assert_helper()` because it's clear that the `DummyTransportTestCase` knows about the `DummyTransport` class.

sethmlarson (changes requested):
One comment, otherwise this PR looks good to me.

sethmlarson:
Jenkins test this please

test_elasticsearch/test_client/common.py
sethmlarson: I don't think this review comment is resolved yet, I still see usages of `assert_helper()`

V1NAY8: Its part of `DummyTransportTestCase` Now we don't pass the client explicitly to it. 
```python
class DummyTransportTestCase:
    def setup_method(self, _):
        self.client = Elasticsearch(transport_class=DummyTransport)

    def assert_helper(self, method, url, count=1):
        calls = self.client.transport.calls
        assert (method, url) in calls
        assert count == len(calls[(method, url)])
        return calls[(method, url)]
```

pquentin:
Jenkins test this please

V1NAY8:
@sethmlarson   Is this looking good? 😀

PR #1675, opened 2021-08-09
Remove product verification

sftim:
Allow requests against any Elasticsearch-like API.

This reverts pull request https://github.com/elastic/elasticsearch-py/pull/1623

PR #1678, opened 2021-08-10
add disable_product_check argument to transport classes

OriHoch:
This PR adds the option to disable the product check for users who wish to do that (whether for performance reasons or if they are willing to take the risk of using an unsupported product)

It adds a `disable_product_check` argument to the transport class (sync and async)

The change is minimal, utilizing the existing `_verified_elasticsearch` variable and adding an additional `False` state to it which means that product check is disabled. Because the `_ProductChecker.check_product` method returns an integer, this new state does not interfere with the existing product check (if not disabled)

Tests were updated accordingly.

### DIFF
--- a/elasticsearch/_async/transport.py
+++ b/elasticsearch/_async/transport.py
@@ -80,6 +80,7 @@ class AsyncTransport(Transport):
             don't support passing bodies with GET requests. If you set this to
             'POST' a POST method will be used instead, if to 'source' then the body
             will be serialized and passed as a query parameter `source`.
+        :arg disable_product_check: If True will not perform the product check.
 
         Any extra keyword arguments will be passed to the `connection_class`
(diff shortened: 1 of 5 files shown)


sethmlarson:
I've already closed an issue that requested this feature, it won't be implemented at this time.

OriHoch:
thanks, would be good to have an explanation as to why.. it was already implemented here, so I guess not much additional work is needed to add it

gris-gris:
@OriHoch 

They're just fighting against AWS, scared probably

Disabling compatibility with AWS is the part of the plan, they're will close any PRs with "disabling product check"

Just use an older version of library, sadly

sftim:
Similar to https://github.com/elastic/elasticsearch-py/issues/1675

PR #1702, opened 2021-08-31
Fix Pipelines

V1NAY8:
Doing this change because the yml has
```yml
  # We change the password to the original one using the hash
  - do:
      security.change_password:
        username: "joe"
        body:  >
          {
            "password_hash" : "$hash"
          }
```


sethmlarson:
jenkins test this please

V1NAY8:
This will fix `test_elasticsearch/test_server/test_rest_api_spec.py::test_rest_api_spec[change_password/10_basic[1]]`

sethmlarson (approved):
LGTM

PR #1710, opened 2021-09-07
documentation python block

ahmadmustafaanis:
image

This Python code block is not inside codes.


pquentin:
@ahmadmustafaanis Your change is unrelated to the problem you're reporting, you only added a VS Code file (which probably should be removed)

pquentin:
@ahmadmustafaanis It's a valid issue though! Are you still interested in submitting a fix?

PR #1724, opened 2021-09-16
Run API version compatibility tests on Jenkins

sethmlarson:


sethmlarson:
So this is good, I've identified a bug in our usage of the compatibility `Accept` header, we also need to send a `Content-Type: application/vnd.elasticsearch+json;compatible-with=X` when we send a JSON body.

We also need to not send `Accept: application/vnd.elasticsearch+json;compatible-with=X` on every API as we're currently doing now as that changes the behavior of APIs like `cat.*`. Need to send the proper headers that are mentioned in the specs and replace `application/json` with the compat header when we're in that mode.

sethmlarson:
Going to create a separate PR for the fix of the compatibility header.

PR #1727, opened 2021-09-21
Add term-missing to pytest

V1NAY8:
This PR adds term-missing parameter to `pytest-cov` to find out what lines we are missing coverage.

Also excludes some lines which probably don't need tests.

@sethmlarson Take a look and once run Jenkins

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -38,8 +38,7 @@ def test(session):
     junit_xml = os.path.join(SOURCE_DIR, "junit", "elasticsearch-py-junit.xml")
     pytest_argv = [
         "pytest",
-        "--cov-report",
-        "term-missing",
+        "--cov-report=term-missing",
         "--cov=elasticsearch",
         "--cov-config=setup.cfg",
         f"--junitxml={junit_xml}",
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -38,7 +38,10 @@ def test(session):
     junit_xml = os.path.join(SOURCE_DIR, "junit", "elasticsearch-py-junit.xml")
     pytest_argv = [
         "pytest",
+        "--cov-report",
```
sethmlarson, line 41: nit: use `--cov-report=term-missing` so everything shows on the same line.

setup.cfg
```diff
@@ -9,3 +9,8 @@ junit_family=legacy
 
 [tool:isort]
 profile=black
+
+[report]
+exclude_lines=
+    raise NotImplementedError
```
sethmlarson, line 15: You can use wildcards so `raise NotImplementedError*` should work I believe?

sethmlarson (changes requested):
Thanks for this! Definitely eager to get coverage better, two comments:

sethmlarson:
jenkins test this please

sethmlarson:
It'd be good to get this in 7.x as well, however it'd go in a slightly different place so can't backport directly. Can you handle this @V1NAY8?

V1NAY8:
Yes, I will do that.

Issue #1745, opened 2021-10-01
Move type information from type stubs to .py files


sethmlarson:
Since we're dropping Python 2.7 we can now safely move the type information in `.pyi` stubs to `.py` files.

V1NAY8:
Does this mean inline type hints?  Perhaps I can get started at get 5 files per PR.

sethmlarson:
@V1NAY8 I have an in-progress branch locally, luckily a lot of files are going away with the move to use [`elastic-transport-python`](https://github.com/elastic/elastic-transport-python/) so I'll skip those for now and remove them in another PR.

PR #1746, opened 2021-10-01
Remove the simplest .pyi stubs

sethmlarson:
Part of #1745


elasticsearch/exceptions.py
```diff
@@ -15,6 +15,8 @@
 #  specific language governing permissions and limitations
 #  under the License.
 
+from typing import Any, Dict, Type, Union
```
sethmlarson, line 18: Exceptions are being redone a bit with the move to elastic-transport hence the large amount of `type: ignore` for now.

Issue #1750, opened 2021-10-07
Performance regression in JSONSerializer.default() in v7.15.0


HenrikOssipoff:
**Elasticsearch version** (`bin/elasticsearch --version`): 7.13.1

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**: 7.15.0

**Description of the problem including expected versus actual behavior**:
We're a bit puzzled at this, but we've narrowed it into being related to us upgrading the Elasticsearch package from 7.14.0 to 7.15.0.

Upon upgrading, we experienced that all of our Elasticsearch calls rose significantly in latency - to the point where is cascaded across all of our systems. It took us a few days to figure out what was going on, but in the end, we simply downgraded the package to 7.14.0, and as seen from the graph here (taken from Elastic APM), it's quite apparent it had an effect:

![Skærmbillede 2021-10-07 kl  14 11 49](https://user-images.githubusercontent.com/4737224/136382786-4ca8352a-c56f-493c-a085-fa854964cd6e.png)

**Steps to reproduce**:
Honestly I'm not sure how to describe a reproducible flow. I'll be very happy to help debugging this against our production environment in any way possible, if someone has ideas on what to look for.

**To others who might be experiencing the same issue**: 
We've pinned our projects to 7.14.0 for now, as this effectively solves the issue for us.

HenrikOssipoff:
@sethmlarson Thanks for getting back to me so quick! Sorry for the late reply.

I'm not entirely sure how the HTTP Client logic works; we don't explicitly define one ourselves, so I assume it selects one based on installed packages? I've attached a `pip freeze` - the project doesn't use async:

```
amqp==5.0.6
anyio==3.3.2
argon2-cffi==21.1.0
asgiref==3.3.4
awesome-slugify==1.6.5
billiard==3.6.4.0
cachetools==4.2.2
celery==5.1.2
certifi==2020.12.5
cffi==1.14.6
chardet==4.0.0
charset-normalizer==2.0.6
click==7.1.2
click-didyoumean==0.0.3
click-plugins==1.1.1
click-repl==0.1.6
cool==3.1.24
coolshop-search-dsl==2.2.1
Django==3.2.8
django-healthz==0.0.5
django-redis==5.0.0
django-storages==1.11.1
djangorestframework==3.12.4
elastic-apm==6.5.0
elasticsearch==7.15.0
elasticsearch-dsl==7.4.0
google-api-core==2.1.0
google-auth==2.3.0
[...]

PR #1751, opened 2021-10-08
Only import numpy/pandas once in JSONSerializer.default()

sethmlarson:
Performance of `JSONSerializer.default()` was impacted when numpy/pandas wasn't installed. This changes the logic to only attempt importing numpy/pandas on the first call to `.default()` and afterwards no more import attempts will be made.

Closes https://github.com/elastic/elasticsearch-py/issues/1750


sethmlarson:
Backported this one manually to 7.x and 7.15.

PR #1752, opened 2021-10-10
fix body parameter documentation in index function

rbrunan:
Python lint with elasticsearch-py:7.15 was warning me about this parameter


rbrunan:
Contributor Agreement signed

PR #1785, opened 2021-11-04
Update bulk-ingest.py

Selenuix:
Fix typo in download_dataset() description.


sethmlarson:
Thanks for opening this. I've got plans to overhaul and add many examples so this would likely be lost during that effort. Going to close but you can take a look at other places to improve the client if you're still interested.

Issue #1680, opened 2021-08-11
Expose JSON body fields as typed parameters


sethmlarson:
## Motivation

Today with the Elasticsearch client you have to write a JSON blob to the `body` to pass any values within the HTTP body.
Unfortunately this is where some of the most complex data structures are for the Elasticsearch API (see query DSL, aggregations)
which means we're unable to provide a good window into this opaque object via types or auto-complete.

This is what writing a search query looks like today:

```python
from elasticsearch import Elasticsearch
client = Elasticsearch("https://localhost:9200")

client.search(
    index="test-index",
    size=10,
    body={
        "runtime_mappings": {
            "day_of_week": {
                "type": "keyword",
                "script": "emit(doc['@timestamp'].value.dayOfWeekEnum)"
            }
        },
        "aggs": {
            "day_of_week_count": {
                "value_count": {
                    "field": "day_of_week"
                }
            }
        }
    }
)
```

And here is what the type hints for the `search` API look like for today:

```python
def search(
    index: Optional[Any] = ...,
    body: Optional[Any] = ...,
    size: Optional[Any] = ...,
    ...,
): ...
```

Something to note for later: the `size` parameter is always serialized in the query string in this example.

## Proposed Solution

With the expanded JSON body fields implemented the same `search` API call can be written like so:

```python
client.search(
    index="test-index",
    size=10,
    runtime_mappings={
        "day_of_week": {
            "type": "keyword",
            "script": "emit(doc['@timestamp'].value.dayOfWeekEnum)"
        }
    },
    aggs={
        "day_of_week_count": {
            "value_count": {
                "field": "day_of_week"
            }
        }
    }
)
```

Notice how the body fields and fields that are serialized elsewhere like `index` are all at the same level and you're writing actual Python code instead of wrangling with JSON at the top-level of the API.

And here is what the function signature and types for the `search` API would look like:

```python
def search(
    index: Optional[Union[str, List[str]]] = ...,
    size: Optional[int] = ...,
    runtime_mappings: Optional[Mapping[str, Any]] = ...,
    aggs: Optional[Mapping[str, Any]] = ...
    ...,
    body: Optional[Any] = ...,
): ...
```

The differences that I want to highlight:

- There is still a `body` parameter for backwards compatibility with queries written before this change. This won't be a breaking change, any queries that are using the `body` parameter will still function exactly as they do today. The `body` parameter will be deprecated on APIs that support expanded body fields.
[...]

PR #1802, opened 2021-11-15
Switch to new API generator

sethmlarson:
- API generator now creates the complete function signature of each parameter
- Parameters are serialized inline instead of within a decorator
- Type hints aren't generated into separate `.pyi` files, instead are inline
- Changes API function decorator to expand `body` and `params` instead of old behavior of pulling parameters into `body` and `params`.
- Changed the sorting of APIs within a file to be alphabetical instead of based on previous file.
- Deletes the old API generator from `utils/`.

Closes https://github.com/elastic/elasticsearch-py/issues/1680


sethmlarson:
Published a fix in elastic-transport 8.0.0-alpha5.

PR #1805, opened 2021-11-20
Update API function decorator in CONTRIBUTING.md

ollipa:
It seems that API function decorator was renamed with the change to new API generator. CONTRIBUTING.md file was still referencing the old decorator name.


sethmlarson:
Thanks for this, I need to go through the contributing guidelines completely as the API generator and API specification have moved to new locations. I'll handle this when I tackle that.

PR #1845, opened 2021-12-24
Fix typos

kianmeng:


sethmlarson:
Thanks for opening. Unfortunately most of these files are automatically generated from other sources like:
- https://github.com/elastic/elasticsearch-specification
- https://github.com/elastic/elasticsearch

You'll have to fix the issues there. Either way, you need to sign the CLA with the same email as what is used in the commits. Let me know if you sign the CLA and update the PR to remove changes to automatically generated files under `elasticsearch/` and `docs/examples` and we can potentially merge the few fixes.

kianmeng:
@sethmlarson Noted. Will do so.

PR #1932, opened 2022-03-09
Change 'Connecting' section to make connecting to SOBD clusters easier

sethmlarson:
Users are reporting difficulties configuring the client with clusters that have generated certificates. This new "Connecting" introduction shows how to connect to Elastic Cloud or to a self-managed cluster with generated CA certificates instead of requiring further reading into the TLS section and "Getting Started with Elasticsearch" docs separately.


docs/guide/connecting.asciidoc
```diff
@@ -3,58 +3,155 @@
...
+
+https://www.elastic.co/guide/en/cloud/current/ec-getting-started.html[Elastic Cloud] is the easiest way to get started with Elasticsearch. When connecting to Elastic Cloud with the Python Elasticsearch client you should always use the `cloud_id` parameter to connect. You can find this value within the "Manage Deployment" page after you've created a cluster (look in the top-left if you're in Kibana).
+
+When you use the `cloud_id` parameter your client will be automatically configured for optimal use with Elastic Cloud including HTTPS and HTTP compression.
+
+```python
+from elasticsearch import Elasticsearch
+
+# Generated by Elasticsearch on creation of deployment
+ELASTIC_PASSWORD = "<password>"
```
pquentin, line 18: Should we suggest reading the password from an environment variable or is that out of scope?

docs/guide/connecting.asciidoc
```diff
@@ -3,58 +3,155 @@
...
-==== Connecting with URLs
+[[connect-self-managed-new]]
+=== Connecting to a self-managed cluster
 
-A single node can be specified via a `scheme`, `host`, `port`, and optional `path_prefix`. These values can either be specified manually via a URL in a string, dictionary, `NodeConfig`, or a list of these values. You must specify at least `scheme`, `host` and `port` for each node. All of the following are valid configurations:
+Starting in 8.0 the default configuration of an Elasticsearch cluster has changed to include Security and TLS being enabled with a generated CA and TLS certificates for each node. These changes mean you'll need to configure the Python Elasticsearch client to use HTTPS with the generated CA certificate in order to make requests successfully.
 
-[source,python]
-----------------------------
+If you're just getting started with Elasticsearch I recommend reading the documentation on https://www.elastic.co/guide/en/elasticsearch/reference/current/settings.html[configuring] and https://www.elastic.co/guide/en/elasticsearch/reference/current/starting-elasticsearch.html[starting Elasticsearch] to ensure your clusters is running as expected.
```
pquentin, line 40:
```suggestion
If you're just getting started with Elasticsearch I recommend reading the documentation on https://www.elastic.co/guide/en/elasticsearch/reference/current/settings.html[configuring] and https://www.elastic.co/guide/en/elasticsearch/reference/current/starting-elasticsearch.html[starting Elasticsearch] to ensure your cluster is running as expected.
```

docs/guide/connecting.asciidoc
```diff
@@ -3,58 +3,155 @@
...
-    {"scheme": "http", "host": "localhost", "port": 9200},
-    {"scheme": "http", "host": "localhost", "port": 9201},
-])
-----------------------------
+NOTE: If you don't specify `ca_certs` or `ssl_assert_fingerprint` then the https://certifiio.readthedocs.io[certifi package] will be used for `ca_certs` by default if available.
 
 [discrete]
-[[connect-ec]]
-==== Connecting to Elastic Cloud
+==== Verifying HTTPS with certificate fingerprints
```
pquentin, line 92: Should we add `(Python 3.10 or later)` to make it even more obvious than the NOTE?

docs/guide/connecting.asciidoc
```diff
@@ -3,58 +3,155 @@
...
+# Successful response!
+client.info()
+# {'name': 'instance-0000000000', 'cluster_name': ...}
+```
+
+[discrete]
+[[connect-no-security]]
+=== Connecting without security enabled
+
+WARNING: Running Elasticsearch without security enabled is strongly recommended against as it leaves your cluster open to data breaches, malware, and ransomware attacks.
```
pquentin, line 141: Is there an Elasticsearch docs page explaining how to disable security that we could link to?

pquentin (changes requested):
Thanks, this will definitely help users! I love the Python 3.10 option :)

docs/guide/connecting.asciidoc
```diff
@@ -3,58 +3,155 @@
...
-    cloud_id="cluster-1:dXMa5Fx..."
+# Fingerprint either from Elasticsearch startup or above script
+CERT_FINGERPRINT = "A5:2D:D9:35:11:E8:C6:04:5E:21:F1:66:54:B7:7C:9E:E0:F3:4A:EA:26:D9:F4:03:20:B5:31:C4:74:67:62:28"
+
+# Generated by Elasticsearch
+ELASTIC_PASSWORD = "<password>"
+
+client = Elasticsearch(
+  "https://localhost:9200",
+  # Colons and uppercase/lowercase don't matter
```
technige, line 126: Move this comment up to where `CERT_FINGERPRINT` is defined.

docs/guide/connecting.asciidoc
```diff
@@ -3,58 +3,155 @@
...
+```
+
+Take this SHA256 fingerprint value and pass it to the Python Elasticsearch client via `ssl_assert_fingerprint`:
+
+```python
 from elasticsearch import Elasticsearch
 
-es = Elasticsearch(
-    cloud_id="cluster-1:dXMa5Fx..."
+# Fingerprint either from Elasticsearch startup or above script
```
technige, line 118: Insert comment on colons and casing here (from below).

docs/guide/connecting.asciidoc
technige: I was thinking something similar: that we should link to a page on how to do this, and what the consequences of doing so are, so that users are free to make an informed choice. Probably this is better to discuss in a centralised document, rather than a client manual.

technige (changes requested):

docs/guide/connecting.asciidoc
```diff
@@ -3,58 +3,155 @@
...
-    "http://localhost:9200",
-    "http://localhost:9201",
-    "http://localhost:9202"
-])
+# Create the client instance
+client = Elasticsearch(
+  "https://localhost:9200",
+  ca_certs="/path/to/http_ca.crt",
+  basic_auth=("elastic", ELASTIC_PASSWORD)
+)
```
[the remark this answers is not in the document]
sethmlarson: Yeah it does, I'll add that section back.

docs/guide/connecting.asciidoc
```diff
@@ -3,58 +3,155 @@
...
+```
 
 [discrete]
-[[connect-url]]
-==== Connecting with URLs
+[[connect-self-managed-new]]
+=== Connecting to a self-managed cluster
 
-A single node can be specified via a `scheme`, `host`, `port`, and optional `path_prefix`. These values can either be specified manually via a URL in a string, dictionary, `NodeConfig`, or a list of these values. You must specify at least `scheme`, `host` and `port` for each node. All of the following are valid configurations:
+Starting in 8.0 the default configuration of an Elasticsearch cluster has changed to include Security and TLS being enabled with a generated CA and TLS certificates for each node. These changes mean you'll need to configure the Python Elasticsearch client to use HTTPS with the generated CA certificate in order to make requests successfully.
```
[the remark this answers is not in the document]
sethmlarson: Gotcha, I think that makes sense to optimize for users that are brand new instead of returning users.

szabosteve (approved):
I left some minor comments, otherwise LGTM! Thank you for writing all of these up!

PR #1935, opened 2022-03-22
Fixed wrong line of documentation in bulk update example

RamiAwar:
Spent an hour figuring this out. The way that was provided did not work, nor did close variations. This seems like the correct way of doing this!

### DIFF
--- a/docs/sphinx/helpers.rst
+++ b/docs/sphinx/helpers.rst
@@ -61,7 +61,7 @@ action (``_op_type`` defaults to ``index``):
         '_op_type': 'update',
         '_index': 'index-name',
         '_id': 42,
-        'doc': {'question': 'The life, universe and everything.'}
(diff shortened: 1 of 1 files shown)


sethmlarson:
Thanks @RamiAwar, could you sign the CLA with the same email as you have configured in git and let me know when you've done so?

RamiAwar:
I believe I already did!

sethmlarson:
Looking at this again, I don't think this change is correct. Our test suite uses `op_type=update doc={...}` but doesn't use `_source`. I'm going to close for now, if you could create a test case that fails without this change please let me know.

RamiAwar:
> 

I've made that change cause I spent so much time trying to get it to work with the latest version of es. The docs are not correct here. Idk how the test suites work but something is off somewhere.

The update I made worked for me but I'm not familiar with es internals to say that its a 100% correct. @sethmlarson

PR #2030, opened 2022-07-14
Fix timeout related parameters in the options() method

### LINKED ISSUE #1947: Calling options() on a client resets timeout related attributes
**Elasticsearch version** (8.1.1):

**`elasticsearch-py` version (`8.1.1`)**: 

When the `options()` method is called for an existing `ElasticSearch` client instance, the timeout related parameters are not preserved. These parameters are:

- `self._request_timeout`
- `self._max_retries`
- `self._retry_on_timeout`
- `self._retry_on_status`

This is problematic in the `helpers` module, particularly `scan()` because that method calls the `options` and resets any timeout related settings we may have made on the client. We can re-specify the `request_timeout` there but retries parameters cannot be replicated

**Steps to reproduce**:
```
>>> import elasticsearch as ES
>>> es=ES.Elasticsearch("http://esmasters:9200", sniff_on_connection_fail=True, sniff_on_start=True, min_delay_between_sniffing=600, request_timeout=600, sniff_timeout=300, max_retries=5, retry_on_timeout=True)
>>> es._retry_on_timeout
True
>>> es2=es.options()
>>> es2._retry_on_timeout
<DEFAULT>
>>> es._max_retries
5
>>> es2._max_retries
<DEFAULT>
>>> es2._retry_on_status
<DEFAULT>
>>> es._retry_on_status
<DEFAULT>
>>> es._request_timeout
600
>>> es2._request_timeout
<DEFAULT>
```

miriam-eid:
Closes #1947

### DIFF
--- a/elasticsearch/_async/client/__init__.py
+++ b/elasticsearch/_async/client/__init__.py
@@ -526,6 +526,8 @@ class AsyncElasticsearch(BaseClient):
             if isinstance(ignore_status, int):
                 ignore_status = (ignore_status,)
             client._ignore_status = ignore_status
+        else:
+            client._ignore_status = self._ignore_status
 
         if max_retries is not DEFAULT:
             if not isinstance(max_retries, int):
--- a/elasticsearch/_async/client/__init__.py
+++ b/elasticsearch/_async/client/__init__.py
@@ -519,6 +519,8 @@ class AsyncElasticsearch(BaseClient):
 
         if request_timeout is not DEFAULT:
             client._request_timeout = request_timeout
+        else:
+            client._request_timeout = self._request_timeout
 
         if ignore_status is not DEFAULT:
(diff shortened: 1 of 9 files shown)


elasticsearch/_async/client/__init__.py
```diff
@@ -519,6 +519,8 @@ def options(
 
         if request_timeout is not DEFAULT:
             client._request_timeout = request_timeout
+        else:
+            client._request_timeout = self._request_timeout
 
         if ignore_status is not DEFAULT:
```
sethmlarson, line 525: We need to add similar logic to `ignore_status` too.

sethmlarson (commented):
We could use some test cases for this logic to ensure we don't regress this behavior.

sethmlarson (approved):
LGTM, nice work.

PR #2087, opened 2022-10-13
Update version parameter

miriam-eid:
Update version parameter of the code generation automation.

### DIFF
--- a/.ci/make.sh
+++ b/.ci/make.sh
@@ -68,11 +68,8 @@ case $CMD in
         TASK_ARGS=("$VERSION" "$output_folder")
         ;;
     codegen)
-        if [ -v $VERSION ]; then
-            echo -e "\033[31;1mTARGET: codegen -> missing version parameter\033[0m"
-            exit 1
-        fi
-        echo -e "\033[36;1mTARGET: codegen API v$VERSION\033[0m"
+        VERSION=$(git rev-parse --abbrev-ref HEAD)
+        echo -e "\033[36;1mTARGET: codegen API $VERSION\033[0m"
         TASK=codegen
         # VERSION is BRANCH here for now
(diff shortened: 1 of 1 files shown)


picandocodigo (approved):

.ci/make.sh
```diff
@@ -68,11 +68,8 @@ case $CMD in
         TASK_ARGS=("$VERSION" "$output_folder")
         ;;
     codegen)
-        if [ -v $VERSION ]; then
-            echo -e "\033[31;1mTARGET: codegen -> missing version parameter\033[0m"
-            exit 1
-        fi
-        echo -e "\033[36;1mTARGET: codegen API v$VERSION\033[0m"
+        VERSION='main'
```
sethmlarson, line 71: Shouldn't this be variable based on the branch? We'll need to codegen on feature branches too.

picandocodigo: Since `branch` is a parameter in the GitHub Action that calls the script (and that's the branch that is going to be checked out by the action), I suggested setting it to `main` in the main branch, and backporting these changes to other branches with their corresponding branch values. Hope that makes sense :sweat_smile: 

In Ruby the branch is not a parameter, it builds the branch you're in, so it's different from the Python generator. So there could be a better solution.

sethmlarson: We can get the current branch from git with this: `$(git rev-parse --abbrev-ref HEAD)` (but will fail on non-main/feature branches or commits not at a branch head) maybe that could be a solution that doesn't need keeping up-to-date?

picandocodigo: Yes! We were discussing this option with @miriam-eid  :+1: 
So `VERSION=$(git rev-parse --abbrev-ref HEAD)` would work and doesn't need any changes when backported to feature branches, right?

picandocodigo (approved):

PR #2092, opened 2022-10-25
Update APIs to 7.17

miriam-eid:


elasticsearch/_async/client/__init__.py
```diff
@@ -305,7 +305,6 @@ async def info(self, params=None, headers=None):
         "version_type",
         "wait_for_active_shards",
         request_mimetypes=["application/json"],
-        response_mimetypes=["application/json"],
```
sethmlarson, line 308: Hmm, it's weird that `response_mimetypes` isn't being generated? Can we figure out why that's happening?

miriam-eid: Yes, I will check that out.

PR #2094, opened 2022-10-26
Update BulkIndexError hint

g39088902:
I added some information about the error code in the error trace output. Therefore, developers can more easily know the specific reasons for the failure of Bulk.


g39088902:
The error trace output was: elasticsearch.helpers.BulkIndexError: 20 document(s) failed to index.

Issue #2084, opened 2022-10-12
Missing some parameters in open_point_in_time


johtani:
**Elasticsearch version** : 8.x 

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**: 8.4.3

**Description of the problem including expected versus actual behavior**:
**Steps to reproduce**:

We can still use `routing` parameter with `open_point_in_time` in Elasticsearch 8.x,
however I got the TypeError from elasticsearch-py. I think it's a bug.
Traceback is here:
```
Traceback (most recent call last):
  File "es-py-sample\main.py", line 20, in <module>
    print_hi('PyCharm')
  File "es-py-sample\main.py", line 11, in print_hi
    res = es.open_point_in_time(index="wikidata",keep_alive="1m",routing="1")
  File "es-py-sample\venv\lib\site-packages\elasticsearch\_sync\client\utils.py", line 414, in wrapped
    return api(*args, **kwargs)
TypeError: open_point_in_time() got an unexpected keyword argument 'routing'
```

Until `7.17` elasticsearch-py, `routing` parameter and some other parameters support in `open_point_in_time`.
https://github.com/elastic/elasticsearch-py/blob/7.17/elasticsearch/client/__init__.py#L2301-L2325

After `8.0`, missing `routing` and some parameters.
https://github.com/elastic/elasticsearch-py/blob/8.0/elasticsearch/_sync/client/__init__.py#L2768

Elasticsearch 8.4 still support `routing`, `expand_wildcards`, `preferences` params in open_point_in_time API.
See [rest-api-spec in Elasticsearch repo](https://github.com/elastic/elasticsearch/blob/8.4/rest-api-spec/src/main/resources/rest-api-spec/api/open_point_in_time.json). Also these params are on [Point in time API Reference](https://www.elastic.co/guide/en/elasticsearch/reference/master/point-in-time-api.html) and [source in Es](https://github.com/elastic/elasticsearch/blob/be7c7415627377a1b795400fb8dfcc6cbdf0e322/server/src/main/java/org/elasticsearch/action/search/RestOpenPointInTimeAction.java#L36).

Unfortunately, only 2 params are in [elasticsearch-specification repo](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_global/open_point_in_time/OpenPointInTimeRequest.ts), I think that is the root cause...

What do you think?

EduardoSaverin:
[...]

PR #2104, opened 2022-11-10
Issue #2084 Fix

EduardoSaverin:
@johtani I have created a fix for your issue.


miriam-eid:
Thank you for opening this. Unfortunately this code is generated from https://github.com/elastic/elasticsearch-specification/blob/main/specification/_global/open_point_in_time/OpenPointInTimeRequest.ts which is the file you need to modify.

You'll have to fix the issue there, and for future contributions you must sign the CLA with the same email address that is used in the commits.

PR #2140, opened 2023-01-24
fix typo in documentation

VladimirFokow:


sethmlarson:
Thanks for opening this @VladimirFokow, unfortunately you made the change against the `8.6` branch rather than `main` so I can't rebase this correctly. Could you reopen with the change against `main`?

PR #2159, opened 2023-02-13
Fix the lint job, build-dists script

sethmlarson:
- New major black release means that there's slightly different formatting.
- Also fixes mypy failures during release builds by allowing mypy to install dependencies to fix types if needed.


ezimuel (approved):
LGTM!

PR #2209, opened 2023-04-30
helpers.async_bulk to return explicit tuple.

bshakur8:
in Python3.11, when `stats_only` is False, it returns a tuple of `(success, errors)` instead of only list of `errors`.
The fix makes sure that it returns a tuple when `stats_only` is True and list otherwise.

Legacy:
```
elasticsearch/_async/helpers.py:348
return success, failed if stats_only else errors
```
will return, `<success>, errors` in case `stats_only` is False./

### DIFF
--- a/elasticsearch/_async/helpers.py
+++ b/elasticsearch/_async/helpers.py
@@ -345,7 +345,7 @@ async def async_bulk(
         else:
             success += 1
 
-    return success, failed if stats_only else errors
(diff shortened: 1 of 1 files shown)


pquentin (commented):
This makes no difference:

```
>>> ("1", "2") if True else "3"
('1', '2')
>>> "1", "2" if True else "3"
('1', '2')
```

PR #2210, opened 2023-04-30
Make async_bulk return also success results.

bshakur8:
When `stats_only` is False, return a tuple of `(results, errors)`, where `results` is the list of bulk results and `errors` is the current errors list.


pquentin:
Sorry for not answering earlier, but this is a breaking change. If you really need this, you can always use the `bulk` API directly, without the helper.

Issue #2142, opened 2023-01-25
Add the Python 3.11 trove classifier to the package


shay314:
Can you please add support in the latest python3.11 version that was released 3 months ago?
Python3.11 is much faster and we want to upgrade, but currently elasticsearch don't support it.

Thanks,

sethmlarson:
Python 3.11 is supported and we [test against it in CI](https://github.com/elastic/elasticsearch-py/blob/main/.github/workflows/ci.yml#L50). I'm guessing by this you mean to add the 3.11 trove classifier?

shay314:
Hi @sethmlarson ,

I guess I should have deeper look.
According to the pypi page the package only support up to python3.10.

Maybe it's worth fixing?
![image](https://user-images.githubusercontent.com/45817758/215308229-a65be521-3bf4-43e8-9e85-a9975cf8e6cf.png)

Thanks,

PR #2211, opened 2023-05-01
List Python v3.11 compatible per #2142

stefnestor:
👋🏼 howdy, team! I believe appending to [setup.py#L78-L92](https://github.com/elastic/elasticsearch-py/blob/master/setup.py#L78-L92) will resolve https://github.com/elastic/elasticsearch-py/issues/2142 to show v3.11 Python compatible on the [PyPI sidebar](https://pypi.org/project/elasticsearch/).


sethmlarson (approved):
Thanks!

PR #2218, opened 2023-05-11
Adapt to v6.8

shencar:


sethmlarson:
The `main` branch of the codebase always targets the latest version of Elasticsearch which is 8.8.0 right now. If you need to use the client against an 6.x Elasticsearch cluster you should install like so:

`$ python -m pip install elasticsearch>=6,<7`

Going to close this PR since we can't accept the changes.

PR #2243, opened 2023-07-04
[8.9][DOCS] Adds getting started content based on the template (#2239)

szabosteve:
## Overview

This PR backports the following changes to the 8.9 branch: 
https://github.com/elastic/elasticsearch-py/pull/2239

PR #2244, opened 2023-07-05
[8.9] Fix mypy error

pquentin:
I noticed this while looking at https://github.com/elastic/elasticsearch-py/pull/2243.

Are you open to pin the versions of black, mypy and other linters?

 * black is now stable but it releases a breaking version every year.
 * mypy reached 1.0 but continues to improve rapidly which means `type: ignore` statements turn into error with each new release.


JoshMock (approved):
not sure why I never approved this :smile:

PR #2254, opened 2023-07-11
[CI] Updates Buildkite configuration

picandocodigo:
Updates configuration to run integration tests on Buildkite.


JoshMock (dismissed):
looks great! I had it on my "as I have time" list to start moving Python over to Buildkite, but you beat me to it.

.buildkite/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# Default UID/GID to 1000
+# it can be overridden at build time
+ARG BUILDER_UID=1000
+ARG BUILDER_GID=1000
+ENV BUILDER_USER elastic
+ENV BUILDER_GROUP elastic
+ENV PATH="${PATH}:/var/lib/elastic/.local/bin"
+
+# Create user
+RUN groupadd --system -g ${BUILDER_GID} ${BUILDER_GROUP} \
```
JoshMock, line 13: I think we established that we don't need to do group/user ID stuff in these containers any more, as that was a requirement for DRA, and Python isn't part of DRA any more. @picandocodigo correct me if I'm misremembering.

JoshMock (approved):
still looks great to me.

JoshMock:
I just double-checked the output of job runs on Buildkite and they're getting pretty similar pass/fail rate as on Jenkins for the integration test suite. Removing unused stuff from `.ci/` is probably the only step left before this is ready to merge.

pquentin (approved):
Thanks! LGTM. I'm considering merging and and address @JoshMock great comments in a future pull request.

.buildkite/Dockerfile
picandocodigo: That's right :+1:

JoshMock:
This can also get a catalog-info.yaml file. https://github.com/elastic/catalog-info/pull/563 was merged earlier today.

pquentin:
8.11 was forked after that pull request.

PR #2268, opened 2023-08-11
Drop support for Python 3.6

pquentin:
Python 3.6 is 5-10% of all downloads, and was EOL'd on 2021-12-23, more than two years ago.


Anaethelion (approved):
LGTM!

PR #2275, opened 2023-08-21
Changed project metadata to use group email address

technige:
This PR removes individual names and email addresses and replaces those with the new (external) [group email address](mailto:client-libs@elastic.co).

The _author_ and _maintainer_ fields have also been collapsed down into just _author_. According to [setuptools](https://setuptools.pypa.io/en/latest/references/keywords.html), the _maintainer_ is used as an override for _author_, so the former is removed to use only the primary field.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,10 +62,8 @@ setup(
     long_description=long_description,
     long_description_content_type="text/x-rst",
     version=package_version,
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! LGTM.

JoshMock (approved):

PR #2279, opened 2023-08-25
Pin untrusted Github Action to a commit hash

JoshMock:
Infosec recommends third-party actions that are maintained by individuals and not on behalf of trusted orgs like GitHub should be pinned to a specific commit.


pquentin (approved):
Thanks! LGTM.

In the future we could ask dependabot to update when the major version change or a vulnerability is found. It even understand commit references like this.

picandocodigo (approved):

PR #2282, opened 2023-08-29
Document type hints individually

pquentin:
The bulk() method used to be documented with an unreadable list of types in the signature:

image

Now types are listed for each parameter specifically:

image

Which is much easier to read.

### DIFF
--- a/docs/sphinx/conf.py
+++ b/docs/sphinx/conf.py
@@ -25,6 +25,8 @@ extensions = ["sphinx.ext.autodoc", "sphinx.ext.doctest", "sphinx.ext.intersphin
 
 autoclass_content = "both"
 
+autodoc_typehints = "description"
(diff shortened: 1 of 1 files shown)


szabosteve (approved):
Great improvement, LGTM!

joshdevins:
Hey @pquentin This is a good improvement! Can we elide the arguments in the method itself? If you look at the Elasticsearch client constructor, it's still a really long list of arguments. Same with `search`.

JoshMock (approved):
Very nice. :clap:

pquentin:
@joshdevins I looked at available Sphinx autodoc parameters but did not find anything that would allow us to do that: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html, sorry.

PR #2289, opened 2023-09-11
Drop run-black.py

pquentin:
run-black.py wraps black and tries to make sure the laptop running black does not overheat, but black is more efficient these days (and I think even uses mypyc) and can go through the already mostly formatted code in the Elasticsearch client much faster than run-black.py allows.


JoshMock (approved):
Good idea. :clap:

PR #2299, opened 2023-09-14
Add Synonyms and Query rules APIs

pquentin:
Closes #2288 

The first commit applies generation and `nox -rs format`, while the second one contains my manual fixes. Tested as follows:

```python3
from elasticsearch import Elasticsearch

# Start with `docker run -p 9200:9200 -it -e xpack.security.enabled=false -e discovery.type=single-node docker.elastic.co/elasticsearch/elasticsearch:8.10.0-SNAPSHOT`
es = Elasticsearch("http://localhost:9200")


es.synonyms.put_synonym(
    id="my-synonyms-set",
    synonyms_set=[
        {"id": "test-1", "synonyms": "hello, hi"},
        {"synonyms": "bye, goodbye"},
        {"id": "test-2", "synonyms": "test => check"},
    ],
)
print(es.synonyms.get_synonyms_sets())

es.query_ruleset.put(
    ruleset_id="my-ruleset",
    query_ruleset={
        "rules": [
            {
                "rule_id": "my-rule1",
                "type": "pinned",
                "criteria": [
                    {
                        "type": "contains",
                        "metadata": "user.query",
                        "values": ["marvel", "dc"],
                    },
                    {"type": "exact", "metadata": "user.country", "values": ["us"]},
                ],
                "actions": {"ids": ["id1", "id2"]},
            }
        ]
    },
)
print(es.query_ruleset.list())
```

Note that `es.query_ruleset.put` is weird, I would have expected to be able to use `es.query_ruleset.put(ruleset_id="my-ruleset", rules=[...])` instead as is idiomatic for the 8.x client. Any thoughts on that @kderusso?

PR #2300, opened 2023-09-14
[8.10] Add Synonyms and Query rules APIs

pquentin:
Backports https://github.com/elastic/elasticsearch-py/pull/2299 with generated code using the 8.10 branch of elasticsearch-specification, which fixes the URLs to the Elasticsearch docs.

### DIFF
--- a/docs/sphinx/api.rst
+++ b/docs/sphinx/api.rst
@@ -144,6 +144,12 @@ Nodes
 .. autoclass:: NodesClient
    :members:
 
+Query rules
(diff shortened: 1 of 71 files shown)


JoshMock (dismissed):
Changes look good to me. Mostly concerned about any changes to the API functions that are made by hand, because those will be overwritten by codegen automation.

pquentin:
The changes made by hand are unfortunate, but they are kept by codegen, thankfully.

PR #2312, opened 2023-09-26
Update generated examples

### LINKED ISSUE #2266: An improper method call in temporary file creation code
This is a security hardening scenario found in the source code. This is found by running a static analysis tool. Not found in a deployed version.

**Elasticsearch version** (`bin/elasticsearch --version`):

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**: 8.8.0

In file: [generate-examples.py](https://generate-examples.py/), there is a method that creates a temporary file using an unsafe API mktemp. The use of this method is discouraged in the Python documentation(https://docs.python.org/3/library/tempfile.html#tempfile.mktemp). I suggested that a temporary file should be created using NamedTemporaryFile(https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) which is a safe API. I replaced the usage of mktemp with NamedTemporaryFile.

Here is a patch for this

```diff
--- utils/[generate-examples.py](https://generate-examples.py/)
+++ utils/[generate-examples.py](https://generate-examples.py/)

@@ -154,7 +154,7 @@
)
)


- tmp_path = Path(tempfile.mktemp())
+ tmp_path = Path(tempfile.NamedTemporaryFile().name)
with tmp_path.open(mode="w") as f:
f.write(t.render(parsed_sources=parsed_sources))

```

Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the Open Source Security Foundation (OpenSSF)(https://openssf.org/): Project Alpha-Omega(https://alpha-omega.dev/). Alpha-Omega is a project partnering with
open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

**Steps to reproduce**:

Not reproduced.

pquentin:
Closes #2266 

@picandocodigo Since Ruby appears to be the most up-to-date client regarding examples, would you mind looking if the generated files look good and tell me if more needs to be done to have up-to-date example in the docs?


picandocodigo (commented):
I think generally it looks good. I checked some of the code changes in the generated examples and they match with Ruby. A thing I'm doing when I generate the code is spin up Elasticsearch and run the generated example code against it. I log the successful runs and the errors to see if there's a client error or a server error (usually because a code example needs some setup). But I add all the successful requests since I'm sure those are fine, and can debug what's wrong with the rest using the logs.

JoshMock (approved):
I didn't look at every one, but spot-checked several and they look good to me.

PR #2324, opened 2023-10-05
Switch from invariant List to covariant Sequence

### LINKED ISSUE #2273: Incorrectly generated signature for `msearch` (`searches` parameter)
**Elasticsearch version** (`bin/elasticsearch --version`):

8.9.0

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**:

8.9.0

The auto generated type annotations for the `searches` parameter (`body`) in `msearch` methods (*and possibly `msearch_template`, but I have never used that API*) under the clients are incorrect and a static type checker will report type errors when a `List` is passed to it.

**Steps to reproduce**:

Since `List`s are invariant (unlike `Tuple`s) the following will trigger a type error;
(*apologies for the slightly "complicated" code but this way the type of the `searches` variable will be correctly resolved thanks to `itertools`*)
```py
import itertools

es = Elasticsearch("http://localhost:9200")

ref_ids_to_search = (
    "9a225de6-53fe-4825-a7bb-089fd57b0238",
    "621b146c-1bdc-4ce3-8a77-d3d637504a80",
)
searches_list = list(
    itertools.chain.from_iterable(
        (
            {},
            {
                "query": {
                    "term": {
                        "ref_id.keyword": ref_id,
                    }
                }
            },
        )
        for ref_id in ref_ids_to_search
    )
)
r = es.msearch(index="my-index", searches=searches_list)  # Argument of type "list[dict[str, dict[str, dict[str, str]]]]" cannot be assigned to parameter "searches" of type "List[Mapping[str, Any]] | Tuple[Mapping[str, Any], ...]" in function "msearch"
```

Changing the signature from this: https://github.com/elastic/elasticsearch-py/blob/b045ce19fb6825a8877203443f2047d221ddd62e/elasticsearch/_async/client/__init__.py#L2568-L2570
to this:
```py
searches: t.Sequence[t.Mapping[str, t.Any]]
```
would (I believe) resolve the issue.

Unsure if this issue would be fixable with changes to [elasticsearch-specification](https://github.com/elastic/elasticsearch-specification) and as the code generator project is private I decided to open an issue here since this feels Python specific.

pquentin:
Closes #2273

To make the snippet from @floxay in #2273 work, I had to explicitly type `searches_list` as `List[Mapping[str, Any]]`, otherwise I was getting:

```
error: Argument "searches" to "msearch" of "Elasticsearch" has incompatible type "list[object]"; expected "Sequence[Mapping[str, Any]]"  [arg-type]
```

### DIFF
--- a/elasticsearch/_async/client/__init__.py
+++ b/elasticsearch/_async/client/__init__.py
@@ -612,14 +612,10 @@ class AsyncElasticsearch(BaseClient):
     async def bulk(
         self,
         *,
-        operations: t.Union[
-            t.List[t.Mapping[str, t.Any]], t.Tuple[t.Mapping[str, t.Any], ...]
-        ],
+        operations: t.Sequence[t.Mapping[str, t.Any]],
         index: t.Optional[str] = None,
         error_trace: t.Optional[bool] = None,
-        filter_path: t.Optional[
-            t.Union[str, t.Union[t.List[str], t.Tuple[str, ...]]]
-        ] = None,
+        filter_path: t.Optional[t.Union[str, t.Sequence[str]]] = None,
         human: t.Optional[bool] = None,
         pipeline: t.Optional[str] = None,
         pretty: t.Optional[bool] = None,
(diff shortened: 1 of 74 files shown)


JoshMock (approved):
I didn't look at every line, but  I get the gist. :smile: LGTM.

PR #2325, opened 2023-10-05
Fix _TYPE_HOSTS to require covariant Sequence

### LINKED ISSUE #2321: Wrong mypy annotation for _TYPE_HOSTS
**Elasticsearch version** (`bin/elasticsearch --version`): 8.9.0 (but it's not relevant to this bug)

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**: 8.10.0

Currently `_TYPE_HOSTS` has type annotation defined as 
```
_TYPE_HOSTS = Union[str, List[Union[str, Mapping[str, Union[str, int]], NodeConfig]]]
```

The problem is that `List` is invariant and thus `List[str]` is not valid subtype for `List[Union[str, Mapping[str, Union[str, int]], NodeConfig]]]`

See common mypy issue https://github.com/python/mypy/issues/3351

The solution is to replace `List` with `Sequence`.

**Steps to reproduce**:
Minimal example
```
import elasticsearch

HOSTS = ['http://localhost:9200', 'http://localhost:10200']  # this will be inferred as list[str]

client = elasticsearch.Elasticsearch(hosts=HOSTS)
```

Mypy output
```
test.py:5: error: Argument "hosts" to "Elasticsearch" has incompatible type "list[str]"; expected "str | list[str | Mapping[str, str | int] | NodeConfig] | None"  [arg-type]
test.py:5: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
test.py:5: note: Consider using "Sequence" instead, which is covariant
```

pquentin:
Closes #2321

### DIFF
--- a/elasticsearch/_sync/client/utils.py
+++ b/elasticsearch/_sync/client/utils.py
@@ -30,6 +30,7 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Sequence,
     Set,
     Tuple,
     Type,
@@ -69,7 +70,9 @@ CLIENT_META_SERVICE = ("es", client_meta_version(__versionstr__))
 # Default User-Agent used by the client
 USER_AGENT = create_user_agent("elasticsearch-py", __versionstr__)
 
-_TYPE_HOSTS = Union[str, List[Union[str, Mapping[str, Union[str, int]], NodeConfig]]]
+_TYPE_HOSTS = Union[
+    str, Sequence[Union[str, Mapping[str, Union[str, int]], NodeConfig]]
+]
 
 _TYPE_ASYNC_SNIFF_CALLBACK = Callable[
     [AsyncTransport, SniffOptions], Awaitable[List[NodeConfig]]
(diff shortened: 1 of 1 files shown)


elasticsearch/_sync/client/utils.py
```diff
@@ -139,7 +142,7 @@ def hosts_to_node_configs(hosts: _TYPE_HOSTS) -> List[NodeConfig]:
     """Transforms the many formats of 'hosts' into NodeConfigs"""
 
     # To make the logic here simpler we reroute everything to be List[X]
-    if not isinstance(hosts, (tuple, list)):
+    if isinstance(hosts, str):
```
pquentin, line 145: This change is required because `hosts` is now a `Sequence` which allows more than just `tuple` and `list`. I considered changing the check to `isinstance(hosts, collections.abc.Sequence)`, but it felt wrong as Python embraces duck typing, which means being a subclass of `collections.abc.Sequence` is not needed to be an actual sequence. (It was much more convincing when it was still in my head! The irony being that we're doing this to fix type hints where all types are very much explicit.)

Anyway, due to the above, I then decided to switch from a denylist to an allowlist, since only two types currently work: `str` and `NodeConfig`. Why disallow `NodeConfig` then? Because the type does not allow it and I'm not expecting anyone to actually be passing `hosts=NodeConfig(...)` as it currently only works by accident. (Note that `hosts=[NodeConfig(...)]` still works.)

JoshMock (approved):

PR #2335, opened 2023-10-10
[DOCS] Adds quickstart guide to ES Py Read The Docs

szabosteve:
## Overview

This PR adds a quick start guide to the Elasticsearch Python Client Read The Docs documentation site.

### DIFF
--- a/docs/sphinx/quickstart.rst
+++ b/docs/sphinx/quickstart.rst
@@ -29,11 +29,8 @@ You can connect to the Elastic Cloud using an API key and the Cloud ID.
 .. code-block:: python
 
     from elasticsearch import Elasticsearch
-    
-    client = Elasticsearch(
-    cloud_id="YOUR_CLOUD_ID",
-    api_key="YOUR_API_KEY"
-    )
+
+    client = Elasticsearch(cloud_id="YOUR_CLOUD_ID", api_key="YOUR_API_KEY")
 
 Your Cloud ID can be found on the **My deployment** page of your deployment 
 under **Cloud ID**.
@@ -77,7 +74,7 @@ This is a simple way of indexing a documentby using the index API:
         document={
             "foo": "foo",
             "bar": "bar",
-        }
+        },
     )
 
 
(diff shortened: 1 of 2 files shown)


docs/sphinx/quickstart.rst
```diff
@@ -0,0 +1,145 @@
...
+------------
+
+- `Python <https://www.python.org/>`_ 3.6 or newer
+- `pip <https://pip.pypa.io/en/stable/>`_
+
+
+Installation
+------------
+
+To install the latest version of the client, run the following command:
```
pquentin, line 17: Not sure I would mention "the latest version of", because if an old version is already installed, it will stay there. The `--upgrade` flag would avoid this, but it seems like an unnecessary complication at this point, hence why I'm suggesting removing "the latest version of".

docs/sphinx/quickstart.rst
```diff
@@ -0,0 +1,145 @@
...
+
+.. code-block:: console
+
+    $ python -m pip install elasticsearch
+
+
+Connecting
+----------
+
+You can connect to the Elastic Cloud using an API key and the Cloud ID.
```
pquentin, line 27:
```suggestion
You can connect to Elastic Cloud using an API key and the Cloud ID.
```

docs/sphinx/quickstart.rst
```diff
@@ -0,0 +1,130 @@
...
+
+Searching documents
+^^^^^^^^^^^^^^^^^^^
+
+This is how you can create a single match query with the Python client: 
+
+
+.. code-block:: python
+
+    client.search(index="my_index", query={"match": {"foo": "foo"}})
```
pquentin, line 97: While your suggestion works, docs suggest this syntax instead: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html.

```suggestion
    client.search(index="my_index", query={"match": {"foo": {"query": "foo"}}})
```

pquentin (dismissed):
Thanks! LGTM with two nits.

pquentin (approved):
Thanks! LGTM.

PR #2347, opened 2023-10-20
Fix type annotation of source parameter.

JannKleen:
This might be completely wrong, but I think the type hinting of the `source` parameter for `Elasticsearch().search()` should be `Sequence[str]` and not `Mapping[str, Any]` (s. https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#source-filtering)
Also `fields` should be accepting `str` as well as `Mapping[str, Any]`


pquentin:
Hello, and thank you! However this code is generated from https://github.com/elastic/elasticsearch-specification/blob/main/specification/_global/search/SearchRequest.ts which is the place to fix if the type hint is wrong. I can check in more details at the end of next week.

pquentin:
Closing in favor of https://github.com/elastic/elasticsearch-specification/pull/2317. Thanks!

PR #2351, opened 2023-10-30
[7.17] Allow urllib3 2.x

pquentin:


pquentin:
Given the trouble I'm having with CI, I'll leave the urllib3 pin for now. At least that should be an incentive to ugprade to elasticsearch-py 8.x.

PR #2367, opened 2023-11-07
Fix CI

pquentin:
This makes CI green, without adding support for Python 3.12 or figuring out why the server tests are not running anymore.


JoshMock (approved):

pquentin (commented):
Forgot to send my comments apparently.

pquentin:
Backports:

 * 8.10: #2369
 * 8.11: #2368

PR #2370, opened 2023-11-08
[8.11] Add Python 3.12 support

pquentin:
I accidentally pushed to main, but would be happy to get a review - I'll fix anything that was wrong here in main.


JoshMock (approved):
LGTM. I think you could have PRed to main and backported to 8.11, right?

pquentin:
> LGTM. I think you could have PRed to main and backported to 8.11, right?

I could! But again I pushed a merge commit to main by accident and reverting would have been difficult.

PR #2382, opened 2023-11-23
Fix mypy

pquentin:


pquentin:
Closing, I did not realize that those fixes were related to code I'm working on.

PR #2383, opened 2023-11-23
Allow unrestricted body again

### LINKED ISSUE #2181: Please bring back `body` parameter to API calls
**Describe the feature**: Currently, when the documentation or implementation is out of sync (see #2179 , #2180 for examples) the API becomes unusable and users have to resort to using the underlying `perform_request` method eliminating some of the benefits of having a dedicated API (parameter checking, URL composition, type hinting, ...). It would be useful to still be able to pass in the entire `body` in those cases.

It would also be great for the common use case of getting an object from Elasticsearch, modifying it, and reapplying it to the cluster, for example index templates, ILM policies, and other configurations. This cannot be now easily done with `**body` since there are some manually introduced changes like `_meta` in the JSON body as opposed to `meta` in the python API. This makes the API extremely brittle for such cases.

pquentin:
Closes #2181 

Allows using the `body` parameter again, not only to allow performance optimizations and use cases where the elasticsearch-specification is not up to date, but also to facilitate migration from 7.x. See #2181 for a detailed discussion. This will be released as part of the next minor release of elasticsearch-py 8.x, likely early next year.

The following rules apply:

 * if the query does not accept body parameters, nothing changes
 * if the body and required body parameter(s) are unset (such as `query` for `search`), the query fails with a `ValueError` about the required parameter(s) (as we can't rely on typing or the default `TypeError` anymore)
 * if only the body is set, then it will be used without validation:
   * as is if already encoded as a string or bytes
   * or serialized to JSON if it's a mapping
 * if both the body and specific parameter(s) are set:
   * if at least one key is the same, the query fails with a `ValueError` exception because we don't know which one to use
   * if the keys are all different, the specific parameters are merged in the body, with a deprecation warning (while this could be an useful way to only use body for parameters missing from elasticsearch-specification, it's confusing and will be removed in a future version)

Note that the above was carefully written to take into account the case where:

 *  a single parameter is used as a body:

    ```python
    # those calls were already working
    es.index(index="test_index", document={"foo": "bar"})
    es.index(index="test_index", document=json.dumps({"foo": "bar"}))
    es.index(index="test_index", document=orjson.dumps({"foo": "bar"}))
    # and those are now possible
    es.index(index="test_index", body={"foo": "bar"})
    es.index(index="test_index", body=json.dumps({"foo": "bar"}))
    es.index(index="test_index", body=orjson.dumps({"foo": "bar"}))
    ```

* multiple parameters are used to reconstruct the body:

    ```python
    # this currently fails as `ignore_missing_component_templates` is not in the specification
    es.indices.put_index_template(
        name="my_template0",
        index_patterns=["my_index*"],
        data_stream={},
        composed_of=["my_data_stream", "my_data_stream_2"],
        ignore_missing_component_templates=["my_data_stream", "my_data_stream_2"],
        priority=1000,
    )
    # this works but the body isn't validated (passing as string and bytes as above works too)
    es.indices.put_index_template(
        name="my_template1",
        body={
            "index_patterns": ["my_index*"],
            "data_stream": {},
            "composed_of": ["my_data_stream", "my_data_stream_2"],
            "ignore_missing_component_templates": ["my_data_stream", "my_data_stream_2"],
            "priority": 1000,
        },
    )
    # using body *and* parameters works but is deprecated:
    es.indices.put_index_template(
        name="my_template2",
        body={"ignore_missing_component_templates": ["my_data_stream", "my_data_stream_2"]},
        index_patterns=["my_index*"],
        data_stream={},
        composed_of=["my_data_stream", "my_data_stream_2"],
        priority=1002,
    )
    ```

The useful commit to review is https://github.com/elastic/elasticsearch-py/pull/2383/commits/7bae67e1b8cf1bf97a4ff7de40fc6e79d80519b0, as the other only updates the generated code.

### DIFF
--- a/docs/guide/migration.asciidoc
+++ b/docs/guide/migration.asciidoc
@@ -167,7 +167,7 @@ with the `security.grant_api_key` API:
 
 [source,python]
 ------------------------------------
-# 8.0+ SUPPORTED USAGES:
+# 8.0+ SUPPORTED USAGE:
 resp = (
     client.options(
         # This is the API key being used for the request
@@ -183,13 +183,36 @@ resp = (
     )
 )
 
-# 7.x DEPRECATED USAGES (Don't do this!):
+# 7.x DEPRECATED USAGE (Don't do this!):
 resp = (
     # This is the API key being used for the request
     client.security.grant_api_key(
         api_key=("request-id", "request-api-key"),
-        # This is the API key being granted
         body={
+            # This is the API key being granted
+            "api_key": {
+                "name": "granted-api-key"
+            },
+            "grant_type": "password",
+            "username": "elastic",
+            "password": "changeme"
+        }
+    )
+)
+------------------------------------
+
+Starting with the 8.12 client, using a body parameter is fully supported again, meaning you can also use `grant_api_key` like this:
+
+[source,python]
+------------------------------------
+# 8.12+ SUPPORTED USAGE:
+resp = (
+    client.options(
+        # This is the API key being used for the request
+        api_key=("request-id", "request-api-key")
+    ).security.grant_api_key(
+        body={
+            # This is the API key being granted
             "api_key": {
                 "name": "granted-api-key"
             },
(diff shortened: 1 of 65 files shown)


JoshMock (approved):
This looks great! I looked at the one commit, 7bae67e1b8cf1bf97a4ff7de40fc6e79d80519b0, and it seemed pretty straightforward to me. :+1:

pquentin:
> I looked at the one commit, [7bae67e](https://github.com/elastic/elasticsearch-py/commit/7bae67e1b8cf1bf97a4ff7de40fc6e79d80519b0), and it seemed pretty straightforward to me. 👍

Really appreciate this comment as I've made a few false starts and worked hard to make this straightforward. :bow:

PR #2392, opened 2023-11-30
[TEST] Add generated docs examples for 8.11.1

leemthompo:
Not sure if this is correct! After a lot flailing I managed to get the generate-examples.py script working locally.

Steps:

1. I ran the https://github.com/elastic/clients-flight-recorder/tree/main/scripts/parsed-alternative-report-recorder script locally
2. Ran generate-examples.py script

PR #2394, opened 2023-12-01
Add Elasticsearch quickstart examples

leemthompo:
We learned in this [test PR](https://github.com/elastic/elasticsearch-py/pull/2392/files) that examples generation workflow isn't working as expected.

This PR:

- adds a subset of examples generated during that test which we know are correct
- the four samples come from this [quickstart](https://www.elastic.co/guide/en/elasticsearch/reference/current/getting-started.html)


pquentin (approved):
Thanks! LGTM.

PR #2415, opened 2024-01-15
Use unquote() directly instead of wrapper

pquentin:
Now that we support Python 3.7+, the wrapper is no longer useful. See https://github.com/elastic/elastic-transport-python/pull/142 for details.


pquentin:
buildkite test this please

Issue #2139, opened 2023-01-24
Add more examples to the documentation


MalfuncEddie:
Currently the examples are very limited and mostly limited to search examples. The documentation should be improved to add examples for:
[elasticsearch.client.IndicesClient.put_mapping](https://elasticsearch-py.readthedocs.io/en/latest/api.html?highlight=mapping#elasticsearch.client.IndicesClient.put_mapping)
https://elasticsearch-py.readthedocs.io/en/latest/api.html?highlight=mapping#elasticsearch.client.IndicesClient.put_mapping
https://elasticsearch-py.readthedocs.io/en/master/api.html?highlight=watcher#elasticsearch.client.WatcherClient

Background: just updated my ES 7 to 8 and the python client scripts I had no longer work.  After several hours plowing through the documentation I just gave up

ezimuel:
@MalfuncEddie we will consider to improve the documentation. Do you have any proposal for a use case? 
Moreover, can you share an error log for the issue that you found moving from ES 7 to 8? Did you check the [Migrating to 8.0](https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/migration.html) documentation page?

MalfuncEddie:
I think even most of de documentation could be automated. The thing is you have a great example at the beginning how to connect to ES and I think that the same examples could be needed elsewhere.


simple examples like:

```
from elasticsearch import Elasticsearch
import json

es =  Elasticsearch(
    "http://localhost:9200",
    api_key=("id", "api_key")
)

if es.ping():
    print('Yay Connect')
else:
    print('Awww it could not connect!')
	
role = "rolename"
roles = ["user"]
rules = {"field" : { "username" : "*" }}'
json_rules = json.loads (rules)
es.security.put_role_mapping(name=role, enabled=enabled,roles=roles, rules=json_rules)
es.close()

```

PR #2417, opened 2024-01-18
Add mappings and bulk to quickstart page

iuliaferoli:
Added a few more examples including dealing with mappings and indexing multiple documents at once (in two alternative ways).
These changes have also been added to the 00 notebook in search-labs, that is also linked to under `interactive examples`.

Response to issue #2139 requesting changes to documentation

### DIFF
--- a/docs/sphinx/quickstart.rst
+++ b/docs/sphinx/quickstart.rst
@@ -41,6 +41,11 @@ You can generate an API key on the **Management** page under Security.
 
 .. image:: ../guide/images/create-api-key.png
 
+Confirm that the connection was successful.
+
+.. code-block:: python
+
+    print(client.info())
 
 Using the client
 ----------------
@@ -49,6 +54,29 @@ Time to use Elasticsearch! This section walks you through the most important
 operations of Elasticsearch. The following examples assume that the Python 
 client was instantiated as above.
 
+Create a mapping for your index
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Set the expected types of your features.
+
+.. code-block:: python
+
+    mappings = {
+        "properties": {
+            "foo": {
+                "type" : "text"
+            },
+            "bar" : {
+                "type" : "text",
+                "fields" : {
+                  "keyword" : {
+                  "type" : "keyword",
+                  "ignore_above" : 256
+                  }
+                 }
+            }
+        }
+    }
 
 Creating an index
 ^^^^^^^^^^^^^^^^^
@@ -57,7 +85,7 @@ This is how you create the `my_index` index:
 
 .. code-block:: python
 
-    client.indices.create(index="my_index")
+    client.indices.create(index="my_index", mappings = mappings)
 
 
 Indexing documents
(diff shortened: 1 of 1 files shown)


docs/sphinx/quickstart.rst
```diff
@@ -76,6 +104,30 @@ This indexes a document with the index API:
...
+.. code-block:: python
+
+    def generate_operations(documents, index_name):
+      operations = []
+      for i, document in enumerate(documents):
+          operations.append({"index": {"_index": index_name, "_id": i}})
+          operations.append(document)
+      return operations
+
+    client.bulk(index=index_name, operations=generate_operations(books, index_name), refresh=True)
```
pquentin, line 118: What are your thoughts on *not* documenting this? It does not scale to a large number of documents and is more verbose. The helper is the way to go. I think I would even mention that `client.bulk` should be avoided.

pquentin (commented):
Thanks!

docs/sphinx/quickstart.rst
iuliaferoli: Added these changes!

docs/sphinx/quickstart.rst
```diff
@@ -76,6 +104,19 @@ This indexes a document with the index API:
...
+
+    from elasticsearch import helpers
+
+    def generate_docs(documents, index_name):
+        for i, document in enumerate(documents):
+            yield dict(_index=index_name, _id=f"{i}", _source=document)
+    
+    helpers.bulk(client, generate_docs(books, index_name))
+
+These helpers are the recommended simple and streamlined way to abstract otherwise complicated and verbose functions such as `client.bulk`.
```
pquentin, line 119: Unlike Markdown, you need two backticks in reStructuredText to format as code: single backticks only add emphasis. What do you think of this wording?

```suggestion
These helpers are the recommended way to perform bulk ingestion. While it is also possible to perform bulk ingestion using ``client.bulk`` directly, the helpers handle retries, ingesting chunk by chunk and more. See the :ref:`helpers` page for more details.
```

I've considered linking to the `client.bulk()` docs using `:meth:`~elasticsearch.Elasticsearch.bulk` but I did not find it clearer as it's only rendered as "bulk()" with a link to the docs.

pquentin (commented):
This looks great, thanks! Only a few more details to iron out.

Did you know that you could see the result of your changes here? https://elasticsearch-py--2417.org.readthedocs.build/en/2417/quickstart.html

docs/sphinx/quickstart.rst
```diff
@@ -57,7 +85,7 @@ This is how you create the `my_index` index:
 
 .. code-block:: python
 
-    client.indices.create(index="my_index")
+    client.indices.create(index="my_index", mappings = mappings)
```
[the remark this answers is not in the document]
pquentin: This is a quickstart, I think there should be only one way to create an index, what do you think?

I would then remove "Creating an index" and rename "Create a mapping for your index" to "Create an index with mappings"

docs/sphinx/quickstart.rst
```diff
@@ -59,6 +63,29 @@ This is how you create the `my_index` index:
...
+                    "keyword": {
+                        "type": "keyword",
+                        "ignore_above": 256,
+                    }
+                },
+            },
+        }
+    }
+
+    client.indices.create(index="my_index", mappings = mappings)
```
pquentin, line 88:
```suggestion
    client.indices.create(index="my_index", mappings=mappings)
```

pquentin (commented):
Thanks! Only two things left.

pquentin (approved):
Thanks for iterating! LGTM.

PR #2419, opened 2024-01-18
Update connecting.asciidoc to add aki_key recommendation

iuliaferoli:
Updated by emphasising recommendation to use api_key for auth rather than username/password.
Did not change all instances of basic auth with api_key because I have not personally tested all of them.
Highlight the use of API_key for the cloud


docs/guide/connecting.asciidoc
```diff
@@ -15,22 +15,24 @@ page after you've created a cluster (look in the top-left if you're in Kibana).
 
 We recommend using a Cloud ID whenever possible because your client will be 
 automatically configured for optimal use with Elastic Cloud including HTTPS and 
-HTTP compression.
+HTTP compression; as well as the API key rather than
+the alternative `basic_auth=(user, password)` option.
```
pquentin, line 19: I believe this is orthogonal, you can use a Cloud ID with basic auth or API keys. (Cloud ID is not a secret, it's only an URL.)

docs/guide/connecting.asciidoc
```diff
@@ -15,22 +15,24 @@ page after you've created a cluster (look in the top-left if you're in Kibana).
...
-
 # Found in the 'Manage Deployment' page
 CLOUD_ID = "deployment-name:dXMtZWFzdDQuZ2Nw..."
+# Generated under "Stack Management/Security" page
+API_KEY = "..."
 
 # Create the client instance
 client = Elasticsearch(
+    # For local development
+    # hosts=["http://localhost:9200"]
```
pquentin, line 33: This is slightly confusing. Maybe it should be another section? Indeed, locally, using API Keys is more difficult - using the provided is much simpler. (Also, by default the URL is `https://` locally too).

docs/guide/connecting.asciidoc
```diff
@@ -336,6 +338,20 @@ es = Elasticsearch(
...
+----
+# Create the client instance
+client = Elasticsearch(
+    # For local development
+    # hosts=["http://localhost:9200"] 
+    cloud_id=CLOUD_ID,
+    api_key=API_KEY,
+)
+----
+
```
pquentin, line 354: I'm not sure I'm following, how is this different from above? That said, an example with basic auth would be useful, since this is what the Elastic Cloud onboarding suggest. Maybe we should add this here?

pquentin (commented):
Thank you for this! Added a few comments.

Issue #2385, opened 2023-11-24
Bug in documentation


ruben1679:
**Describe the feature**:

**Elasticsearch version** (`bin/elasticsearch --version`):

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**: latest

Please make sure the major version matches the Elasticsearch server you are running.

**Description of the problem including expected versus actual behavior**:

Function documentation has a bug in the API reference:
https://elasticsearch-py.readthedocs.io/en/latest/api/elasticsearch.html

In function: https://elasticsearch-py.readthedocs.io/en/latest/api/elasticsearch.html#elasticsearch.Elasticsearch.search
the key `from` is invalid, please remove.

```
from – Starting document offset. Needs to be non-negative. By default, you cannot page through more than 10,000 hits using the from and size parameters. To page through more hits, use the search_after parameter.
```

there is a `from_` at the bottom, so this is just a stale parameter description.

**Steps to reproduce**:
check the documentation

Issue #2167, opened 2023-02-27
SSL Context might ignore certificate `verify_mode`


b-deam:
**Elasticsearch version** (`bin/elasticsearch --version`): 8.x

**`elasticsearch-py` version (`elasticsearch.__versionstr__`)**:  8.6.1

Please make sure the major version matches the Elasticsearch server you are running.

**Description of the problem including expected versus actual behavior**:

According to https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/config.html#_using_an_sslcontext
we should be able to omit the `verify_certs` parameter in the client constructor if we correctly setup an ssl context object.

**Steps to reproduce**:

1. Start a local Elasticsearch cluster with SSL enabled and a self-signed certificate
2. Execute this reproduction script
  ```python
    from elasticsearch import Elasticsearch
    import ssl
    import certifi
    
    ssl_context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH, cafile=certifi.where())
    ssl_context.check_hostname = False
    ssl_context.verify_mode = ssl.CERT_NONE
    
    # works, setting verify_certs means we don't verify the certificate 
    es = Elasticsearch(
      hosts=["https://localhost:9200"],
      ssl_context=ssl_context,
      verify_certs=False,
      basic_auth=("elastic", "changeme"),
    )
    print(es.info())
    
    # fails, even though the SSL context object's verify_mode is set to NONE it still attempts to verify the certificate 
    es = Elasticsearch(
      hosts=["https://localhost:9200"],
      ssl_context=ssl_context,
      basic_auth=("elastic", "changeme"),
    )
    print(es.info())

4lissonsilveira:
Hi @b-deam what could be a potential fix? Perhaps defaulting `verify_certs` to False if `ssl_context` is specified by the user and `verify_certs` is not?

e.g.: adding the below block of code to https://github.com/elastic/elasticsearch-py/blob/650ad2a338d957fa4308bed01a9797fe44714d7f/elasticsearch/_sync/client/utils.py#L112

```
    # if there is a ssl context, verify_certy should default to False
    if (
            node_options.get("ssl_context") and
            node_options.get("verify_certs") is None
    ):
        node_options["verify_certs"] = False

```

many thanks!

PR #2421, opened 2024-01-18
Removing the from parameter per issue #2385

JessicaGarson:
I removed the from parameter to fix the bug in the documentation reported [here](https://github.com/elastic/elasticsearch-py/issues/2385). 

I also wanted to know if you knew of a good spot to add back to the description.


pquentin:
Hello! I really appreciate your efforts here, but unfortunately that code is generated, so that change will disappear at the next edit. I need to fix the generator, which is not public.

If you're still interested in working on the Elasticsearch client, maybe https://github.com/elastic/elasticsearch-py/issues/2167 is a good fit? It involves a few changes in https://github.com/elastic/elastic-transport-python/. I can provide more explanations if needed. Thank you.

PR #2430, opened 2024-02-07
Switch to 2024 black style

pquentin:


noxfile.py
```diff
@@ -53,7 +53,7 @@ def test(session):
 
 @nox.session()
 def format(session):
-    session.install("black", "isort", "flynt", "unasync")
+    session.install("black", "isort", "flynt", "unasync", "setuptools")
```
pquentin, line 56: setuptools is no longer installed by default in recent versions of Python, but unasync requires it.

pquentin:
buildkite test this please

PR #2345, opened 2023-10-13
Remove .ci directory

pquentin:

PR #2437, opened 2024-02-13
Restore .ci/make.sh file

pquentin:
This file was removed by mistake in https://github.com/elastic/elasticsearch-py/pull/2345.


picandocodigo (commented):
We talked about moving this file to `.github` after migrating to Buildkite, since we wouldn't need the `.ci` folder anymore. The automations already check for `make.sh` in `.github`, so it should work like before.

pquentin:
Great point, thanks! I had forgot about this. Done, please take another look.

picandocodigo (commented):
Occurrences of `.ci` need to be replaced by `.github` (or `.buildkite` in the case of Dockerfile in line 117) in the script. Other than that LGTM :+1:

pquentin:
Another good catch! Tested with `.github/make.sh assemble 8.12.0` this time (since codegen requires a token). Please take another look.

picandocodigo (approved):

Issue #2435, opened 2024-02-12
Add OpenTelemetry support


pquentin:
There are currently three ways to instrument an application that uses the Elasticsearch Python client:

 * Using the Elastic [APM Python agent](https://www.elastic.co/guide/en/apm/agent/python/current/supported-technologies.html#automatic-instrumentation-db-elasticsearch)
 * Using the [OpenTelemetry contrib elasticsearch-py instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-elasticsearch)
 * For completeness, I suppose using the [Elastic Intake API](https://www.elastic.co/guide/en/observability/current/api-events.html) manually is an option.

The main benefit of the existing approaches is that while they can handle past versions of this client, they're fragile to changes in elasticsearch-py, are difficult to test and may be suboptimal as the code may need refactorings to emit spans with complete information. In the case of the OpenTelemetry instrumentations, they don't have clear owners. For this reason, it makes sense to implement this in elasticsearch-py where the feature will be tested and maintained.

---

We will follow the existing [Semantic conventions for Elasticsearch](https://opentelemetry.io/docs/specs/semconv/database/elasticsearch/) and emit spans for all requests when the [opentelemetry-api](https://pypi.org/project/opentelemetry-api/) optional package will be detected. Correctly configuring OpenTelemetry SDK will be left to the user.

The elasticsearch-ruby OpenTelemetry instrumentation will be used as inspiration:

 * https://github.com/elastic/elasticsearch-ruby/pull/2179 makes sure to send enough metadata to the transport
 * https://github.com/elastic/elastic-transport-ruby/pull/54 performs the actual instrumentation

## Configuration

* `OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_ENABLED` (default: true) **Enable / Disable the OpenTelemetry instrumentation**. With this configuration option you can enable (default) or disable the built-in OpenTelemetry instrumentation.
[...]

PR #2457, opened 2024-02-22
Pass endpoint_id and path_parts to transport

pquentin:
As part of https://github.com/elastic/elasticsearch-py/issues/2435, the transport needs to know the endpoint id and path parts, and can only get them from the client. This pull request makes sure to pass that info to the client.

There are two commits here:

* [`5ff3169` (#2457)](https://github.com/elastic/elasticsearch-py/pull/2457/commits/5ff3169a3e14422b2d830d917ee132a64448a71c) which accepts endpoint_id/path_parts in base clients, updates tests and depends on the main branch of elastic-transport
* [`2ce3c32` (#2457)](https://github.com/elastic/elasticsearch-py/pull/2457/commits/2ce3c3205f0a1e016d760873f1218bb5e74c4a2d) that runs code generation to define endpoint_id/path_parts in all APIs

This PR depends on https://github.com/elastic/elastic-transport-python/pull/151.

I'm wondering about the type of those new parameters. The natural way in elastic-transport was `str | DefaultType` while here it's `str | None`. `DEFAULT ` is a way to make `None` meaningful, but that's not needed here.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
-elastic-transport>=8.0.0b1, <9
+# TODO switch back to elastic-transport>=8,<9 between elastic-transport release and elasticsearch-py release
+elastic-transport @ git+https://github.com/elastic/elastic-transport-python
 requests>=2, <3
 aiohttp
 pytest
--- a/elasticsearch/_async/client/__init__.py
+++ b/elasticsearch/_async/client/__init__.py
@@ -674,9 +674,12 @@ class AsyncElasticsearch(BaseClient):
             )
         elif operations is not None and body is not None:
             raise ValueError("Cannot set both 'operations' and 'body'")
+        __path_parts: t.Dict[str, str]
         if index not in SKIP_IN_PATH:
-            __path = f"/{_quote(index)}/_bulk"
+            __path_parts = {"index": _quote(index)}
+            __path = f'/{__path_parts["index"]}/_bulk'
         else:
+            __path_parts = {}
             __path = "/_bulk"
         __query: t.Dict[str, t.Any] = {}
         if error_trace is not None:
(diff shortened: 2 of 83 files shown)


pquentin:
This approach is incorrect for a few reasons:

 * the typing of endpoint_id and path_parts has to be fixed to match the definition in elastic-transport-python. It's not `Optional`, but `Union[DefaultType, ...]`. Maybe `Optional` was fine, we use both styles in the code, not sure why.
 * the values should be quoted (so that eg. ["a", "b"] becomes "a,b")
 * we should only pass values that are set, indeed many APIs have multiple forms, like `_cat/recovery` and `_cat/recovery/{index}`

The last two can be solved by defining `__path_parts` alongside `__path`, like this:

```python
if index not in SKIP_IN_PATH:
    __path_parts = {"index": _quote(index)}
    __path = f"/_cat/recovery/{__path_parts['index']}"
else:
    __path_parts = {}
    __path = "/_cat/recovery"
```

dev-requirements.txt
```diff
@@ -1,4 +1,5 @@
-elastic-transport>=8.0.0b1, <9
+# TODO switch back to elastic-transport>=8,<9 between elastic-transport release and elasticsearch-py release
+elastic-transport @ git+https://github.com/elastic/elastic-transport-python
```
xrmx, line 2: I vaguely remember some pain in specifying git urls without an hash, like when trying to upgrade a package checked out from git.
BTW you didn't release a elastic-transport-python because you want to check that all layers are working right?

xrmx (approved):

dev-requirements.txt
pquentin: That's right. I don't intend to release anything with git dependencies. Maybe I should use a pre-release instead though!

pquentin:
8.13 backport is https://github.com/elastic/elasticsearch-py/pull/2468. This should not be backported to 8.12, I nearly forgot.

PR #2470, opened 2024-03-12
Make test_otel more robust to elastic-transport changes

pquentin:
Depends on https://github.com/elastic/elastic-transport-python/pull/153


pquentin:
This should *not* be passing.

pquentin:
Merging to fix the main and 8.13 branches.

xrmx (commented):
Completely missed this, LGTM. I think is is the first time I've seen items() comparison that is not equality  :sweat_smile:

PR #2477, opened 2024-03-20
Changed API Key from tuple to encoding in the docs

iuliaferoli:
~~still to be added - the mention in the readthedocs index~~
no longer applicable with the redirect from master to latest

### DIFF
--- a/docs/guide/connecting.asciidoc
+++ b/docs/guide/connecting.asciidoc
@@ -265,7 +265,7 @@ es.options(
 # You can persist the authenticated client to use
 # later or use for multiple API calls:
 auth_client = es.options(
-    api_key=("api-key-id", "api-key-secret")
+    api_key="api_key"
 )
 for i in range(10):
     auth_client.index(
@@ -321,8 +321,8 @@ es = Elasticsearch(
 ==== API Key authentication
 
 You can configure the client to use {es}'s API Key for connecting to your 
-cluster. Note that you need the values of `id` and `api_key` to
-https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html[authenticate via an API Key].
+cluster. Note that you need the API Key credential representing the Base64-encoding of the id and api_key value 
+https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html[generated from the API Key].
 
 [source,python]
 ----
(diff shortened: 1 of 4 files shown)


docs/guide/getting-started.asciidoc
```diff
@@ -36,7 +36,7 @@ from elasticsearch import Elasticsearch
 
 client = Elasticsearch(
     "https://...",  # Elasticsearch endpoint
-    api_key=('api-key-id', 'api-key-secret'),  # API key ID and secret
+    api_key='api_key',  # API key encoding
```
pquentin, line 39: I think the comment is confusing here, I would remove it.

docs/guide/connecting.asciidoc
```diff
@@ -321,8 +321,8 @@ es = Elasticsearch(
 ==== API Key authentication
 
 You can configure the client to use {es}'s API Key for connecting to your 
-cluster. Note that you need the values of `id` and `api_key` to
-https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html[authenticate via an API Key].
+cluster. Note that you need the API Key credential representing the Base64-encoding of the id and api_key value 
+https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html[generated from the API Key].
```
pquentin, line 325: This looks good and I don't want to delay this pull request further, but for future work I think we should encourage users going to Kibana instead which has a clear UI and docs (https://www.elastic.co/guide/en/kibana/current/api-keys.html) and displays the correct encoding by default:

image

pquentin (approved):
Great, thanks!

iuliaferoli:
Also added the same change in the __init__ so it shows up on https://elasticsearch-py.readthedocs.io/en/latest/api/elasticsearch.html

docs/guide/connecting.asciidoc
pquentin: Here's the result: https://elasticsearch-py_bk_2477.docs-preview.app.elstc.co/guide/en/elasticsearch/client/python-api/master/connecting.html#auth-apikey.

I don't understand the sentence:

> You can configure the client to use Elasticsearch’s API Key for connecting to your cluster. Note that you need the API Key encoding you can [generated in Kibana](https://elasticsearch-py_bk_2477.docs-preview.app.elstc.co/guide/en/kibana/current/api-keys.html#create-api-key).

We also need to keep the Elasticsearch link for users that don't have Kibana. This all could have been done in a future pull request

iuliaferoli: Changed to include both methods

pquentin (approved):
Thanks! LGTM.

PR #2483, opened 2024-03-22
Make helpers examples runnable

iuliaferoli:
adding examples of how to import the helpers;
adding the client definition since ```bulk``` and ```es``` were used
without context / definition
changed from es to client to stay consistent with the quickstart naming convention


docs/sphinx/helpers.rst
```diff
@@ -5,6 +5,15 @@ Helpers
 
 Collection of simple helper functions that abstract some specifics of the raw API.
 
+Connecting
+----------
+
+.. code-block:: python
+
+    from elasticsearch import Elasticsearch, helpers
```
pquentin, line 13: Any reason to import helpers here? Do we use it anywhere?

pquentin (commented):
Great idea, thanks!

docs/sphinx/helpers.rst
iuliaferoli: We don't; I just wanted to have at least one example of importing the rest of them, other than bulk?
But we can also skip this one

pquentin: Yes it seems weird if we don't use. Either we call `helpers.bulk` or we drop this IMO.

iuliaferoli: fixed

docs/sphinx/helpers.rst
```diff
@@ -5,6 +5,15 @@ Helpers
...
 Collection of simple helper functions that abstract some specifics of the raw API.
 
+Connecting
+----------
+
+.. code-block:: python
+
+    from elasticsearch import Elasticsearch
+    
+    client = Elasticsearch(cloud_id="YOUR_CLOUD_ID", api_key="YOUR_API_KEY")
```
pquentin, line 15:
```suggestion
    client = Elasticsearch("https://.../", api_key="YOUR_API_KEY")
```

pquentin (approved):

PR #2486, opened 2024-03-22
Rename es to client in examples for consistency

iuliaferoli:
Changed examples / docs to only refer to the client instance as `client` rather than `es` so we now have uniform naming conventions for clarity.


docs/sphinx/index.rst
```diff
@@ -65,22 +65,22 @@ Example Usage
     from datetime import datetime
     from elasticsearch import Elasticsearch
 
-    es = Elasticsearch("http://localhost:9200")
+    client = Elasticsearch(cloud_id="YOUR_CLOUD_ID", api_key="YOUR_API_KEY")
```
pquentin, line 68: I don't want to encourage Cloud IDs which don't really have an advantage over simple endpoints:

```suggestion
    client = Elasticsearch("http://localhost:9200/", api_key="YOUR_API_KEY")
```

pquentin (commented):
Great work, thanks!

`git grep "\bes\b" | grep -v '{es}' | grep -v '^test_elasticsearch'` gave three more places to update:

* https://github.com/elastic/elasticsearch-py/blob/66798fd5ea723026ee99db6fcbe98272c994965e/elasticsearch/helpers/actions.py#L659-L663
* https://github.com/elastic/elasticsearch-py/blob/66798fd5ea723026ee99db6fcbe98272c994965e/elasticsearch/_async/helpers.py#L395-L402
* https://github.com/elastic/elasticsearch-py/blob/66798fd5ea723026ee99db6fcbe98272c994965e/docs/guide/connecting.asciidoc?plain=1#L251-L272

docs/sphinx/index.rst
iuliaferoli: added those three and made the change you suggested, thanks!

pquentin (approved):

Issue #2429, opened 2024-02-03
[DOC] Add more Python Client code examples to main Elasticsearch Docs


stefnestor:
👋 howdy, team!

**Describe the feature**:

@scranberryjam & I would like to add more [Elasticsearch Client](https://www.elastic.co/guide/en/elasticsearch/client/index.html) content to the [Elasticsearch API docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/rest-apis.html). 

<details><summary>example difference</summary>

---

Examples for [Cluster API's](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster.html)
- does have [Cluster Health](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html)
    <img width="438" alt="image" src="https://github.com/elastic/elasticsearch-py/assets/26751266/08654939-0c4f-44cd-9b78-52158a1a7e31">
- does NOT have [Allocation Explain](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-allocation-explain.html)
    <img width="432" alt="image" src="https://github.com/elastic/elasticsearch-py/assets/26751266/e94b7e89-7c1b-478c-a498-0d6fb82a00e0">

---

</details>

(Taking Python as the first language, since I'm most comfortable with it,) It doesn't appear these code examples are in-line in the [Elasticsearch docs code](https://github.com/elastic/elasticsearch/tree/main/docs/reference). Instead it appears Python examples surface in Elasticsearch Docs via [this repo's code for "doc examples"](https://github.com/elastic/elasticsearch-py/tree/main/docs/examples). 

Checking these files' history, it appears these may be auto-generated (last done 2023Sep via [#2312](https://github.com/elastic/elasticsearch-py/pull/2312), [commit before](https://github.com/elastic/elasticsearch-py/commit/e3e109a362f98430e123c5e970275712daf4b58d))? (Is that part of the [Auto-generate API](https://github.com/elastic/elasticsearch-py/blob/main/CONTRIBUTING.md#api-code-generation)?) 

Do these need to be re-generated periodically or is there a project here I could help on to fill in more pages?

Initial examples confirmed not having ES-doc Python code examples: [Elasticsearch doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/index.html) 👉 [ES-Py doc > ClusterClient](https://elasticsearch-py.readthedocs.io/en/v8.12.0/api/cluster.html)
[...]

PR #2488, opened 2024-03-24
Add inference doc code examples

iuliaferoli:
I'm adding inference doc code examples per https://github.com/elastic/elasticsearch-py/issues/2429

### DIFF
--- a/docs/examples/00fea15cbca83be9d5f1a024ff2ec708.asciidoc
+++ b/docs/examples/00fea15cbca83be9d5f1a024ff2ec708.asciidoc
@@ -4,7 +4,7 @@
 ----
 resp = client.inference.put_model(
     task_type="text_embedding",
-    model_id="my-e5-model",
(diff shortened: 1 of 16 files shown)


pquentin:
Thanks Iulia! Note that simply adding the pages is not enough, you also have to run `python utils/generate-examples.py` after having cloned the `client-flights-recorder` repository. And then check that the results make sense. I'm happy to assist more.

iuliaferoli:
Hey Quentin! 
I've generated the examples and fixed a few minor details.
For example, it generated inference_id instead of model_id; so I think this also needs to be changed in the docs here: https://elasticsearch-py.readthedocs.io/en/latest/api/inference.html

Also I've had some trouble with the branches so it looks quite messy - still trying to get rid of having the client changes in here as well instead of in the separate PR.

pquentin (approved):

PR #2508, opened 2024-03-27
Ensure optional dependencies are optional

pquentin:
orjson is an optional dependency, but I was checking its absence with `ModuleNotFoundError`, a child exception of `ImportError`. While `import orjson` will indeed raise `ModuleNotFoundError`, `from .serializer import OrjsonSerializer` raises `ImportError` instead because the module (`.serializer`) was found.

OpenTelemetry is already tested in separate integration tests, but checking it explicitly in lint seemed useful as it's only one more line.


ezimuel:
LGTM!

PR #2528, opened 2024-04-17
VectorStore for GenAI integrations

maxjakob:
We want to add a higher-level abstraction for using Elasticsearch in Python. The main motivation is to make integrations into 3rd party GenAI libraries easier and give users a handful of options to choose from without diving into the Query DSL. Ideally the abstraction also works for direct users of the elasticsearch-py client.

## Integrations
- LangChain: [current](https://github.com/langchain-ai/langchain-elastic/blob/main/libs/elasticsearch/langchain_elasticsearch/vectorstores.py) and [draft branch](https://github.com/langchain-ai/langchain-elastic/blob/orchestration-lib/libs/elasticsearch/langchain_elasticsearch/vectorstores.py) utilizing this PR
- LlamaIndex: [current](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py) and [draft branch](https://github.com/maxjakob/llama_index/blob/es-use-orchestration-lib/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py) utilizing this PR
- Haystack: [current](https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/elasticsearch/src/haystack_integrations)

## TODO
- [x] Agree on shape of the interfaces <-- nobody complained ;)
- [x] Move client creation code to 3rd parties, receive client object
- [x] Separate classes for sync and async
- [x] Move embedding service calls to the store, batch the inference
- [x] Move module to `helpers.vectorstore`
- [x] Deduplicate field names between store and strategies (as in the original implementation)
- [x] Deduplicate application of metadata mapping
- [x] Semantic strategy: ~test it once there is a Docker image with it or~ remove it
- [x] Make integration tests work in CI
- [x] ~Consider limiting values for `k` and `num_candidates` for (async) MMR~ --> CPU-intensive MMR function is now sync only
- [x] Create extra requirements for MMR-related dependencies
- [x] Adhere to repo-specific code styles
- [x] Inquire what needs to be done for the `-serverless` package --> nothing here, @pquentin will take care of this.
- [x] Add default arg for user agent
- [x] Documentation? <-- We can worry about it later when the integrations are stable.


elasticsearch/store/_utilities.py
```diff
@@ -0,0 +1,153 @@
...
+        if client_params is not None:
+            connection_params.update(client_params)
+
+        client = AsyncElasticsearch(**connection_params)
+
+    if not isinstance(client, AsyncElasticsearch):
+        raise TypeError("Elasticsearch client must be AsyncElasticsearch client")
+
+    # Add integration-specific usage header for tracking usage in Elastic Cloud.
+    # client.options preserces existing (non-user-agent) headers.
```
pquentin, line 53:
```suggestion
    # client.options preserves existing (non-user-agent) headers.
```

pquentin (commented):
Thanks for this draft! I've reviewed with an eye towards the big things that could be improved to make this feel native in the Python client. I've not looked at the detail of the code, since I don't expect to bring value here.

The big question is going to be the name of this module and its location in the client.

 * I don't think `store` works here: we have to communicate that this is an LLM/GenAI thing. I would thus call it `vectorstore` or `vector_store`. (Indeed, according to PEP 8, [underscores can be used in the module name if it improves readability](https://peps.python.org/pep-0008/#package-and-module-names).)
 * Currently, all the manually written code is in [`elasticsearch.helpers`](https://elasticsearch-py.readthedocs.io/en/v8.13.0/helpers.html) which felt like the natural location for this new code initially. But I'm less sure now, given existing helpers are simply functions and this is a lot more code with a lot more state. So I'm tempted to think that `elasticsearch.vector_store` is a better location than `elasticsearch.helpers.vector_store`.

elasticsearch/store/_utilities.py
```diff
@@ -0,0 +1,153 @@
+from typing import Any, Dict, List, Optional, Union
+
+import numpy as np
```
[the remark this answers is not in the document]
maxjakob: I guess we _could_ implement MRR without numpy, but it's not going to be pretty and it's going to be slower.

I like having an optional dependency in general. Makes it clear that this is an add-on.

elasticsearch/store/_utilities.py
```diff
@@ -0,0 +1,153 @@
...
+
+def create_elasticsearch_client(
+    agent_header: str,
+    client: Optional[AsyncElasticsearch] = None,
+    url: Optional[str] = None,
+    cloud_id: Optional[str] = None,
+    api_key: Optional[str] = None,
+    username: Optional[str] = None,
+    password: Optional[str] = None,
+    client_params: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
maxjakob: I moved this function to the orchestration to not have to repeat the silly ifs. We can totally change the names if we keep this function here. I'm also not opposed to duplicating this if you feel it's not a good native fit for the client lib. 

(Currently users can specify their details as arguments to the the 3rd party libs vector stores. They already work with some of these names so on that end we are bound to them.)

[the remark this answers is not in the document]
maxjakob: `agent_header` is new and not used anywhere yet, will rename to `user_agent`.

maxjakob:
>  * I don't think `store` works here: we have to communicate that this is an LLM/GenAI thing. I would thus call it `vectorstore` or `vector_store`. (Indeed, according to PEP 8, [underscores can be used in the module name if it improves readability](https://peps.python.org/pep-0008/#package-and-module-names).)

It was called vectorstore for the longest time :) I renamed it because of the BM25 capabilities. But you're right, it's more appealing/familiar to people with vector in the name.

test_elasticsearch/test_store_integration/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
+version: "3"
```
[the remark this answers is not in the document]
pquentin: TIL about services! They're handy but don't work locally. Our current solution is https://github.com/elastic/elasticsearch-py/blob/main/.buildkite/run-elasticsearch.sh. Tests that do run against Elasticsearch should be in `test_elasticsearch/test_server` and can accept `sync_client` or `elasticsearch_url` pytest fixtures. They will run in Buildkite, for historical reasons. You can define `ELASTICSEARCH_URL` to run them locally against the instance of your choice.

Test that don't need Elasticsearch (such as unit tests) should be in `test_elasticsearch` and will run in GitHub Actions.

elasticsearch/store/store.py
```diff
@@ -0,0 +1,399 @@
...
+        if exists.meta.status == 200:
+            logger.debug(f"Index {self.index_name} already exists. Skipping creation.")
+        else:
+            await self.retrieval_strategy.create_index(
+                client=self.es_client,
+                index_name=self.index_name,
+                metadata_mapping=self.metadata_mapping,
+            )
+
+    def max_marginal_relevance_search(
```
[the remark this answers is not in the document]
pquentin: OK, I guess I was wondering if this was meant to run async at all since you're only calling `embed_query` and not `embed_query_async` here. If we do want to run this async, we may have to use tricks depending on how long this typically takes.

elasticsearch/store/_utilities.py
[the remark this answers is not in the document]
pquentin: > I moved this function to the orchestration to not have to repeat the silly ifs. We can totally change the names if we keep this function here. I'm also not opposed to duplicating this if you feel it's not a good native fit for the client lib.

Yeah I'm not sure to be honest. The native way to do that would be to just accept a client instance. I'm wondering if we should stop calling `create_elasticsearch_client` from `ElasticsearchEmbeddings.__init__` and `ElasticsearchStore.__init__` but accept a client and just call `.options` there. Then in the third-party libraries we can call `create_elastiscearch_client` independently. This is more friction though.

> (Currently users can specify their details as arguments to the the 3rd party libs vector stores. They already work with some of these names so on that end we are bound to them.)

We're not bound to them here just yet, though: the translation from say `username`/`password` to `basic_auth` could happen in the 3rd party libs vector stores.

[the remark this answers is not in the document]
ezimuel: I agree with @pquentin to design the code to be more client native. I think it's better to pass just a client instance, instead of all the possible parameters for building a client. In the 3rd party integrations we can then decide to pass the parameter and build a client object to be passed to the client helper function.

ezimuel:
>  * I don't think `store` works here: we have to communicate that this is an LLM/GenAI thing. I would thus call it `vectorstore` or `vector_store`. (Indeed, according to PEP 8, [underscores can be used in the module name if it improves readability](https://peps.python.org/pep-0008/#package-and-module-names).)

I think `vectorstore` (or `vector_store`) is a better name, since it puts evidence on the semantic search feature.

>  * Currently, all the manually written code is in [`elasticsearch.helpers`](https://elasticsearch-py.readthedocs.io/en/v8.13.0/helpers.html) which felt like the natural location for this new code initially. But I'm less sure now, given existing helpers are simply functions and this is a lot more code with a lot more state. So I'm tempted to think that `elasticsearch.vector_store` is a better location than `elasticsearch.helpers.vector_store`.

Since this code consumes the `client` as an additional layer, I think putting it into `elasticsearch.helpers` namespace is the perfect fit. The `helpers` should collects all the high level features that facilitate the usage of the Elasticsearch endpoints.

elasticsearch/store/store.py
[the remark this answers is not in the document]
maxjakob: The computationally expensive part in Python can be limited by limiting `k` and `num_candidates`. We could think of sensible maximum values to enforce here.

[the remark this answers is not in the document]
pquentin: For reference, slow in the context of asyncio depends on the application but a good baseline is 100ms: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.slow_callback_duration. If it usually takes more time than that, we can think about mitigation measures.

elasticsearch/store/embedding_service.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Args:
+            text: The query text to generate an embedding for.
+
+        Returns:
+            The embedding for the input query text.
+        """
+
+
+class ElasticsearchEmbeddings(EmbeddingService):
```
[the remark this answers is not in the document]
maxjakob: I like this, really declutters the code I work with.

One question: are test cases also handled, including `@pytest.mark.asyncio` markers?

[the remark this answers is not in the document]
pquentin: Great question! It's not handled, because `unasync` only handles single tokens. See https://github.com/elastic/elasticsearch-dsl-py/pull/1714#discussion_r1527918832 for the same discussion in the DSL client. For now we can enable pytest-asyncio's auto mode.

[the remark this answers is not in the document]
pquentin: I discussed quickly with @miguelgrinberg and actually it's best to handle this with sed, as we do to remove `import asyncio` and `asyncio.run()` from DSL examples: https://github.com/elastic/elasticsearch-dsl-py/blob/f56d3a52278e3caff045bfce320415526c56bb2d/utils/run-unasync.py#L98-L109.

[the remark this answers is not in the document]
maxjakob: Two more questions:
1. What about Generator type annotations? `AsyncGenerator[A,B]` takes 2 types and `Generator[A, B, C]` takes 3. It seems this is not handled. 
2. It seems that import paths are not adjusted or maybe I'm doing something wrong. `from elasticsearch.vectorstore._async._utils import foo` remains the same.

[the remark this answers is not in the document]
maxjakob: Findings:
1. Can be solved by using `AsyncIterator[A]`.
2. Can be solved through `additional_replacements`.
3. I guess the import hurts noone.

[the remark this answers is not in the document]
pquentin: 1. Good if AsyncIterator works for you.
2. Correct, see https://github.com/elastic/elasticsearch-dsl-py/blob/b5800d48139c99ca5de3540f74f7863c8d51adf9/utils/run-unasync.py#L51
3. It will probably fail lint, you could use sed to remove it alongside the pytest-asyncio decorators.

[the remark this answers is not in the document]
miguelgrinberg: @maxjakob I have used unasync to do the async conversion of the DSL client recently. As you noticed there are some things that unasync does not handle, and for the DSL client I have added an extra pass that addresses these using `sed`. Happy to discuss the details if it is helpful.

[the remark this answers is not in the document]
maxjakob: @miguelgrinberg I'm already heavily copying from you :) see the changes in run-unasync

test_elasticsearch/test_server/test_vectorstore/test_vectorestore copy.py1
```diff
@@ -0,0 +1,986 @@
+import logging
```
pquentin, line 1: Should this file be deleted?

examples/bulk-ingest/bulk-ingest.py
```diff
@@ -6,13 +6,14 @@
 """Script that downloads a public dataset and streams it to an Elasticsearch cluster"""
 
 import csv
-from os.path import abspath, join, dirname, exists
+from os.path import abspath, dirname, exists, join
```
pquentin, line 9: Please use `nox -rs format` to avoid unrelated changes like this. While we should apply isort/black to those files, it should be in a separate pull request.

pquentin (commented):
Reviewed two small things, but please let me know when you need a more complete review.

test_elasticsearch/test_store_integration/docker-compose.yml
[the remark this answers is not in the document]
maxjakob: I'm relying on a `elasticsearch_url` fixture now but that doesn't seem to do the trick yet. The integration tests fail in GitHub actions (`SKIPPED://localhost:9200 [...] Connection refused`) and in BuildKite (`SSLError [...] unable to get local issuer certificate`).

maxjakob:
@pquentin Can I get your eyes on this again? 👀 Some questions from my side:
1. Do we agree on the general interfaces?
2. Can you give me some pointers on how to best the vectorstore up as an optional dependency?
3. Do we need to do anything special for the `-serverless` package?

test_elasticsearch/test_store_integration/docker-compose.yml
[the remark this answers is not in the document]
pquentin: Yes, in Buildkite we run Elasticsearch with a specific TLS certificate that you need to set, either by using `sync_client` (I can add `async_client` too) or setting `ca_certs` as done here:

https://github.com/elastic/elasticsearch-py/blob/57210c683ea8a6ef36fcc5867e192e35af524935/test_elasticsearch/test_server/conftest.py#L37-L39

test_elasticsearch/test_server/test_helpers_vectorstore/docker-compose.yml
maxjakob, line 1: We can also remove this and I keep this for local testing somewhere else.

test_elasticsearch/test_server/conftest.py
```diff
@@ -30,24 +30,30 @@
 ELASTICSEARCH_REST_API_TESTS = []
 
 
+def _create(elasticsearch_url, transport=None):
+    # Configure the client with certificates
+    kw = {}
+    if elasticsearch_url.startswith("https://"):
```
maxjakob, line 36: added this condition so that security-disabled local testing also works

test_elasticsearch/test_server/conftest.py
```diff
@@ -30,24 +30,30 @@
...
-
-        # We do this little dance with the URL to force
-        # Requests to respect 'headers: None' within rest API spec tests.
-        client = elasticsearch.Elasticsearch(elasticsearch_url, **kw)
-
-        # Wipe the cluster before we start testing just in case it wasn't wiped
-        # cleanly from the previous run of pytest?
-        wipe_cluster(client)
-
+        client = _create(elasticsearch_url)
```
maxjakob, line 56: there might be a more pytestonic way to parametrize the fixture to optionally take a custom transport class, but I think this might be easier to understand?

maxjakob:
Alright @pquentin, this is ready for a bigger round of review. Let me know if you have question or if you prefer for me to guide you through this.

elasticsearch/helpers/vectorstore/_async/_utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+async def model_must_be_deployed(client: AsyncElasticsearch, model_id: str) -> None:
+    try:
+        dummy = {"x": "y"}
+        await client.ml.infer_trained_model(model_id=model_id, docs=[dummy])
+    except NotFoundError as err:
+        raise err
+    except ConflictError as err:
+        raise NotFoundError(
```
miguelgrinberg, line 33: I assume a 409 in this case means that the model is found, but it is in a state that it cannot perform inferences. If this is correct, why do we want to switch the error to not found, which can be misleading?

elasticsearch/helpers/vectorstore/_async/embedding_service.py
```diff
@@ -0,0 +1,97 @@
...
+
+from elasticsearch import AsyncElasticsearch
+
+
+class AsyncEmbeddingService(ABC):
+    @abstractmethod
+    async def embed_documents(self, texts: List[str]) -> List[List[float]]:
+        """Generate embeddings for a list of documents.
+
+        Args:
```
miguelgrinberg, line 29: Will sphinx be able to parse this syntax for args and return values? I would suggest for consistency with the existing docs you use the standard format:

```
    :param texts: A list of document strings to generate embeddings for.

    :return: A list of embeddings, one for each document in the input.
```

elasticsearch/helpers/vectorstore/_async/embedding_service.py
```diff
@@ -0,0 +1,97 @@
...
+    ):
+        """
+        Args:
+            agent_header: user agent header specific to the 3rd party integration.
+                Used for usage tracking in Elastic Cloud.
+            model_id: The model_id of the model deployed in the Elasticsearch cluster.
+            input_field: The name of the key for the input text field in the
+                document. Defaults to 'text_field'.
+            num_dimensions: The number of embedding dimensions. If None, then dimensions
+                will be infer from an example inference call.
```
miguelgrinberg, line 72: typo: s/infer/inferred

elasticsearch/helpers/vectorstore/_async/embedding_service.py
```diff
@@ -0,0 +1,97 @@
...
+    - https://www.elastic.co/guide/en/machine-learning/current/ml-nlp-deploy-models.html
+    """  # noqa: E501
+
+    def __init__(
+        self,
+        es_client: AsyncElasticsearch,
+        user_agent: str,
+        model_id: str,
+        input_field: str = "text_field",
+        num_dimensions: Optional[int] = None,
```
miguelgrinberg, line 62: Is `num_dimensions` used anywhere else? Because in this module it is stored but not used for anything, right?

elasticsearch/helpers/vectorstore/_async/strategies.py
```diff
@@ -0,0 +1,452 @@
...
+        pass
+
+    def needs_inference(self) -> bool:
+        """
+        TODO
+        """
+        return False
+
+
+class AsyncSparseVector(AsyncRetrievalStrategy):
```
miguelgrinberg, line 92: I'm wondering if there is any risk of naming collisions for these strategies. The names `DenseVector`, `SparseVector`, `BM25` seem pretty vague and likely to represent other things that are not strategies. Any reason why we are not using a `Stragegy` suffix on all these classes?

elasticsearch/helpers/vectorstore/_async/vectorstore.py
```diff
@@ -0,0 +1,384 @@
...
+        self,
+        es_client: AsyncElasticsearch,
+        index_name: str,
+        retrieval_strategy: AsyncRetrievalStrategy,
+        embedding_service: Optional[AsyncEmbeddingService] = None,
+        num_dimensions: Optional[int] = None,
+        text_field: str = "text_field",
+        vector_field: str = "vector_field",
+        metadata_mappings: Optional[Dict[str, Any]] = None,
+        user_agent: str = f"es-py-vs/{lib_version}",
```
miguelgrinberg, line 54: Are we planning for this user agent to only represent standalone use of this library, and for LangChain/LlamaIndex/etc. integrations to continue overwriting the agent with their own?

test_elasticsearch/test_server/conftest.py
```diff
@@ -30,24 +30,30 @@
...
-        if "PYTHON_CONNECTION_CLASS" in os.environ:
-            kw["node_class"] = os.environ["PYTHON_CONNECTION_CLASS"]
-
-        # We do this little dance with the URL to force
-        # Requests to respect 'headers: None' within rest API spec tests.
-        client = elasticsearch.Elasticsearch(elasticsearch_url, **kw)
-
-        # Wipe the cluster before we start testing just in case it wasn't wiped
-        # cleanly from the previous run of pytest?
-        wipe_cluster(client)
```
miguelgrinberg, line 49: This wipe before the yield has been removed in the refactored version, I think.

test_elasticsearch/test_server/test_helpers_vectorstore/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+
+  # Currently fails on Mac: https://github.com/elastic/elasticsearch/issues/106206
+  elasticsearch-with-models:
+    image: docker.elastic.co/eland/eland
+    depends_on:
+      - elasticsearch
+    restart: no
+    command: sh -c "sleep 10 && eland_import_hub_model --url http://elasticsearch:9200 --hub-model-id sentence-transformers/msmarco-minilm-l-12-v3 --start"
+
+  kibana:
```
miguelgrinberg, line 29: Do we need Kibana for anything here? Based on the work I've done to get an easy method of installation, I believe this is not going to give you a working Kibana since the Kibana password isn't configured.

utils/run-unasync.py
```diff
@@ -16,42 +16,97 @@
...
+                "AsyncTransport": "Transport",
+                "AsyncVectorStore": "VectorStore",
+                "async_bulk": "bulk",
+                "_async": "_sync",
+            },
+        ),
+        cleanup_patterns=[
+            "/^import asyncio$/d",
+            "/^import pytest_asyncio*/d",
+            "/ *@pytest.mark.asyncio$/d",
```
miguelgrinberg, line 106: These replacements would be necessary only for tests and examples, so they would not be needed on the vectorstore directory.

miguelgrinberg:
I did a lightweight review. Overall this looks great, I'm excited that with this we will be able to show high-level examples without requiring the use of a third-party library.

I noted a few comments in the code (nothing major). And there are a couple of things that are missing that I assume you are aware but will note them here just in case:

- Adding the new classes to the documentation (this may require reformatting the docstrings to follow the sphinx format, as I noted in the code)
- Async fixtures and tests

elasticsearch/helpers/vectorstore/_async/embedding_service.py
maxjakob: Good point, this is a left-over from when I had the number of dimensions be part of the embedding service, which proved to create some problems. Cleaning up.

elasticsearch/helpers/vectorstore/_async/strategies.py
maxjakob: Mainly because aesthetically I didn't want to have the Java-esque `AsyncDenseVectorScriptScoreStrategy` 😂 
But I get your point and it's probably the right thing to do.

elasticsearch/helpers/vectorstore/_async/vectorstore.py
maxjakob: Yes exactly. But we also want to track how many people do use this class directly.

test_elasticsearch/test_server/conftest.py
maxjakob: Great spot, eagle eye 🦅

test_elasticsearch/test_server/test_helpers_vectorstore/docker-compose.yml
maxjakob: No, don't need it. Left-over from the original file I copied. Removing.

elasticsearch/helpers/vectorstore/_async/_utils.py
maxjakob: Let me reproduce this again to understand the failure modes. At the time I was thinking that for the user it's the same, they need to deploy the model (I think because in one case it's not there at all and in the other it's downloaded but not deployed, but let me confirm).

maxjakob: Confirmed. I will raise the specific error, agree it's better

elasticsearch/helpers/vectorstore/_async/strategies.py
```diff
@@ -0,0 +1,454 @@
...
+        vector_field: str,
+        k: int,
+        num_candidates: int,
+        filter: List[Dict[str, Any]] = [],
+    ) -> Dict[str, Any]:
+        if not query_vector:
+            raise ValueError("specify a query_vector")
+
+        if self.distance is DistanceMetric.COSINE:
+            similarityAlgo = (
```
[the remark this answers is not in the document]
maxjakob: Testing the error states now. Assertions for the ES queries are part of the integration tests.

elasticsearch/helpers/vectorstore/_async/vectorstore.py
```diff
@@ -0,0 +1,384 @@
...
+
+    Depending on the strategy, vector embeddings are
+      - created by the user beforehand
+      - created by this AsyncVectorStore class in Python
+      - created in-stack by inference pipelines.
+    """
+
+    def __init__(
+        self,
+        es_client: AsyncElasticsearch,
```
pquentin, line 53: Can we call this `client`? We're trying to standardize on that. https://github.com/elastic/elasticsearch-py/pull/2486 is an example.

elasticsearch/helpers/vectorstore/_async/vectorstore.py
```diff
@@ -0,0 +1,384 @@
...
+        index_name: str,
```
pquentin, line 54: Can we call this `index` for consistency with the existing APIs?

pquentin, line 54: I believe everything that is not client should be keyword-only, again for consistency with the other APIs. This can be done by adding a `*` between, `client` and `index`.

test_elasticsearch/test_strategies.py
```diff
@@ -29,11 +29,27 @@ def test_sparse_vector_strategy_raises_errors():
...
-        strategy.es_query(None, None, "text_field", "vector_field", 10, 20, [])
+        strategy.es_query(
+            query=None,
+            query_vector=None,
+            text_field="text_field",
+            vector_field="vector_field",
+            k=10,
+            num_candidates=20,
+            filter=[],
+        )
```
pquentin, line 40: Much clearer, thanks.

pquentin (approved):
Thanks! LGTM.

PR #2531, opened 2024-04-22
Add Connector API

pquentin:
Each new API unfortunately requires manual steps.

### DIFF
--- a/docs/sphinx/api.rst
+++ b/docs/sphinx/api.rst
@@ -21,6 +21,7 @@ arguments are required for all calls.
    api/cat
    api/ccr
    api/cluster
+   api/connector
    api/dangling-indices
    api/enrich-policies
    api/eql
--- a/docs/sphinx/api/connector.rst
+++ b/docs/sphinx/api/connector.rst
@@ -1,0 +1,9 @@
+.. _cluster:
+
+Connector
+---------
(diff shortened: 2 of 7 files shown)


JoshMock (approved):

miguelgrinberg (approved):
The change looks good, but I have questions:

1. Why are we not adding tests for this new API?
2. How does the spec fit in this? It seems this is an API that has been added manually. When will we be able to auto-generate this code from a common spec?
3. What's up with the builds, which have lots of connection errors, report 14% coverage, and report success?

pquentin:
1. Tests need to be added there: https://github.com/elastic/elasticsearch-clients-tests. (But I have to switch from the Elasticsearch server tests to those first.)
2. The Python code generator is limited: it only generated the connector.py classes, but everything else (connector.py imports, exposing this new namespace to the rest of the client, and docs) is manual. This work is only needed when there's a new namespace, but it happens a lot these days even though we "only" have ~40 namespaces.
3. The build is separated in two: the tests that don't need an Elasticsearch server run in GitHub, and the one that do are skipped in GitHub and only run in Buildkite. The real server tests are missing for now, but they add about 40 minutes of CI time currently. I hope that https://github.com/elastic/elasticsearch-clients-tests will be much faster.

PR #2537, opened 2024-04-30
Async tests and generated sync tests

maxjakob:
Move a chunk of tests to the unasync pattern. There are still a lot left to migrate. My proposal would be to start with this and move the rest later.

## Currently not part of PR
```
├── test_async
│   └── test_transport.py
├── test_cases.py
├── test_client
│   ├── test_cluster.py
│   ├── test_deprecated_options.py
│   ├── test_indices.py
│   ├── test_options.py
│   ├── test_overrides.py
│   ├── test_requests_auth.py
│   ├── test_rewrite_parameters.py
│   ├── test_serializers.py
│   └── test_utils.py
├── test_exceptions.py
├── test_helpers.py
├── test_otel.py
├── test_serializer.py
├── test_server
│   ├── conftest.py
│   ├── test_clients.py
│   ├── test_otel.py
│   ├── test_responses.py
│   └── test_vectorstore
│       ├── conftest.py
│       ├── docker-compose.yml
│       ├── test_embedding_service.py
│       └── test_vectorstore.py
├── test_strategies.py
├── test_transport.py
├── test_types
│   ├── aliased_types.py
│   ├── async_types.py
│   └── sync_types.py
```


pquentin:
> Why is this failing for 3.7 now?

This is because pytest 8 dropped support for Python 3.7 which is using pytest 7. It looks like pytest 8 allows things that pytest 7 does not.

test_elasticsearch/_async/test_server/test_clients.py
```diff
@@ -24,8 +24,8 @@
 
 
 @pytest.mark.parametrize("kwargs", [{"body": {"text": "привет"}}, {"text": "привет"}])
-async def test_indices_analyze_unicode(async_client, kwargs):
```
maxjakob, line 27: I think I might have introduced this warning:
```
test_elasticsearch/_async/test_server/test_clients.py::test_indices_analyze_unicode[kwargs0]
--
  | /code/elasticsearch-py/.nox/test-3-12/lib/python3.12/site-packages/pytest_asyncio/plugin.py:759: DeprecationWarning: The event_loop fixture provided by pytest-asyncio has been redefined in
  | /code/elasticsearch-py/test_elasticsearch/_async/conftest.py:30
  | Replacing the event_loop fixture with a custom implementation is deprecated
  | and will lead to errors in the future.
  | If you want to request an asyncio event loop with a scope other than function
  | scope, use the "scope" argument to the asyncio mark when marking the tests.
  | If you want to return different types of event loops, use the event_loop_policy
  | fixture.
```
https://buildkite.com/elastic/elasticsearch-py-integration-tests/builds/1638#018f729e-d596-42a8-aa94-6ed2036e0573

but I can also see it for a test that I didn't touch: `
test_elasticsearch/test_transport.py::test_warning_header[headers1]`.

maxjakob: Have to look more at event loop policies https://docs.python.org/3/library/asyncio-policy.html

maxjakob: We get this warning for Python 3.8 and above. For 3.7 there is no warning. And I added override of the even loop fixture to fix an error with 3.7:
https://github.com/elastic/elasticsearch-py/blob/370975ef6d01692e7626b3ae3be3863363946796/test_elasticsearch/_async/conftest.py#L29-L31

@pquentin This might be the trade-off we have to make here. We live with the new warning until we remove support for Python 3.7. What do you think?

maxjakob: Decided to stop testing 3.7.

maxjakob:
`test-3.7` might still be set as a required job in GitHub.

pquentin (commented):
While I admire your ability to make things work (including the Python 3.7 issues), I'm seeing a few problems here:

 * This duplicates code between _async/_sync and the tests that are not unasynced yet. This is expected since it's a partial migration, but we don't have a clear path towards a full migration yet, as I have other (conflicting) priorities, most notably updating the serverless client to keep up with the main client and integrating https://github.com/elastic/elasticsearch-clients-tests
 * This does not test `httpx` and doing so with the current framework would double the number of tests unnecessarily, as it would test each transport twice: urllib3 with aiohttp, urllib3 with httpx, requests with aiohttp and requests with httpx.
 * The solution to the above problem is to specify "urllib3,aiohttp" and "requests,httpx", and then transform that in sync connection class and async connection class inside pytest, somehow? Adding even more complexity to our pytest setup.
 * We instantiate clients in many ways today (_create with/without node class and with/without transport, sync_client_request_saving_factory, sync_client_request_saving, sync_client, async_client) with conftest.py files all over the place, which adds a lot of complexity too.
 * I haven't even looked at the actual tests changes yet.
 * Additionally, the value of async tests is a bit low, as we have confidence in our existing testing of node classes and reliability of unasync. And indeed this pull request did not uncover any issues.

As a result, I'm not going to merge this now, but instead try to tame the current complexity, and slowly work towards adding more async tests in the future while keeping the code base maintainable. I will definitely reuse this work at some point, though, so this effort was not a waste of time.

Thank you for your understanding.

PR #2561, opened 2024-05-17
Drop Python 3.7 support

kloczek:
According to https://endoflife.date/python 3.7 has been EOSed almost year ag.
Filter all code over `pyupgrade --py38-plus`.


kloczek:
CLA signed

setup.py
```diff
@@ -75,9 +74,6 @@
         "License :: OSI Approved :: Apache Software License",
         "Intended Audience :: Developers",
         "Operating System :: OS Independent",
-        "Programming Language :: Python",
```
pquentin, line 78: Please keep this line.

pquentin (commented):
Thanks for this. I know Python 3.7 is EOL since June 2023, but it's still ~10% of elasticsearch-py 8.13 downloads. My current plan is to drop Python 3.7 support in version 8.15, so I will review and merge this after 8.14 is out.

pquentin:
buildkite test this please

pquentin (approved):

miguelgrinberg (approved):
LGTM, but should this be labeled as a code refresh? Looks like pyupgrade was applied to the code base, and that is more representative of the change than the drop of 3.7.

pquentin:
@miguelgrinberg In my experience running pyupgrade and making various code changes as part of dropping a Python version is pretty common. For example, in urllib3:

* https://github.com/urllib3/urllib3/pull/1429
* https://github.com/urllib3/urllib3/pull/2044
* https://github.com/urllib3/urllib3/pull/2336
* https://github.com/urllib3/urllib3/pull/3143

Admittedly, some of the changes here are not only related to Python 3.7, so I can open a pull request to run pyupgrade --py37-plus before merging this if you prefer.

miguelgrinberg:
@pquentin No need to change anything, this is fine. I think this is just that I'm used to handle Python version upgrades differently. My assumption when I saw "Drop 3.7"  was that we would just stop testing 3.7, which is what I do for my personal projects (because I intentionally want to let people continue using the dropped version even after I stopped testing it or supporting it). Your meaning of "Drop 3.7" is larger in scope than mine. :)

pquentin:
Ah right! When I drop a version, I update `python_requires`, so anyone still using Python 3.7 will be stuck on 8.14. And it's the best moment to drop all the workarounds that were need for previous versions. In the early 3.x days, many changes were related to the ssl module, and now it's mostly typing.

Issue #789, opened 2018-05-11
[BUG] http_compress  error  with python 2.7.x


hysia:
[https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/http_urllib3.py#L164](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/http_urllib3.py#L164)

`gzip.compress()`  is not correct in python2.7 use `gzip.zlib.compress()` instead otherwise it will throw an error  `object has no attribute "compress"`  
  
[![C0NTN8.md.png](https://s1.ax1x.com/2018/05/11/C0NTN8.md.png)](https://imgchr.com/i/C0NTN8)

PR #2563, opened 2024-05-22
[DOCS] Add ES|QL doc structure

marciw:
WIP, not ready to review

Adds an esql.asciidoc file that sets up a basic structure for the ES|QL documentation. Based on @ szabosteve's Java client docs PR ([#789](https://github.com/elastic/elasticsearch-java/pull/789/files)).

This includes adding a folder structure for helpers instead of keeping the docs all on one page (but we can undo that).

@ pquentin This needs expanding and tailoring for the Python client (it's really just placeholder text from the Java client docs). All other feedback is welcome, especially because I'm still so new to Elastic and to the client docs.

Preview
coming soon

### DIFF
--- a/docs/guide/helpers/esql.asciidoc
+++ b/docs/guide/helpers/esql.asciidoc
@@ -1,0 +1,79 @@
+[[esql]]
+=== ES|QL in the Python client
+++++
+<titleabbrev>ES|QL</titleabbrev>
+++++
+
+This page helps you understand and use {ref}/esql.html[ES|QL] in the
+Python client.
+
+There are two ways to use ES|QL in the Python client:
+
+* Use the Elasticsearch {es-docs}/esql-apis.html[ES|QL API] directly: This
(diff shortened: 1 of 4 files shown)


marciw:
>The main issue here is that Python does not have ES|QL helpers yet and there are no immediate plans to add some. 

Ohhhhhh, ha, thanks for clarifying 👍 

> One option would be to add https://www.elastic.co/search-labs/blog/esql-pandas-dataframes-python under "Integrations -> Pandas".

This sounds like a great idea in the short term!

> Another option would be to add a new page explaining how to make queries, including the Query DSL and ES|QL. Any thoughts?

... and then this sounds like a great idea in the long (or medium) term, at which point we could decide whether to keep the Pandas integration doc with tweaks, or move the content to the new queries page.

WDYT? Either way, I can start the PR(s) w/ some structure and placeholder text, if you like. (Will probably close this one)

PR #2599, opened 2024-06-26
Convert README to markdown for static HTML embedding

pquentin:
Removing the custom processing in setup.py will allow moving to Hatch in the future, to reduce differences with elasticsearch-serverless-python.

You can preview the README here: https://github.com/pquentin/elasticsearch-py/blob/readme.md/README.md but the logo is broken so I've attached a screenshot:

image


README.md
```diff
@@ -0,0 +1,109 @@
...
+Language clients are forward compatible; meaning that the clients support
+communicating with greater or equal minor versions of Elasticsearch without
+breaking. It does not mean that the clients automatically support new features
+of newer Elasticsearch versions; it is only possible after a release of a new
+client version. For example, a 8.12 client version won't automatically support
+the new features of the 8.13 version of Elasticsearch, the 8.13 client version
+is required for that. Elasticsearch language clients are only backwards
+compatible with default distributions and without guarantees made.
+
+| Elasticsearch Version | Elasticsearch-PHP Branch | Supported |
```
miguelgrinberg, line 75:
```suggestion
| Elasticsearch Version | Elasticsearch-Python Branch | Supported |
```

miguelgrinberg (approved):
LGTM, but there is a PHP mention that needs to be fixed.

README.md
pquentin: Oh wow, great catch! PHP already had the table in Markdown so I copied it and missed that it said PHP. Thanks for reading carefully.

PR #2602, opened 2024-07-01
feat: extend default settings in AsyncVectorStore

minglu7:
This commit introduces a new parameter, custom_settings, to the AsyncVectorStore class. This allows users to provide their own settings that will extend the default settings. This increases the flexibility of the class and allows it to be tailored to specific use cases. The custom settings are applied in the _create_index_if_not_exists method.

### DIFF
--- a/elasticsearch/helpers/vectorstore/_async/vectorstore.py
+++ b/elasticsearch/helpers/vectorstore/_async/vectorstore.py
@@ -60,7 +60,7 @@ class AsyncVectorStore:
         vector_field: str = "vector_field",
         metadata_mappings: Optional[Dict[str, Any]] = None,
         user_agent: str = f"elasticsearch-py-vs/{lib_version}",
-        custom_settings: Optional[Dict[str, Any]] = None,
+        custom_index_settings: Optional[Dict[str, Any]] = None,
     ) -> None:
         """
         :param user_header: user agent header specific to the 3rd party integration.
@@ -73,6 +73,12 @@ class AsyncVectorStore:
             the embedding vector goes in this field.
         :param client: Elasticsearch client connection. Alternatively specify the
             Elasticsearch connection with the other es_* parameters.
+        :param custom_index_settings: A dictionary of custom settings for the index.
+        This can include configurations like the number of shards, number of replicas,
+        analysis settings, and other index-specific settings. If not provided, default
+        settings will be used. Note that if the same setting is provided by both the user
+        and the strategy, the value from the strategy will take precedence. This behavior
+        ensures that the default settings have a higher priority.
         """
         # Add integration-specific usage header for tracking usage in Elastic Cloud.
         # client.options preserves existing (non-user-agent) headers.
@@ -91,7 +97,7 @@ class AsyncVectorStore:
         self.text_field = text_field
         self.vector_field = vector_field
         self.metadata_mappings = metadata_mappings
-        self.custom_settings = custom_settings
+        self.custom_index_settings = custom_index_settings
 
     async def close(self) -> None:
         return await self.client.close()
@@ -308,9 +314,9 @@ class AsyncVectorStore:
             vector_field=self.vector_field,
             num_dimensions=self.num_dimensions,
         )
-        if self.custom_settings:
-            self.custom_settings.update(settings)
-            settings = self.custom_settings
+        if self.custom_index_settings:
+            self.custom_index_settings.update(settings)
+            settings = self.custom_index_settings
 
         if self.metadata_mappings:
             metadata = mappings["properties"].get("metadata", {"properties": {}})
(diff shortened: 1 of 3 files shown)


elasticsearch/helpers/vectorstore/_async/vectorstore.py
```diff
@@ -60,6 +60,7 @@ def __init__(
         vector_field: str = "vector_field",
         metadata_mappings: Optional[Dict[str, Any]] = None,
         user_agent: str = f"elasticsearch-py-vs/{lib_version}",
+        custom_settings: Optional[Dict[str, Any]] = None,
```
miguelgrinberg, line 63:
```suggestion
        custom_index_settings: Optional[Dict[str, Any]] = None,
```

I think it is best to clarify that these are settings for the index.

elasticsearch/helpers/vectorstore/_async/vectorstore.py
```diff
@@ -306,6 +308,10 @@ async def _create_index_if_not_exists(self) -> None:
             vector_field=self.vector_field,
             num_dimensions=self.num_dimensions,
         )
+        if self.custom_settings:
+            self.custom_settings.update(settings)
```
miguelgrinberg, line 312: The way you coded this means that if the same setting is provided by the user and set by the strategy, then the value from the strategy wins. Is this what you intended? To me it seems more reasonable to do the reverse, which would allow the caller to override settings.

Either way, the behavior should be documented in the docstring above.

elasticsearch/helpers/vectorstore/_async/vectorstore.py
minglu7: i see, thx

elasticsearch/helpers/vectorstore/_async/vectorstore.py
minglu7: I believe the settings in the strategy should have a higher priority because I am concerned that users might overlook the default strategy settings, which could lead to unknown errors. Of course, I also respect your possible different opinion, and if you think it's necessary to make other changes, please let me know.

pquentin:
buildkite test this please

pquentin (changes requested):
We're getting there. The two immediate issues;

* There's still a difference between the `async` and `sync` version. Running `nox -rs format` should fix it.
* You need to call `add_texts()` in your tests, as done in the previous test. The index is created only when calling `add_texts()`.

pquentin:
buildkite test this please

minglu7:
> buildkite test this please

it seems that docs/sphinx/conf.py should be reformatted

minglu7:
buildkite test this please

minglu7:
> buildkite test this please

send it again please

elasticsearch/helpers/vectorstore/_async/vectorstore.py
miguelgrinberg: @pquentin do you have an opinion on this?

I'm not sure I like the idea that a custom setting that you pass is silently ignored if the selected strategy uses a different value, first because it is silent so the developer may not realize a setting isn't being applied, and second because it provides to option to override the settings in the strategy.

The ideal solution would be:

- if there is a collision between settings passed by the user and by the strategy, then raise an error
- add another boolean argument that the user can set to `True` to indicate that overriding of strategy settings should be allowed, in which case the error will not be raised and the custom settings from the user take precedence.

pquentin:
buildkite test this please

elasticsearch/helpers/vectorstore/_async/vectorstore.py
pquentin: Yes, I also think that if you specify a setting, it should be used and not silently discarded as done now.

I'm happy to raise an error if there's a collision, but I don't think we should bother with another boolean argument. The strategy settings exist for a reason, and the custom user settings are going to be quite different anyway (number of primary/replicas, etc.). For this reason, we can probably limit collision detection to top-level settings keys for now. And all of this can be reevaluated if we have use case that requires more complexity.

minglu7:
> buildkite test this please
i have changed the files in above suggestions

elasticsearch/helpers/vectorstore/_async/vectorstore.py
miguelgrinberg: @minglu7 Let's go with @pquentin simplified proposal for now.

- If a top-level setting is in both groups of settings, then raise an error
- Else, a merge can be done cleanly so no problem.

In the future, we can re-evaluate if an override option is needed, but for now let's avoid this complication.

minglu7:
> buildkite test this please

check these changed files, please

elasticsearch/helpers/vectorstore/_async/vectorstore.py
minglu7: @miguelgrinberg i have did these changes as you wish, please check them

pquentin:
buildkite test this please

miguelgrinberg:
buildkite test this please

miguelgrinberg (approved):
Looks good. Thank you so much @minglu7 !

PR #2605, opened 2024-07-01
check that unasync'ed files are up to date

miguelgrinberg:
This is adapted from a lint check we've added to the Python DSL client to make sure we never commit unasync'ed files that are stale or missing.


pquentin (approved):
Thanks! LGTM.

PR #2607, opened 2024-07-02
Merge Query Ruleset API into new Query Rules API

pquentin:
As done in https://github.com/elastic/elasticsearch-serverless-python/pull/71.


miguelgrinberg:
Not happy about breaking backwards compatibility as previously discussed, but...

PR #2619, opened 2024-07-31
Remove quotes around typing.Literal

pquentin:
Now that all supported Python versions know about it.


miguelgrinberg (approved):
The CI says this is okay, so I agree ;-)

PR #2621, opened 2024-08-01
Add Connector API

pquentin:
As added in https://github.com/elastic/elasticsearch-py/pull/2531 initially, but reverted in https://github.com/elastic/elasticsearch-py/pull/2553 because it was not ready for 8.14.


miguelgrinberg (approved):
Interestingly, I feel great about allowing this type of change because the request-converter indirectly is going to make sure the client is doing the right thing here.

PR #2636, opened 2024-08-09
Release 8.15.0

miguelgrinberg:


pquentin:
Can you please add https://github.com/elastic/elasticsearch-specification/pull/2622 to the release notes?

miguelgrinberg:
@pquentin added https://github.com/elastic/elasticsearch-specification/pull/2622 to the notes, but as far as I can see there is nothing specific that changed in the client to support semantic text. Does it make sense for spec changes in the areas that are not code generated to be reflected in the client change log? I'm not sure if there's something I'm missing, but I believe the 8.14 client can do semantic text as well as the 8.15 client can, as long as you are talking to an 8.15+ server.

pquentin (approved):
Of course, you're right, it's not needed. I guess we can keep it now it's there, to avoid questions, but it's a bit misleading and can be avoided in the future. Thanks! LGTM.

PR #2641, opened 2024-08-14
Configure mypy to ignore missing imports during build

miguelgrinberg:
The build script runs mypy without the `ignore-missing-imports` flag. This used to be fine, but with the introduction of pyarrow the flag is needed because this library does not have type hints. The mypy build triggered from nox works fine because it already has this option enabled.


pquentin (approved):
Thanks! LGTM. We need to use the configuration file more.

PR #2643, opened 2024-08-22
use request converter to generate python examples

miguelgrinberg:
This PR replaces the `generate-examples.py` script with a new version written in JavaScript (heavily based on the JS client's example generator) that uses the request converter library.


utils/generate-docs-examples/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "generate-docs-examples",
+  "version": "1.0.0",
+  "main": "generate-docs-examples.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
pquentin: I realize this is the default package.json, but the license isn't ISC but Apache 2.0 and we don't need tests. Maybe all those lines can go away?

pquentin (approved):
Thanks! LGTM.

utils/generate-docs-examples/package.json
miguelgrinberg: Sorry, I need to fix this. ISC was added as a default when I ran `npm init -y` and I didn't notice. I'll change it to Apache.

utils/generate-docs-examples/README.md
```diff
@@ -0,0 +1,14 @@
+This directory contains a script that generates the Python documentation examples in `docs/examples`.
+
+To use this script you need a recent version of Node.js (18+). First install the dependencies:
```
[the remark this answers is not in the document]
miguelgrinberg: The request converter supports the three current LTS releases, 18, 20 and 22. I think it is best to let the user choose one instead of making a choice with an .nvmrc file.

PR #2644, opened 2024-08-26
Updated connector API documentation examples

miguelgrinberg:
I regenerated the examples with the latest request-converter, which recognizes the Connector and Inference APIs as supported by the Python client.


pquentin (approved):
Thanks! LGTM.

PR #2647, opened 2024-08-27
Remove redundat np.float_ to fix numpy 2.X in 7.X elastic

### LINKED ISSUE #2646: [Dependency]: Numpy 2.X
# Description

`numpy.float_` has been deprecated in favor of `numpy.float64` because after the update numpy always maintain float precision. This package uses numpy only when JsonSerializer is used and declare np.float_ in version 7.X [here](https://github.com/elastic/elasticsearch-py/blob/7.17/elasticsearch/serializer.py#L202)

# Version
elasticsearch : 7.17

# Proposal
Update basecode as defined in the 8.X version but for 7.X, if an update is not posible for 7.X package version, blocking numpy dependency to <2.X would be enough to avoid incompatibilites.

Please let me know if a PR is required and I will prepare it!

Thanks in advance

alvarogf97:
Resolves #2646 on 7.X

### DIFF
--- a/elasticsearch/serializer.py
+++ b/elasticsearch/serializer.py
@@ -199,7 +199,6 @@ def _attempt_serialize_numpy(data):
         elif isinstance(
             data,
             (
-                np.float_,
                 np.float16,
                 np.float32,
                 np.float64,
--- a/test_elasticsearch/test_serializer.py
+++ b/test_elasticsearch/test_serializer.py
@@ -96,7 +96,6 @@ class TestJSONSerializer(TestCase):
 
         ser = JSONSerializer()
         for np_type in (
-            np.float_,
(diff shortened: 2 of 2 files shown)


pquentin:
Thanks! The CI failures are expected. However, I can't merge this without the CLA, sorry. Make sure to use the same email as in your commit: https://github.com/elastic/elasticsearch-py/commit/b897656b1594fb68b09653291bf5e36794ad0425.patch

pquentin (approved):
Thanks! LGTM.

pquentin:
Thank you for your patience. https://github.com/elastic/elasticsearch-py/releases/tag/v7.17.12 is now out with this change.

PR #2652, opened 2024-09-08
Release 8.15.1

pquentin:

PR #2653, opened 2024-09-09
Fix build

pquentin:
* The latest release of SimSIMD breaks mypy: https://github.com/ashvardanian/SimSIMD/issues/171
* Sphinx 8 is not supported by the Sphinx RTD theme, which downgrades to 0.5 unless forced to 2.0+


pquentin:
Closing since https://github.com/ashvardanian/SimSIMD/issues/171 was fixed at warp speed and #2652 already includes the docs fix.

PR #2657, opened 2024-09-10
Updated the SparseVectorStrategy class to use sparse_vector query

miguelgrinberg:
This change updates the `SparseVectorStrategy` class to use the newer `sparse_vector` query instead of `text_expansion`, which generates a deprecation warning since 8.15.

### DIFF
--- a/elasticsearch/helpers/vectorstore/_async/strategies.py
+++ b/elasticsearch/helpers/vectorstore/_async/strategies.py
@@ -96,7 +96,7 @@ class AsyncRetrievalStrategy(ABC):
 
 
 class AsyncSparseVectorStrategy(AsyncRetrievalStrategy):
-    """Sparse retrieval strategy using the `text_expansion` processor."""
+    """Sparse retrieval strategy using the `sparse_vector` processor."""
 
     def __init__(self, model_id: str = ".elser_model_2"):
         self.model_id = model_id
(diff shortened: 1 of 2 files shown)


joemcelroy:
The complication with this issue is that `sparse_vector` query is not compatible with the `rank_features` mapping. This change will work perfectly for those who setup the index with this change (creates a `sparse_vector` type field) but will impact those who upgrade and still using `rank_features` mapping.

Possible workarounds for this is introduce a boolean to allow the "legacy" `text_expansion` query to be used in the strategy constructor. By default its the `sparse_vector` query. Add a better error to surface this configuration when ES responds with 400 error that they need to use the legacy boolean. Pain points is it will be a breaking change on upgrade.

miguelgrinberg:
@joemcelroy Discussing this within our team we were thinking that we should hold off on this change until 9.0, where we would have more freedom to introduce it as a breaking change.

Another option would be to have both the old and the new solutions in the client, and select which one to use based on the version reported by the server. But I don't think this would address a case where an index was created while running a <= 8.14 stack, but then the stack is upgraded to >= 8.15.

joemcelroy:
Yeh think thats wise to wait for 9.x. Theres nothing pressing to make this change, just the legacy warning.

> But I don't think this would address a case where an index was created while running a <= 8.14 stack, but then the stack is upgraded to >= 8.15.

Yep thats it. Also will impact those who haven't pinned their ES client to their version of ES.

pquentin (approved):

PR #2666, opened 2024-09-30
Vectorstore: use a retriever query for hybrid search

### LINKED ISSUE #2651: [Bug] Vectorstore Helper: Hybrid does not work in Elasticsearch Serverless
Expected:
When I enable hybrid and im using Elasticsearch v8.14+, I should be able to retrieve documents from both a bm25 based search and NN search using dense/sparse. 

Actual:
Cannot perform the search as RRF is not supported

Reason:
Hybrid performs two retrieval searches and relies on RRF to balance the different result scores. In 8.14+, top level rank was removed and now the only way to do RRF is through retrievers query.

This is the code that builds the query that retrieves results from different methods and balances with RRF. [link](https://github.com/elastic/elasticsearch-py/blob/main/elasticsearch/helpers/vectorstore/_sync/strategies.py#L309)

suggested untested code change to equilivent retriever is:

```
{
  "retriever": {
    "rrf": {
      "retrievers": [
        {
          "standard": {
            "query": {
              "match": {
                                self.text_field: {
                                    "query": query,
                                }
                            }
            }
          }
        },
        {
          "standard": {
            "query": {
              "knn": knn
            }
          }
        }
      ]
    }
  }
}

```

Full example of a query:

```
{
  "retriever": {
    "rrf": {
      "retrievers": [
        {
          "standard": {
            "query": {
              "multi_match": {
                "query": "{query}",
                "fields": [
                  "content"
                ]
              }
            }
          }
        },
        {
          "standard": {
            "query": {
              "knn": {
                "field": "content_embedding",
                "num_candidates": 100,
                "query_vector_builder": {
                  "text_embedding": {
                    "model_id": "openai_embeddings",
                    "model_text": "{query}"
                  }
                }
              }
            }
          }
        }
      ]
    }
  }
}
```

miguelgrinberg:
Fixes #2651

### DIFF
--- a/elasticsearch/helpers/vectorstore/_async/strategies.py
+++ b/elasticsearch/helpers/vectorstore/_async/strategies.py
@@ -283,31 +283,46 @@ class AsyncDenseVectorStrategy(AsyncRetrievalStrategy):
     ) -> Dict[str, Any]:
         # Add a query to the knn query.
         # RRF is used to even the score from the knn query and text query
-        # RRF has two optional parameters: {'rank_constant':int, 'window_size':int}
+        # RRF has two optional parameters: {'rank_constant':int, 'rank_window_size':int}
         # https://www.elastic.co/guide/en/elasticsearch/reference/current/rrf.html
+        rrf_options = {}
+        if isinstance(self.rrf, Dict):
+            if "rank_constant" in self.rrf:
+                rrf_options["rank_constant"] = self.rrf["rank_constant"]
+            if "window_size" in self.rrf:
+                # 'window_size' was renamed to 'rank_window_size', but we support
+                # the older name for backwards compatiblit
+                rrf_options["rank_window_size"] = self.rrf["window_size"]
+            if "rank_window_size" in self.rrf:
+                rrf_options["rank_window_size"] = self.rrf["rank_window_size"]
         query_body = {
-            "knn": knn,
-            "query": {
-                "bool": {
-                    "must": [
+            "retriever": {
+                "rrf": {
+                    "retrievers": [
                         {
-                            "match": {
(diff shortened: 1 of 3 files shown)


pquentin (commented):
While this looks good, can you please explain why we can no longer disable RRF with `AsyncDenseVectorStrategy(..., rrf=False)`? Should we change the annotation of that `rrf` parameter from ` Union[bool, Dict[str, Any]]` to `Dict[str, Any]`?

miguelgrinberg:
I did not consider the case of `rrf = False`, I just assumed that `rrf = False` means that hybrid search is off, but now I realize that we have two booleans.

I'm not sure what would be the expected response when having hybrid search enabled and RRF disabled. Does that even make sense? I guess I need to try it out to see what response I get.

elasticsearch/helpers/vectorstore/_async/strategies.py
```diff
@@ -285,44 +285,50 @@ def _hybrid(
...
-                    **rrf_options,
-                },
-            },
+                    "filter": filter,
+                }
+            }
         }
+
+        if self.rrf is False:
+            query_body = {
```
miguelgrinberg, line 306: With `rrf=False` we have to use a standard query, since the retriever query can only combine results from multiple searches with RRF.

miguelgrinberg:
Okay, new revision of this fix. When the user asks for `hybrid=True` and `rrf=False` we use a regular query with `knn` and `query` sections in the body. Only when `rrf=True` or `rrf={<rrf-options>}` we use the new retriever query.

pquentin (approved):
Thanks! LGTM.

PR #2667, opened 2024-10-16
Switch from Elasticsearch YAML tests to client tests

pquentin:
This is mostly adapted from the Elasticsearch Serverless tests that already do this.

@miguelgrinberg This also bumps the tested version of Elasticsearch to 8.16. How do you want to handle the vector store tests? Do we keep the conditionals just in case we want to run those tests with 8.13 in the future?


miguelgrinberg (approved):

miguelgrinberg:
@pquentin yes, I would keep the version checks for a while, in case we want to run the test suite locally.

PR #2669, opened 2024-10-17
Make BulkIndexError and ScanError serializable

seagrine:
I encountered an issue with pickling/unpickling BulkindexError.

`unpickle_exception\n    inst = func(*args)\n           ^^^^^^^^^^^\nTypeError: BulkIndexError.**init**() missing 1 required positional argument: 'errors`

I'm adding an implementation of `__reduce__` to this (and ScanError) so they can be pickled and unpickled correctly. Added tests for it as well.


pquentin:
buildkite test this please

elasticsearch/helpers/errors.py
```diff
@@ -15,18 +15,26 @@
...
+    ) -> Tuple[Type["BulkIndexError"], Tuple[str, List[Dict[str, Any]]]]:
+        return (self.__class__, (self.args[0], self.errors))
+
 
 class ScanError(Exception):
     scroll_id: str
 
-    def __init__(self, scroll_id: str, *args: Any, **kwargs: Any) -> None:
-        super().__init__(*args, **kwargs)
+    def __init__(self, scroll_id: str, shards_message: str) -> None:
```
miguelgrinberg, line 35: Are the typing changes in both exeception classes and the argument changes in `ScanError` necessary? I think you could have just added a `__reduce__` method that writes all the arguments assuming the caller is passing pickle-friendly data. Not a huge deal, but to me it seems there is no point in restricting the arguments in `ScanError`.

miguelgrinberg (approved):
LGTM, but I've noted that some changes in this PR did not seem necessary.

elasticsearch/helpers/errors.py
pquentin: They're not necessary, but `**kwargs` was simply not working, and `*args` was overly broad. Restricting to what was actually used also really improved the types. Are you thinking that other libraries could be raising those exceptions with different parameters?

pquentin: https://github.com/search?q=%22raise+BulkIndexError%22+language%3APython&type=code shows that raising BulkIndexError() is pretty common, but I've only changed the type annotation here, and all the samples I've seen indeed pass a string.

I've not seen public examples of raising `ScanError`, but I can restore `*args`.

pquentin: I've restored backwards-compatibility in [`0db50e4` (#2669)](https://github.com/elastic/elasticsearch-py/pull/2669/commits/0db50e4cee62542c74c1d5a4431bfaa170b760ce), please take another look.

pquentin:
buildkite test this please

elasticsearch/helpers/errors.py
miguelgrinberg: It's highly unlikely that someone is raising `ScanError`, but yes, I just thought it was a change that is unrelated to making the class work with pickle. But as I said, it's a small observation, I think it is unlikely to cause problems.

miguelgrinberg (approved):

pquentin:
Thank you for your contribution, it is now released as part of 8.16.0: https://github.com/elastic/elasticsearch-py/releases/tag/v8.16.0

PR #2675, opened 2024-11-06
Emit Python warnings for beta and tech preview APIs

### LINKED ISSUE #2674: Emit deprecation warnings for tech preview/beta APIs
- [ ] elasticsearch-py
- [ ] elasticsearch-serverless-python

pquentin:
Closes #2674 

The first commit is the manual-written code, while the second one is auto-generated.


elasticsearch/_sync/client/utils.py
```diff
@@ -450,6 +461,43 @@ def wrapped(*args: Any, **kwargs: Any) -> Any:
...
+            elif stability == Stability.EXPERIMENTAL:
+                warnings.warn(
+                    "This API is in technical preview and may be changed or removed in a future release. "
+                    "Elastic will work to fix any issues, but features in technical preview are not subject to the support SLA of official GA features.",
+                    category=GeneralAvailabilityWarning,
+                    stacklevel=warn_stacklevel(),
+                )
+            elif stability == Stability.DEPRECATED:
+                warnings.warn(
+                    f"This API was deprecated in Elasticsearch {version}. {message}",
```
miguelgrinberg, line 489: I don't have a solution for this, but I can foresee confusion around versions displayed in these messages. People are not going to understand that this is a client-side version.

miguelgrinberg (approved):

elasticsearch/_sync/client/utils.py
pquentin: Well, it's not a client-side version, this is really a deprecation on the server side. Do you think we could make this message clearer?

miguelgrinberg: I don't know how to solve this. My point is that the version of the client is what determines what warnings are displayed, even though the server can be on a completely different version. It is going to be confusing.

The ideal would be that these deprecation warnings are tied to the server version, but I understand that is extremely difficult or impossible to do.

pquentin: We actually print the deprecation warnings that are handed to us by the server:

https://github.com/elastic/elasticsearch-py/blob/94573cf5b050627a9778d22e5cb16149917767ad/elasticsearch/_sync/client/_base.py#L372-L384

Maybe we should stick to that and not print other warnings. I'll check if we get a warning for the two deprecations added in this pull request.

miguelgrinberg: Does the server warn for tech preview features? I think the most important reason for these warnings we are adding is to inform people about features that are likely to change.

pquentin: No, it does not warn about tech preview features. This thread is about warning about the two deprecated features or not. I did check, and `_knn_search` does not warn, but `post_data` does, which results in a double warning.

```python
$ python t.py
/Users/quentin/src/elasticsearch-py/t.py:41: DeprecationWarning: This API was deprecated in Elasticsearch 7.11.0. Posting data directly to anomaly detection jobs is deprecated, in a future major version a datafeed will be required.
  client.ml.post_data(
/Users/quentin/src/elasticsearch-py/t.py:41: ElasticsearchWarning: Posting data directly to anomaly detection jobs is deprecated, in a future major version it will be compulsory to use a datafeed
  client.ml.post_data(
```

So I'm going to leave out that part, and consider the fact that the server does not warn about `_knn_search` an Elasticsearch bug.

pquentin: No I understand that your comment also applies to tech preview features that are indeed only client-based! (I was confused because the line you commented on is about deprecated features.) And yes, if an API becomes GA in version 8.x but you're using client 8.(x-1), you're going to get the warning. There's not much that I can do about that.

miguelgrinberg: Right, I know it is a problem with no solution, and for tha treason I approved. This is better than nothing.

pquentin:
I added a few commits (stop warning about deprecated features, and fix Python 3.12 that was seeing unrelated warnings in `recwarn.list`). I'd appreciate a new review.

miguelgrinberg (approved):

PR #2678, opened 2024-11-07
Pass parameter `nodes` instead of `node_id` from the list function in TasksClient

margaretpearce:
Closes https://github.com/elastic/elasticsearch-specification/issues/3112

Pass a parameter named `nodes` instead of `node_id` to `perform_request` if `node_id` function argument is not None on `list` in TasksClient.


pquentin:
Thank you! Appreciate that you took the time to open a pull request. However this code is generated, so I am closing in favor of https://github.com/elastic/elasticsearch-specification/issues/3112.

PR #2694, opened 2024-11-11
Fix import when trace is missing from opentelemetry

nicoloboschi:
otel dependency is optional but if I try to use latest elastic client it fails to bootstrap due to 
```
../../../vdp/config_verification/elastic_sink/elastic_config_verification.py:21: in <module>
../../../../.venv/lib/python3.11/site-packages/elasticsearch/__init__.py:45: in <module>
    from ._async.client import AsyncElasticsearch as AsyncElasticsearch
../../../../.venv/lib/python3.11/site-packages/elasticsearch/_async/client/__init__.py:38: in <module>
    from ._base import (
../../../../.venv/lib/python3.11/site-packages/elasticsearch/_async/client/_base.py:48: in <module>
    from ..._otel import OpenTelemetry
../../../../.venv/lib/python3.11/site-packages/elasticsearch/_otel.py:25: in <module>
    from opentelemetry import trace
E   ImportError: cannot import name 'trace' from 'opentelemetry' (unknown location)
```

## Fix
While is usually correct to catch `ModuleNotFoundError`, in case of opentelemetry this is a bit different since the actual "opentelemetry" package doesn't exist at all (we got opentelemetry-api and opentelemetry-sdk) and in that case Python (3.11 in my case) throws a more generic ImportError

### DIFF
--- a/elasticsearch/_otel.py
+++ b/elasticsearch/_otel.py
@@ -25,7 +25,7 @@ try:
     from opentelemetry import trace
 
     _tracer: trace.Tracer | None = trace.get_tracer("elasticsearch-api")
-except ModuleNotFoundError:
+except ImportError:
     _tracer = None
 
 from elastic_transport import OpenTelemetrySpan


pquentin (approved):
> While is usually correct to catch `ModuleNotFoundError`, in case of opentelemetry this is a bit different since the actual "opentelemetry" package doesn't exist at all (we got opentelemetry-api and opentelemetry-sdk) and in that case Python (3.11 in my case) throws a more generic ImportError

Actually, the _distribution packages_ opentelemetry-api and opentelemetry-sdk can be accessed with the _import package_ opentelemetry. See https://packaging.python.org/en/latest/discussions/distribution-package-vs-import-package/ to understand the difference. And indeed, if I run `pip install opentelemetry-api`, then I can run `from opentelemetry import trace` without error.

Most of our CI runs without `opentelemetry-api` installed, from Python 3.8 to 3.13. so catching `ModuleNotFoundError` is generally correct. What seems to happen is that you have an `opentelemetry` import package that does not define `trace` somehow. This is possible if you run `pip install opentelemetry-sdk` (which will also install `opentelemetry-api`) and then manually run `pip uninstall opentelemetry-api`. At this point, the import package `opentelemetry` still exists (and allows access to `opentelemtry.sdk`) but `opentelemetry.trace` no longer exists. I don't think the OpenTelemetry SDK can be used without the API, but this should not break the Elasticsearch client, so the pull request looks good to me. Thank you!

pquentin:
buildkite test this please

pquentin:
Thank you for your contribution, it is now released as part of 8.16.0: https://github.com/elastic/elasticsearch-py/releases/tag/v8.16.0

PR #2707, opened 2024-11-13
Build dists as part of CI

pquentin:
The build-dists.py step often fails during publishing, let's add it to CI.


miguelgrinberg (approved):

pquentin (commented):
Forgot to post my review before merging, posting it anyway.

PR #2717, opened 2024-12-05
Add custom sleep parameter to async_bulk and async_streaming_bulk

### LINKED ISSUE #2484: Helpers for `bulk` method such as `async_bulk` sleep in blocking manner, preventing graceful shutdown
We tried using `async_bulk` and `async_streaming_bulk` helpers to ingest data into Elasticsearch and they work great, but we've found out that they prevent our code from gracefully shutting down when CTRL+C is pressed.

Example code that sleeps: https://github.com/elastic/elasticsearch-py/blob/main/elasticsearch/_async/helpers.py#L241

It would be great to have a way to:

- Either define how the sleep happens by passing sleep function into the client
- Make Elasticsearch client internally cancel all sleeps when the client is closed

girolamo-giordano:
Add custom sleep parameter to async_bulk and async_streaming_bulk allowing users to define their custom sleep action on async_bulk and async_streaming_bulk calls
Closes #2484


pquentin:
Hello, and thank you! We'll also need to add this to the sync version and add tests in `test_helpers`.

cla/check

miguelgrinberg:
@pquentin Do you think we can close this one? I assume the purpose was to be able to use trio's sleep?

PR #2736, opened 2025-01-10
Add Elasticsearch-DSL code

miguelgrinberg:
This change adds the Elasticsearch-DSL code. Note that the documentation will be merged in a separate PR.

- The imported source tree has the same structure as in the `elasticsearch_dsl` package, except that its root is now `elasticsearch.dsl`. Users can replace the `_` with a `.` in all imports.
- I had to make minor changes to some of the DSL's test fixtures, because this client uses function scope for the client fixtures and DSL used session scope. I have switched all the DSL fixtures to function scope.
- I had to make the DSL client fixtures work with `https://` Elasticsearch endpoints, since that is what we are using in integration test runs.
- The unasync of the DSL classes remains in a separate `run-unasync-dsl.py` script. We can consider merging the two.
- The `generator.py` script that generates the query, aggregation and response source files was moved to this project as `generator-dsl.py`, along with its Jinja templates. We can consider moving this to the Python client generator project.


pyproject.toml
```diff
@@ -43,6 +43,8 @@ keywords = [
 dynamic = ["version"]
 dependencies = [
     "elastic-transport>=8.15.1,<9",
+    "python-dateutil",
```
pquentin, line 46: For what it's worth, this one makes me a bit sad, as it still pulls six to this day: https://inspector.pypi.io/project/python-dateutil/2.9.0.post0/packages/ec/57/56b9bcc3c9c6a792fcbaf139543cee77261f3651ca9da0c93f5c1221264b/python_dateutil-2.9.0.post0-py2.py3-none-any.whl/python_dateutil-2.9.0.post0.dist-info/METADATA

miguelgrinberg: Let me see if we can drop it.

pquentin (commented):
> * The unasync of the DSL classes remains in a separate run-unasync-dsl.py script. We can consider merging the two.
> * The generator.py script that generates the query, aggregation and response source files was moved to this project as generator-dsl.py, along with its Jinja templates. We can consider moving this to the Python client generator project.

Yes, please, as a follow-up.

> @pquentin The builds are currently not generating code coverage data, and I haven't been able to figure out why. And I get coverage data when I run the tests locally. I'll continue looking for this, but if you have any ideas why this is happening let me know. Other than this the PR is ready for a review.

I haven't looked in detail yet, but the current build has 13% coverage, and we don't use coverage for anything meaningful today. However it would be nice to fix this issue as I'm seeing 0% coverage here. (In the future, we can exclude generated code to make coverage more useful.)

test_elasticsearch/test_dsl/__init__.py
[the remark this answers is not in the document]
miguelgrinberg: I selected this name for consistency with the other folders at the level, correct.

elasticsearch/dsl/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+    MultiSearch,
+    Search,
+)
+from .update_by_query import AsyncUpdateByQuery, UpdateByQuery
+from .utils import AttrDict, AttrList, DslBase
+from .wrappers import Range
+
+VERSION = (8, 17, 1)
+__version__ = VERSION
+__versionstr__ = ".".join(map(str, VERSION))
```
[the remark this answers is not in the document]
miguelgrinberg: Removed.

noxfile.py
```diff
@@ -86,19 +89,33 @@ def lint(session):
...
     session.run("black", "--check", *SOURCE_FILES)
     session.run("python", "utils/run-unasync.py", "--check")
+    session.run("python", "utils/run-unasync-dsl.py", "--check")
     session.run("flake8", *SOURCE_FILES)
     session.run("python", "utils/license-headers.py", "check", *SOURCE_FILES)
 
     session.install(".[async,requests,orjson,pyarrow,vectorstore_mmr]", env=INSTALL_ENV)
 
     # Run mypy on the package and then the type examples separately for
     # the two different mypy use-cases, ourselves and our users.
```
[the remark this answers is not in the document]
miguelgrinberg: Instead I added a separate comment for the type checking of the DSL examples.

elasticsearch/dsl/document_base.py
```diff
@@ -0,0 +1,444 @@
...
+    TypeVar,
+    Union,
+    get_args,
+    overload,
+)
+
+try:
+    from types import UnionType
+except ImportError:
+    UnionType = None  # type: ignore
```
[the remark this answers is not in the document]
miguelgrinberg: This is great, I didn't know about this option!

examples/dsl/async/sparse_vectors.py
```diff
@@ -0,0 +1,198 @@
...
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+"""
+# Sparse vector database example
+
+Requirements:
+
+$ pip install nltk tqdm elasticsearch[async]
```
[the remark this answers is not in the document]
miguelgrinberg: Good call, added to all the examples.

miguelgrinberg:
@pquentin The last commit addresses your feedback. One thing I want to direct your attention to, is the handling of the user agent, which I only noticed now. The connection class used to add an `elasticsearch-dsl-py/$VERSION` user agent. I changed that to use `elasticsearch-py/$VERSION`, but I can't decide if this is a good idea or we should better leave it with the DSL user agent in case we want to see how many people use the DSL classes going forward. Let me know what your thoughts are on this.

elasticsearch/dsl/connections.py
```diff
@@ -116,16 +116,16 @@ def get_connection(self, alias: Union[str, _T] = "default") -> _T:
             raise KeyError(f"There is no connection with alias {alias!r}.")
 
     def _with_user_agent(self, conn: _T) -> _T:
-        from . import __versionstr__  # this is here to avoid circular imports
+        from elasticsearch import (
+            __versionstr__,  # this is here to avoid circular imports
```
pquentin, line 120: Technically, this is not a circular import, as `elasticsearch` does not import `elasticsearch.dsl` today.

elasticsearch/dsl/connections.py
```diff
@@ -116,16 +116,16 @@ def get_connection(self, alias: Union[str, _T] = "default") -> _T:
...
 
         # try to inject our user agent
         if hasattr(conn, "_headers"):
             is_frozen = conn._headers.frozen
             if is_frozen:
                 conn._headers = conn._headers.copy()
-            conn._headers.update(
-                {"user-agent": f"elasticsearch-dsl-py/{__versionstr__}"}
-            )
+            conn._headers.update({"user-agent": f"elasticsearch-py/{__versionstr__}"})
```
pquentin, line 128: Thanks for calling this out. I'd rather keep `elasticsearch-dsl-py` here. It's the same package, but a very different way of using it.

pquentin (commented):
Thanks! Sorry, I did not realize my mypy suggestion would bring so much changes.

elasticsearch/dsl/connections.py
miguelgrinberg: Correct, this can be moved to the top of the file.

elasticsearch/dsl/connections.py
miguelgrinberg: I reverted this to continue to use a dsl specific user agent.

PR #2740, opened 2025-01-14
Add sphinx.configuration to .readthedocs.yml

pquentin:
As mentioned in https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/. There was a brownout yesterday, which is how I noticed this.


pquentin:
Merged to 7.17 in https://github.com/elastic/elasticsearch-py/pull/2745.

PR #2756, opened 2025-01-21
Auto-generated code for main

pquentin:
* Remove Unfreeze index API


pquentin:
The docs failure will be fixed by the following pull request adding Markdown support.

PR #2757, opened 2025-01-21
Format endpoint descriptions as Markdown

pquentin:


miguelgrinberg (approved):
I'm approving, but I wonder how this will sit with people who read the docs on their IDEs. I think HTML in the docstrings should not be an end goal, but an intermediate step into something better to be figured out in the future.

PR #2761, opened 2025-01-23
Added DSL documentation to Guide

miguelgrinberg:
This adds the DSL documentation (not including the reference sections) to the documentation guide.

The contents are mostly translated from the original RST sources, without any major changes. There are improvements that can be made to the content as well, such as showing the new typed syntax in code blocks, which I'll address separately.


docs/guide/elasticsearch-dsl.asciidoc
```diff
@@ -0,0 +1,48 @@
+[[elasticsearch-dsl]]
+== Elasticsearch DSL
```
pquentin, line 2: I would call this the "Elasticsearch Python DSL" to clarify that it is not a general-purpose DSL. Now that it is documented next to the Elasticsearch docs, this seems important.

docs/guide/elasticsearch-dsl.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+Python objects: defining mappings, retrieving and saving documents,
+wrapping the document data in user-defined classes.
+
+To use the other Elasticsearch APIs (eg. cluster health) just use the
+regular client.
+
+include::dsl/configuration.asciidoc[]
+include::dsl/tutorials.asciidoc[]
+include::dsl/howto.asciidoc[]
+include::dsl/examples.asciidoc[]
```
pquentin, line 48: Can we also add links to those sections? On mobile the navigation is at the bottom of the page and easy to miss.

pquentin (approved):
Thanks! This literally looks great.

docs/guide/elasticsearch-dsl.asciidoc
miguelgrinberg: I agree that it would be good to add these links, but the way this is currently implemented is hacky. The links that appear in the body (as opposite to those that are in the sidebar) are implemented in an HTML file, not an asciidoc file. The only links that are currently in the body are links to top-level sections. I'm not sure there is a robust way of adding links to sub-sections, since these links are generated by asciidoc.

I'm not going to implement this right now, I think we should wait until the docs migrate and then we can see what can be done within the new docs system.

PR #2775, opened 2025-01-29
fix flaky integration test

miguelgrinberg:
The DSL vectors test is missing a refresh on the index, which causes the test to occasionally give unexpected results.


pquentin (approved):
Much better, thank you!

PR #2776, opened 2025-01-29
Reorganization of the Sphinx docs to only include reference pages

miguelgrinberg:
Reorganization of the Sphinx docs to only include reference pages.


pquentin (approved):
LGTM, but we should wait for HTTP redirects to be in place before merging IMO.

Including a redirect from /api.html to /index.html.

miguelgrinberg:
@pquentin Redirect table proposal:

source | target
-- | --
/en/latest/ | No change
/en/latest/quickstart.html | https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/getting-started-python.html
/en/latest/interactive.html | https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/examples.html
/en/latest/api.html | /en/latest/index.html
/en/latest/exceptions.html | No change
/en/latest/async.html | https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/async.html
/en/latest/helpers.html | https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/client-helpers.html

The `api.html`, `async.html` and `helpers.html` exist in both old and new designs. Would it make sense rename these pages in the redesigned docs, so that we can add the redirects above and also have pages we can link from now on?

pquentin:
> The `api.html`, `async.html` and `helpers.html` exist in both old and new designs. Would it make sense rename these pages in the redesigned docs, so that we can add the redirects above and also have pages we can link from now on?

Yes, good point.

> And aside from this, it is unfortunate but I think we are going to have to create redirects for each version that we publish, until we decide this isn't necessary anymore, because we want the <= 8.17 docs to be redirect free, meaning that we cannot do generic redirects across all versions.

That makes sense. The free plan we're using allows for 100 redirects, but I'm assuming we will only need redirects for latest, stable, 8.18, 8.19, 9.0 and 9.1, not much more than that?

miguelgrinberg:
All pages renamed so that there are no collisions with the redirects. I guess we can wait until the new docs system is in place to see if the redirects above need to be adjusted, and only merge this PR after we put those in place.

miguelgrinberg:
Rebased to incorporate minor changes made during the docs migration to markdown.

docs/sphinx/es_api.rst
pquentin, line 1: Why was this renamed? This is going to break existing links, I believe

docs/sphinx/index.rst
```diff
@@ -1,140 +1,22 @@
 Python Elasticsearch Client
 ===========================
```
pquentin, line 3: Should we link to the main docs?

docs/sphinx/es_api.rst
miguelgrinberg: This is to avoid collisions when we add the redirects. Look at the table above. Requests to `/en/latest/api.html` will be redirected to `/en/latest/index.html`, since now the whole site is reference documentation. This page is renamed because if not there would be no way to create a direct link to it, given that it collides with the redirect.

docs/sphinx/index.rst
miguelgrinberg: Yes, we should add links to the guide in all the reference pages where there is a reasonable page to link to. I'm looking into this.

miguelgrinberg:
@pquentin I have updated the [redirect table](https://github.com/elastic/elasticsearch-py/pull/2776#issuecomment-2627705283). It now has redirects for 8.18 and 9.0+ separately. I need to confirm that I've got the 9.0 links right, since these are not deployed yet and I'm building them based on previews and descriptions of what the URL system is going to be.

Can you confirm that this looks good?
There was an outstanding question you had regarding the rename of api.html to es-api.html. Do you now understand why I did this, or do we need to discuss this more?

pquentin:
Sounds good! We can set up the 9.0 redirects when the docs are live. Yes, es-api.html, makes sense.

miguelgrinberg:
Updated table of redirects:

source | 8.18 | 9.0+
-- | -- | --
/en/latest/ | No change | No change
/en/latest/quickstart.html | https://www.elastic.co/guide/en/elasticsearch/client/python-api/8.18/getting-started-python.html | https://www.elastic.co/docs/reference/elasticsearch/clients/python/getting-started
/en/latest/interactive.html | https://www.elastic.co/guide/en/elasticsearch/client/python-api/8.18/examples.html | https://www.elastic.co/docs/reference/elasticsearch/clients/python/examples
/en/latest/api.html | /en/latest/index.html | /en/latest/index.html
/en/latest/exceptions.html | No change | No change
/en/latest/async.html | https://www.elastic.co/guide/en/elasticsearch/client/python-api/8.18/async.html | https://www.elastic.co/docs/reference/elasticsearch/clients/python/async
/en/latest/helpers.html | https://www.elastic.co/guide/en/elasticsearch/client/python-api/8.18/client-helpers.html | https://www.elastic.co/docs/reference/elasticsearch/clients/python/client-helpers

PR #2798, opened 2025-02-14
Explain how to use sub clients in API docs

pquentin:
I have not found a way to not display the name of the class. But I removed the parameters and added an explanation text.

# Before
Screenshot 2025-02-14 at 18 17 23

# After
Screenshot 2025-02-14 at 18 17 03


miguelgrinberg (approved):
Big improvement.

miguelgrinberg (approved):

PR #2843, opened 2025-03-14
Add querying docs

pquentin:
It's not always obvious to go from an API to a function call, and this page contains the details.


docs/reference/querying.md
```diff
@@ -0,0 +1,108 @@
+# Sending queries to Elasticsearch
+
+The Python Elasticsearch client provides several ways to send queries to Elasticsearch. This document explains the details of how to construct and execute queries using the client. This document does not cover the DSL client.
```
miguelgrinberg, line 3: I have removed all references to a "DSL client" to avoid confusion, and instead used "DSL module", now that the clients are unified.

docs/reference/querying.md
```diff
@@ -0,0 +1,108 @@
...
+
+* Global namespace: `client.search(...)` and `client.index(...)`
+* Other namespaces:
+  * Indices: `client.indices.create(...)`
+  * ES|QL: `client.esql.query(...)`
+
+How can you figure out the namespace?
+
+* The [Elasticsearch API docs](https://www.elastic.co/docs/api/doc/elasticsearch/) can help, even though the tags it uses do not fully map to namespaces.
+* You can also use the client docs, by:
```
miguelgrinberg, line 25:
```suggestion
* You can also use the client documentation, by:
```

docs/reference/querying.md
```diff
@@ -0,0 +1,108 @@
...
+  * browing the [Elasticsearch API Reference](https://elasticsearch-py.readthedocs.io/en/stable/api.html) page, or
+  * search for your endpoint using [Read the Docs](https://elasticsearch-py.readthedocs.io/) search, which is powered by Elasticsearch!
```
miguelgrinberg, line 27:
```suggestion
  * searching for your endpoint using [Read the Docs](https://elasticsearch-py.readthedocs.io/) search, which is powered by Elasticsearch!
```

docs/reference/querying.md
```diff
@@ -0,0 +1,108 @@
...
+```python
+resp = client.indices.create(
+    index="logs",
+    aliases={"logs-alias": {}},
+    mappings={"name": {"type": "text"}},
+)
+print(resp)
+```
+
+(In a real query, don't use `...` but real mappings.)
```
miguelgrinberg, line 58: Was this a note for an earlier version of the above example? Because you have a "real" mapping now.

docs/reference/querying.md
```diff
@@ -0,0 +1,108 @@
...
+    body={
+        "aliases": {"logs-alias": {}},
+        "mappings": {"name": {"type": "text"}},
+        "missing_parameter": "foo",
+    }
+)
+print(resp)
+```
+
+In the event where an API is missing, you need to call the low-level `perform_request` function:
```
miguelgrinberg, line 86:
```suggestion
In the event where an API is missing, you need to use the low-level `perform_request` function:
```

docs/reference/querying.md
```diff
@@ -0,0 +1,108 @@
...
+    mappings={"name": {"type": "text"}},
+)
+print(resp)
+```
+
+(In a real query, don't use `...` but real mappings.)
+
+In this case, the client will send to Elasticsearch the following JSON body:
+
+```json
```
miguelgrinberg, line 62:
```suggestion
```console
```

Not 100% sure, but I think the new docs use `console` and not `json` for ES requests.

miguelgrinberg (approved):
Just a few minor comments and suggestions that you are welcome to ignore if you disagree.

PR #2857, opened 2025-03-24
Use class-based queries and type-hinted documents in DSL documentation examples

miguelgrinberg:
This change updates code blocks in DSL documentation to use the coding style introduced in recent releases based on query and aggregation classes and type hints for document definition. There is also some minor fixes/improvements, such as the addition of a few docstrings to methods of the `Search` class, and a new section on migrating from the `elasticsearch-dsl` package.


pquentin (approved):
Thanks! LGTM. Excellent work. Sorry for the reviewing delay.

PR #2858, opened 2025-03-24
Fix cyclic dependency between clients and sub clients

pquentin:
While the Python GC can handle those dependencies, it can cause latency spikes when .option() is used for each query, which causes many clients to be garbage collected, which can be slower in the presence of cycles.

https://github.com/python/cpython/blob/main/InternalDocs/garbage_collector.md

## objgraph backrefs, before and after

Namespaced sub clients reference back to the main client:

refs-main

No backreferences:

refs-cyclic-dependency-fixed

## 10 000 calls to client.options().info(), before and after

histogram-main

histogram-cyclic-fixed

Calling `options()` in a loop is 15% faster on average, and 2x faster for P100.

### DIFF
--- a/elasticsearch/_async/client/__init__.py
+++ b/elasticsearch/_async/client/__init__.py
@@ -20,10 +20,12 @@ import logging
 import typing as t
 
 from elastic_transport import (
+    ApiResponse,
     AsyncTransport,
     BaseNode,
     BinaryApiResponse,
     HeadApiResponse,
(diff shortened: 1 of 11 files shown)


gioboa (approved):
:+1: it looks nice

pquentin:
> 👍 it looks nice

Thank you for the review! So it looked nice with 14 changed lines, but it did not work. :)

PR #2860, opened 2025-03-27
Handle new aggregation range types Pythonically

miguelgrinberg:
This change makes the DSL code generator work with the new aggregation range types introduced in https://github.com/elastic/elasticsearch-specification/pull/2725. Basically this change makes the generator ignore these types and use a manually added one in wrappers.py, exactly like ranges are handled for queries.

In addition, I have changed the way the DSL generator determines which version of the schema to use, because the main branch reports 8.17.2 as current version. With this change we are using the git branch instead of the version reported in the code. And some schema changes that are unrelated to aggregation ranges are also included.

### DIFF
--- a/elasticsearch/dsl/aggs.py
+++ b/elasticsearch/dsl/aggs.py
@@ -35,6 +35,7 @@ from typing import (
 
 from elastic_transport.client_utils import DEFAULT
 
+from . import wrappers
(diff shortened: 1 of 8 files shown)


pquentin (approved):
Thanks! LGTM.

PR #2861, opened 2025-03-27
[docs] Add param notes to DSL search.delete documentation, fix broken links

lunalucadou:
This MR was previously opened against the DSL project but I was told I should submit it here instead: elastic/elasticsearch-dsl-py#1970

Earlier today I was trying to pass `wait_for_completion=True` to `Search.delete` but found `Search.delete` does not support `**kwargs`.

I was going to just add support for `**kwargs` (it would be a fairly trivial change), but after seeing elastic/elasticsearch-dsl-py#1115 and elastic/elasticsearch-dsl-py#1395, it seems the preferred way is to call `Search.params` first. However, this is not called out in the documentation for `Search.delete`, so I instead opted to add it to the documentation.

While I was at it, I noticed `Search.scan` and `FacetedSearch.params` had broken links so I fixed those as well.

### DIFF
--- a/docs/reference/dsl_how_to_guides.md
+++ b/docs/reference/dsl_how_to_guides.md
@@ -84,6 +84,15 @@ s = Search(index='i').query("match", title="python")
 response = s.delete()
 ```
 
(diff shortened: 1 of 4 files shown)


miguelgrinberg (approved):

miguelgrinberg:
Thanks @lunalucadou !

PR #2871, opened 2025-04-01
Only wipe user content between tests

miguelgrinberg:
This reduces the buildkite build time.


pquentin:
Sounds good! Should we clean up `is_xpack`?

PR #2897, opened 2025-04-10
use internal aggregation range class for date aggregation ranges

miguelgrinberg:
After https://github.com/elastic/elasticsearch-specification/pull/4243 I realized that the DSL code generator wasn't properly assigning the type to date aggregation range attributes.

Note that the changes in this diff aside from the change in `dsl-generator.py` are unrelated. These are just minor changes in the spec that are unrelated to this PR.


pquentin (approved):
Thanks! LGTM.

PR #2911, opened 2025-04-14
Fix release notes URL

pquentin:
Closes #2888


CHANGELOG.md
```diff
@@ -1 +1 @@
-See: https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/release-notes.html
+See: https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/release-notes.html
```
miguelgrinberg, line 1: We'll need to change this again to point to the new docs, right?

miguelgrinberg (approved):

CHANGELOG.md
pquentin: Yes, at least for main and 9.0. Which makes me wonder if we should point to a local file in the repo instead of an URL

PR #2912, opened 2025-04-14
[8.x] Remove table of contents in release notes

pquentin:
It's displayed on the right already and hides real content below the browser fold.


miguelgrinberg (approved):
Looks much better! 👍

PR #2925, opened 2025-04-15
8.18 client documentation updates

miguelgrinberg:
These are client documentation updates that were included in the migration to md, but for some reason were lost in the asciidoc sources


pquentin (approved):
Thanks! LGTM.

Should this be opened against 8.x and then backported to 8.18?

PR #2952, opened 2025-04-29
Intro paragraph for the readthedocs API reference

miguelgrinberg:


pquentin (commented):
Is this what we had in index.rst in the past? Not sure it works well today with the DSL module in the client.

Additionally, as discussed offline, [since clients are under "Reference" in elastic.co docs](https://www.elastic.co/docs/reference/elasticsearch-clients/), maybe we should choose another name, like "API documentation" instead or "reference documentation"?

miguelgrinberg:
It isn't exactly what we had in the past, but it is close. I have removed references to "low-level client" for example. Just checked and the original wording including the reference to low-level client made it to the new docs on elastic.co as well, so I'll think about something more up to date to say as introduction and update both sides.

pquentin (approved):
Thanks! LGTM.

PR #2986, opened 2025-06-19
update install_requires_urlib3

kashewnuts:


pquentin:
Hello! I'm assuming you opened that pull request by mistake, but for the record elasticsearch-py 7.13.0 was released in May 2021 and isn't maintained anymore. For this reason, even though urllib3 1.26.x is affected by one moderate vulnerability, we don't have plans to relax the urllib3 constraints and release a new 7.x version. You can use elasticsearch-py 8.x or 9.x instead which support urllib3 2.x, but you need to make sure to upgrade Elasticsearch itself first.

You can see more details about our EOL policy here: https://www.elastic.co/support/eol. Thank you!

PR #2987, opened 2025-06-20
Allow sending format with esql.async_query_get

jetn-ttc:
The format of received async queries is determined by the async_query_get call, not the async_query (unless it returns immediately). This change allows one to send the format specifier whenever useful. Note that sending format=arrow with a to small completion timeout results in an empty response and no way to get the ID. A more useful strategy is therefore: async_query with no format (or JSON) to get the ID, and only then async_query_get with the received ID and format=arrow.

### DIFF
--- a/elasticsearch/_sync/client/esql.py
+++ b/elasticsearch/_sync/client/esql.py
@@ -248,6 +248,14 @@ class EsqlClient(NamespacedClient):
         drop_null_columns: t.Optional[bool] = None,
         error_trace: t.Optional[bool] = None,
         filter_path: t.Optional[t.Union[str, t.Sequence[str]]] = None,
+        format: t.Optional[
+            t.Union[
+                str,
+                t.Literal[
+                    "arrow", "cbor", "csv", "json", "smile", "tsv", "txt", "yaml"
+                ],
+            ]
+        ] = None,
         human: t.Optional[bool] = None,
         keep_alive: t.Optional[t.Union[str, t.Literal[-1], t.Literal[0]]] = None,
         pretty: t.Optional[bool] = None,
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Hi @jetn-ttc, thanks for your PR!

The file that you edited is auto-generated directly from the Elasticsearch specification. It appears that we do not have `format` listed as a valid query parameter for the `GET` method of this endpoint, just for `POST`.

I'm not super familiar with the ES|QL endpoints myself, so we need to find out if the specification is wrong, and in that case the fix needs to be made [there](https://github.com/elastic/elasticsearch-specification/blob/07be12f2085cb3f81c3638a1a87a86e8a11532b3/specification/esql/async_query_get/AsyncQueryGetRequest.ts#L42), so that then all our clients, documentation and other projects that derive from the spec are updated accordingly.

jetn-ttc:
I see ! It does seem somewhat generated now you say it :-D

I can see from my own modifications that it works (on 8.18.1 at least). 

Testing with curl and the same workflow shows the ID you need is actually also sent in the header, so the server might be ok it's only the python code that doesn't take that into consideration. I'll perhaps just open an issue on the repo you linked.

miguelgrinberg:
Yes, I have checked internally and this does seem to be an omission in the specification, so I'll add `format` there. Feel free to write an issue on that repo if you like to track this work over there. Thanks!

miguelgrinberg:
I'm closing this issue. The fix in the Elasticsearch specification has been applied. All future patch releases of this client in the 9.0.x and 8.18.x lines plus the upcoming 9.1 and 8.19 minor releases will include this as a fix.

PR #2990, opened 2025-06-23
Upgrade `urllib3` on 7.17

4c0n:
I'd like to keep using elasticsearch 7 for a bit longer, however there are some security issues with `urllib3` < 2.5.0.
This PR aims to upgrade `urllib3` to at least version 2.5.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ packages = [
 ]
 
 install_requires = [
-    "urllib3>=1.21.1, <2",
+    "urllib3>=2.5.0, <3",
     "certifi",
 ]
 tests_require = [


pquentin:
As mentioned in https://github.com/elastic/elasticsearch-py/pull/2986#issuecomment-2987562925, elasticsearch-py 7.x has [reached end of maintenance in April 2025](https://www.elastic.co/support/eol#prior-versions). For this reason, even though urllib3 1.26.x is affected by one moderate vulnerability, we don't have plans to relax the urllib3 constraints and release a new 7.x version (especially as it's [more complicated than that](https://github.com/elastic/elastic-transport-python/pull/121)). You can use elasticsearch-py 8.x or 9.x instead which support urllib3 2.x, but you need to make sure to upgrade Elasticsearch itself first. That said, elasticsearch-py itself is not affected as it does not use the PoolManager interface, directly or indirectly. (Source: I also happen to be an urllib3 maintainer.)

4c0n:
@pquentin Thank you for your response, good to hear the vulnerability has no effect on this package.

PR #2997, opened 2025-07-01
ES|QL query builder

miguelgrinberg:
This PR adds a Python ES|QL query builder.

- [x] Implement all ES|QL commands
- [x] Add docstrings
- [ ] Integration with DSL module --> Will be implemented later in a separate PR.
- [x] API documentation
- [x] Client documentation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/docs/reference/esql-query-builder.md
+++ b/docs/reference/esql-query-builder.md
@@ -1,0 +1,210 @@
+# ES|QL Query Builder
+
+The ES|QL Query Builder allows you to construct ES|QL queries using Python syntax. Consider the following example:
+
+```python
+>>> from elasticsearch.esql import ESQL
+>>> query = (
+    ESQL.from_("employees")
+    .sort("emp_no")
+    .keep("first_name", "last_name", "height")
+    .eval(height_feet="height * 3.281", height_cm="height * 100")
+    .limit(3)
+)
+```
+
+You can then see the assembled ES|QL query by printing the resulting query object:
+
+```python
+>>> query
+FROM employees
+| SORT emp_no
+| KEEP first_name, last_name, height
+| EVAL height_feet = height * 3.281, height_cm = height * 100
+| LIMIT 3
+```
+
+To execute this query, you can cast it to a string and pass the string to the `client.esql.query()` endpoint:
+
+```python
+>>> from elasticsearch import Elasticsearch
+>>> client = Elasticsearch(hosts=[os.environ['ELASTICSEARCH_URL']])
+>>> response = client.esql.query(query=str(query))
+```
+
+The response body contains a `columns` attribute with the list of columns included in the results, and a `values` attribute with the list of results for the query, each given as a list of column values. Here is a possible response body returned by the example query given above:
+
+```python
+>>> from pprint import pprint
+>>> pprint(response.body)
+{'columns': [{'name': 'first_name', 'type': 'text'},
+             {'name': 'last_name', 'type': 'text'},
+             {'name': 'height', 'type': 'double'},
+             {'name': 'height_feet', 'type': 'double'},
+             {'name': 'height_cm', 'type': 'double'}],
+ 'is_partial': False,
+ 'took': 11,
+ 'values': [['Adrian',
+             'Wells',
+             2.424,
+             7.953144,
+             242.4],
+            ['Aaron',
+             'Gonzalez',
+             1.584,
+             5.1971,
+             158.4],
+            ['Miranda',
+             'Kramer',
+             1.55,
+             5.08555,
+             155]]}
+```
+
+## Creating an ES|QL query
+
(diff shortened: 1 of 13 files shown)


miguelgrinberg:
@pquentin this PR is ready for review. There are currently two remaining items:

- The 3.8 integration tests continue to fail, unrelated to this change. There are 10 tests that fail, always the same ones, since the 15th. I haven't been able to figure out the cause, but all the failures seem to be related to indices that for some reason weren't wiped and consequently have more documents than expected or fail to be created again. In one of the 10 failures there is a permission failure when attempting to wipe an index, but the user is `elastic`, so I'm not quite follow why. For now I'm ignoring this problem.
- I'm not going to write a full integration with the DSL module for the 9.1 release, as I don't want to rush this solution, which is not a trivial one. It is possible to do a lightweight integration by passing index or field names (as you can see in some of the tests), but what's missing is the ability to run an ES|QL query directly from a document class, which I'm going to design with more time.

docs/reference/esql-query-builder.md
```diff
@@ -0,0 +1,210 @@
...
+| EVAL height_feet = height * 3.281, height_cm = height * 100
+| LIMIT 3
+```
+
+To execute this query, you can cast it to a string and pass the string to the `client.esql.query()` endpoint:
+
+```python
+>>> from elasticsearch import Elasticsearch
+>>> client = Elasticsearch(hosts=[os.environ['ELASTICSEARCH_URL']])
+>>> response = client.esql.query(query=str(query))
```
miguelgrinberg, line 36: I'm not sure what your thoughts are about being able to pass the query object directly here and have the conversion to string in the generated method. Maybe something we can think about for the future as an improvement.

pquentin: That's a good idea, and I don't think this can be a problem in the future

docs/reference/esql-query-builder.md
```diff
@@ -0,0 +1,210 @@
...
+            ['Aaron',
+             'Gonzalez',
+             1.584,
+             5.1971,
+             158.4],
+            ['Miranda',
+             'Kramer',
+             1.55,
+             5.08555,
+             155]]}
```
pquentin: I get a much nicer output, should we use that instead?

```
{'columns': [{'name': 'first_name', 'type': 'text'},
             {'name': 'last_name', 'type': 'text'},
             {'name': 'height', 'type': 'double'},
             {'name': 'height_feet', 'type': 'double'},
             {'name': 'height_cm', 'type': 'double'}],
 'is_partial': False,
 'took': 11,
 'values': [['Adrian', 'Wells', 2.424, 7.953144, 242.4],
            ['Aaron', 'Gonzalez', 1.584, 5.1971, 158.4],
            ['Miranda', 'Kramer', 1.55, 5.08555, 155]]}
```

docs/reference/esql-query-builder.md
```diff
@@ -0,0 +1,210 @@
...
+from elasticsearch.esql import ESQL
+
+query1 = ESQL.from_("employees")
+query2 = ESQL.from_("<logs-{now/d}>")
+query3 = ESQL.from_("employees-00001", "other-employees-*")
+query4 = ESQL.from_("cluster_one:employees-00001", "cluster_two:other-employees-*")
+query5 = ESQL.from_("employees").metadata("_id")
+```
+
+Note how in the last example the optional `METADATA` clause of the `FROM` command is added as a chained method.
```
pquentin, line 85: Should we link to https://www.elastic.co/docs/reference/query-languages/esql/esql-metadata-fields or show the string representation? This applies to all other examples. It could be as simple as:

```python
# FROM employees METADATA _id
query5 = ESQL.from_("employees").metadata("_id")
```

docs/reference/esql-query-builder.md
```diff
@@ -0,0 +1,210 @@
...
+    .eval(height_feet=E("height") * 3.281, height_cm=E("height") * 100)
+)
+```
+
+Here the `E()` helper function is used as a wrapper to the column name that initiates an ES|QL expression. The `E()` function transforms the given column into an ES|QL expression that can be modified with Python operators.
+
+Here is a second example, which uses a conditional expression in the `WHERE` command:
+
+```python
+from elasticsearch.esql import ESQL, E
```
pquentin:
```suggestion
from elasticsearch.esql import ESQL
```

docs/reference/esql-query-builder.md
miguelgrinberg: I can add the ES|QL code for each example, sure.

docs/reference/esql-query-builder.md
miguelgrinberg: I have modified the Python client generator to allow this, so once we get that merged we can eliminate the requirement of converting the queries to strings.

PR #2999, opened 2025-07-03
Update opentelemetry conventions

miguelgrinberg:
Updates to OpenTelemetry conventions for span attributes.

Note that these changes will be released with 9.1+. We are not planning to backport to 8.19 to avoid a backward incompatibility with older transport versions.

### DIFF
--- a/elasticsearch/__init__.py
+++ b/elasticsearch/__init__.py
@@ -28,7 +28,7 @@ from ._version import __versionstr__
 
 # Ensure that a compatible version of elastic-transport is installed.
 _version_groups = tuple(int(x) for x in re.search(r"^(\d+)\.(\d+)\.(\d+)", _elastic_transport_version).groups())  # type: ignore[union-attr]
-if _version_groups < (8, 0, 0) or _version_groups > (9, 0, 0):
+if _version_groups < (9, 1, 0) or _version_groups > (10, 0, 0):
     raise ImportError(
         "An incompatible version of elastic-transport is installed. Must be between "
         "v8.0.0 and v9.0.0. Install the correct version with the following command: "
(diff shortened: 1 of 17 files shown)


elasticsearch/__init__.py
```diff
@@ -28,7 +28,7 @@
 
 # Ensure that a compatible version of elastic-transport is installed.
 _version_groups = tuple(int(x) for x in re.search(r"^(\d+)\.(\d+)\.(\d+)", _elastic_transport_version).groups())  # type: ignore[union-attr]
-if _version_groups < (8, 0, 0) or _version_groups > (9, 0, 0):
+if _version_groups < (9, 1, 0) or _version_groups > (10, 0, 0):
     raise ImportError(
         "An incompatible version of elastic-transport is installed. Must be between "
         "v8.0.0 and v9.0.0. Install the correct version with the following command: "
```
pquentin: Can you please update the text too?

elasticsearch/dsl/field.py
```diff
@@ -4081,6 +4081,9 @@ def __init__(
 class SparseVector(Field):
     """
     :arg store:
+    :arg index_options: Additional index options for the sparse vector
+        field that controls the token pruning behavior of the sparse
+        vector field.
```
pquentin: I'd love if we could get this is in before the Python 3.8 drop or those OTel changes

elasticsearch/__init__.py
miguelgrinberg: Now sure how I missed this... Sorry.

elasticsearch/dsl/field.py
miguelgrinberg: The DSL files should not have been in this PR. Removed.

PR #3001, opened 2025-07-24
Drop support for Python 3.8

miguelgrinberg:


elasticsearch/dsl/field.py
```diff
@@ -4099,6 +4102,9 @@ def __init__(
         self,
         *args: Any,
         store: Union[bool, "DefaultType"] = DEFAULT,
+        index_options: Union[
+            "types.SparseVectorIndexOptions", Dict[str, Any], "DefaultType"
+        ] = DEFAULT,
```
pquentin: Should we removes those unrelated changes?

pquentin (commented):
`git grep 3\\.8` gives other things to change:

```
docs/reference/getting-started.md:* [Python](https://www.python.org/) 3.8 or newer
examples/fastapi-apm/dockerfiles/Dockerfile.app:FROM python:3.8
examples/fastapi-apm/dockerfiles/Dockerfile.ping:FROM python:3.8
utils/dsl-generator.py:                    # When dropping Python 3.8, use `removesuffix("Function")` instead
```

Additionally, running `pyupgrade --py38-plus **/*.py` produces a few more interesting changes.

elasticsearch/dsl/field.py
miguelgrinberg: Oops, yes, this should be done separately.

PR #3041, opened 2025-08-12
DSL: preserve the `skip_empty` setting in `to_dict()` recursive serializations

pquentin:
I wanted to fix https://github.com/elastic/elasticsearch-dsl-py/issues/1577, but realized I wasn't able to reproduce it.

### DIFF
--- a/utils/templates/field.py.tpl
+++ b/utils/templates/field.py.tpl
@@ -119,8 +119,15 @@ class Field(DslBase):
     def __getitem__(self, subfield: str) -> "Field":
         return cast(Field, self._params.get("fields", {})[subfield])
 
-    def _serialize(self, data: Any) -> Any:
+    def _serialize(self, data: Any, skip_empty: bool) -> Any:
         return data
+
+    def _safe_serialize(self, data: Any, skip_empty: bool) -> Any:
+        try:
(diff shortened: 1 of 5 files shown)


miguelgrinberg:
@pquentin added a 2nd commit with a small adjustment to your attempt that does reproduce the issue. The 3rd commit contains the fix, which involves the plumbing to pass `skip_empty` through recursive serializations.

utils/templates/field.py.tpl
```diff
@@ -119,9 +119,16 @@ class Field(DslBase):
...
+    def _serialize(self, data: Any, skip_empty: bool) -> Any:
         return data
 
+    def _safe_serialize(self, data: Any, skip_empty: bool) -> Any:
+        try:
+            return self._serialize(data, skip_empty)
+        except TypeError:
+            # older method signature, without skip_empty
+            return self._serialize(data)  # type: ignore[call-arg]
+
```
pquentin, line 131: Why do we need this? Is this for third-party libraries that want to subclass `Field`?

miguelgrinberg: This is needed for people who created their own custom fields and use the old signature in their `_serialize()` methods.

miguelgrinberg (approved):

PR #3043, opened 2025-08-15
add rerank() command to the ES|QL query builder

miguelgrinberg:


elasticsearch/esql/esql.py
```diff
@@ -1046,6 +1095,65 @@ def _render_internal(self) -> str:
...
+            return (
+                f"RERANK {self._format_id(column)} = {json.dumps(query)} "
+                f"ON {', '.join([self._format_id(field) for field in self._fields])} "
+                f"WITH {json.dumps(with_)}"
+            )
+        else:
+            return (
+                f"RERANK {json.dumps(self._query[0])} "
+                f"ON {', '.join([self._format_id(field) for field in self._fields])} "
+                f"WITH {json.dumps(with_)}"
```
pquentin: nit: the code duplication makes me slightly nervous, given this code is not generated

miguelgrinberg: I thought it best to keep the templates for each version of the command separate so that you can see what the command looks like more easily, but sure, it does add some repetition. I'll see if I can come up with a better way to structure this and other similar commands.

PR #3048, opened 2025-08-22
ES|QL query builder integration with the DSL module

miguelgrinberg:


elasticsearch/dsl/_async/document.py
```diff
@@ -520,3 +523,83 @@ async def __anext__(self) -> Dict[str, Any]:
...
+        # nested fields, which ES|QL does not return, causing an error. When passing
+        # `ignore_missing_fields=True` the list will be generated with
+        # `for_esql=True`, so the error will not occur, but the documents will
+        # not have any nested objects in them.
+        doc_fields = set(cls._get_field_names(for_esql=ignore_missing_fields))
+        if not ignore_missing_fields and not doc_fields.issubset(set(query_columns)):
+            raise ValueError(
+                f"Not all fields of {cls.__name__} were returned by the query. "
+                "Make sure your document does not use Nested fields, which are "
+                "currently not supported in ES|QL."
```
pquentin: Should we mention `ignore_missing_fields` here?

elasticsearch/dsl/_async/document.py
```diff
@@ -520,3 +523,83 @@ async def __anext__(self) -> Dict[str, Any]:
...
+            columns returned by the query in the second element.
+        :arg ignore_missing_fields: if ``False`` (the default), all the fields of
+            the document must be present in the query, or else an exception is
+            raised. Set to ``True`` to allow missing fields, which will result in
+            partially initialized document objects.
+        :arg using: connection alias to use, defaults to ``'default'``
+        :arg kwargs: additional options for the ``client.esql.query()`` function.
+        """
+        es = cls._get_connection(using)
+        response = await es.esql.query(query=str(query), **kwargs)
```
pquentin, line 553: Do we need `str(query)` here?

elasticsearch/dsl/_async/document.py
```diff
@@ -520,3 +523,83 @@ async def __anext__(self) -> Dict[str, Any]:
...
+        query_columns = [col["name"] for col in response.body.get("columns", [])]
+
+        # Here we get the list of columns defined in the document, which are the
+        # columns that we will take from each result to assemble the document
+        # object.
+        # When `for_esql=False` is passed below by default, the list will include
+        # nested fields, which ES|QL does not return, causing an error. When passing
+        # `ignore_missing_fields=True` the list will be generated with
+        # `for_esql=True`, so the error will not occur, but the documents will
+        # not have any nested objects in them.
```
pquentin: Do I read it correctly that `ignore_missing_fields` is the only way to work with nested objects? If yes, should this be mentioned in the docs? Or should we link to the ES|QL limitations page?

In any case, maybe capitalizing Nested could help clarify that this is a very specific thing in Elasticsearch.

elasticsearch/dsl/_async/document.py
```diff
@@ -43,6 +44,8 @@
 if TYPE_CHECKING:
     from elasticsearch import AsyncElasticsearch
 
+    from ...esql.esql import ESQLBase
```
pquentin: nit: should this be an absolute import?

elasticsearch/dsl/document_base.py
```diff
@@ -602,3 +603,44 @@ def to_dict(self, include_meta: bool = False, skip_empty: bool = True) -> Dict[s
 
         meta["_source"] = d
         return meta
+
+    @classmethod
+    def _get_field_names(
+        cls, for_esql: bool = False, nested_class: Optional[type[InnerDoc]] = None
```
pquentin, line 609: Should this be named `ignore_missing_fields` too? After all, ES|QL may support nested objects at some point.

elasticsearch/esql/esql.py
```diff
@@ -97,7 +97,10 @@ class ESQLBase(ABC):
     commands, used to build ES|QL queries.
     """
 
-    def __init__(self, parent: Optional["ESQLBase"] = None):
+    def __init__(
+        self,
+        parent: Optional["ESQLBase"] = None,
+    ):
```
pquentin: nit: no-op change

elasticsearch/esql/functions.py
```diff
@@ -649,7 +649,7 @@ def min_over_time(field: ExpressionType) -> InstrumentedExpression:
 
 
 def multi_match(
-    query: ExpressionType, fields: ExpressionType, options: ExpressionType = None
+    query: ExpressionType, *fields: ExpressionType, options: ExpressionType = None
```
pquentin: I agree that this is nicer, but I think it's a breaking change?

elasticsearch/dsl/_async/document.py
miguelgrinberg: Only for as long as we don't refresh the generated code. I left it there because of that. Once we release patches on all versions we can remove the cast.

elasticsearch/dsl/_async/document.py
miguelgrinberg: I wouldn't say that it is the only way to work with nested fields. I think a more appropriate way to put it is that ES|QL does not support nested fields, so it should not be used with indexes that have them. You can write regular ES|QL queries to work with the non-nested fields in the index, but using the DSL-based query builder with nested fields should not be recommended, because the ignore flag will cause your documents to be incomplete.

elasticsearch/dsl/document_base.py
miguelgrinberg: Hmm, not sure I understand this. The ignore flag makes sense when you have some results, and you determine that some fields are missing in those results. Here we are asking for the list of fields, so the concept of "missing" does not exist. Besides, the `for_esql` flag can be adapted to include nested fields in the future, when ES|QL adds support for them.

elasticsearch/esql/esql.py
miguelgrinberg: I blame black for this. I'll revert it and reformat to see if black decides to keep it.

elasticsearch/esql/functions.py
miguelgrinberg: Will commit this change separately, so that we have a log entry to include in the release notes.

PR #3064, opened 2025-09-08
Add `flush_after_seconds` option to `streaming_bulk()`

### LINKED ISSUE #3051: Support timeout-based flushing in `async_streaming_bulk`
Currently, `async_streaming_bulk` only flushes requests based on `chunk_size` and `max_chunk_bytes`. This works well for high-throughput use cases, but in scenarios with low or bursty traffic, requests may remain unflushed indefinitely until a batch fills.

It would be helpful to support a **timeout fallback** mechanism that flushes after a configurable duration, ensuring progress is made even when throughput is low. 

**Proposed approach / discussion points:** 

- Introduce a configurable timeout parameter for `async_streaming_bulk`.
- Modify `ActionChunker` to operate as an async iterator, allowing explicit calls to `next()`.
- Use `asyncio.wait_for` (or `anyio.fail_after`) to implement time-based flushing. 

> Initial design approach is from discussing with @pquentin in slack

miguelgrinberg:
This change adds a timeout controlled flush to the streaming bulk helper.

Fixes #3051 

Tasks:

- [x] flush sentinel object
- [x] flush unit tests
- [x] flush integration tests
- [x] timeout async support
- [x] timeout sync support
- [x] timeout integration tests
- [x] docstring updates

### DIFF
--- a/elasticsearch/_async/helpers.py
+++ b/elasticsearch/_async/helpers.py
@@ -235,6 +235,9 @@ async def async_streaming_bulk(
     :arg actions: iterable or async iterable containing the actions to be executed
     :arg chunk_size: number of docs in one chunk sent to es (default: 500)
     :arg max_chunk_bytes: the maximum size of the request in bytes (default: 100MB)
+    :arg flush_after_seconds: time in seconds after which a chunk is written even
+        if hasn't reached `chunk_size` or `max_chunk_bytes`. Set to 0 to not use a
+        timeout-based flush. (default: 0)
     :arg raise_on_error: raise ``BulkIndexError`` containing errors (as `.errors`)
         from the execution of the last chunk when some occur. By default we raise.
     :arg raise_on_exception: if ``False`` then don't propagate exceptions from
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -451,6 +451,9 @@ def streaming_bulk(
     :arg actions: iterable containing the actions to be executed
     :arg chunk_size: number of docs in one chunk sent to es (default: 500)
     :arg max_chunk_bytes: the maximum size of the request in bytes (default: 100MB)
+    :arg flush_after_seconds: time in seconds after which a chunk is written even
+        if hasn't reached `chunk_size` or `max_chunk_bytes`. Set to 0 to not use a
+        timeout-based flush. (default: 0)
     :arg raise_on_error: raise ``BulkIndexError`` containing errors (as `.errors`)
         from the execution of the last chunk when some occur. By default we raise.
     :arg raise_on_exception: if ``False`` then don't propagate exceptions from
@@ -636,6 +639,9 @@ def parallel_bulk(
     :arg thread_count: size of the threadpool to use for the bulk requests
     :arg chunk_size: number of docs in one chunk sent to es (default: 500)
     :arg max_chunk_bytes: the maximum size of the request in bytes (default: 100MB)
+    :arg flush_after_seconds: time in seconds after which a chunk is written even
+        if hasn't reached `chunk_size` or `max_chunk_bytes`. Set to 0 to not use a
+        timeout-based flush. (default: 0)
     :arg raise_on_error: raise ``BulkIndexError`` containing errors (as `.errors`)
         from the execution of the last chunk when some occur. By default we raise.
     :arg raise_on_exception: if ``False`` then don't propagate exceptions from


elasticsearch/helpers/actions.py
```diff
@@ -44,13 +46,28 @@
...
 _TYPE_BULK_ACTION_HEADER_AND_BODY = Tuple[
     _TYPE_BULK_ACTION_HEADER, _TYPE_BULK_ACTION_BODY
 ]
 
+_TYPE_BULK_ACTION_WITH_META = Union[bytes, str, Dict[str, Any], BulkMeta]
+_TYPE_BULK_ACTION_HEADER_WITH_META = Union[Dict[str, Any], BulkMeta]
+_TYPE_BULK_ACTION_HEADER_WITH_META_AND_BODY = Union[
+    Tuple[_TYPE_BULK_ACTION_HEADER, _TYPE_BULK_ACTION_BODY],
+    Tuple[BulkMeta, Any],
+]
```
miguelgrinberg, line 68: The intention of the new "with meta" types above is to differentiate the types that are used in the `expand_actions()` callback (which is public and customizable by users) from the ones we use internally, which may carry metadata that we use for flushing and timeouts. This prevents us from creating backwards incompatible typing changes that may cause failures for users who define their own callback.

elasticsearch/helpers/actions.py
```diff
@@ -152,39 +171,42 @@ def feed(
         ]
     ]:
         ret = None
-        raw_action = action
-        raw_data = data
```
miguelgrinberg, line 156: I couldn't find any purpose for these two raw variables, so I removed them.

elasticsearch/helpers/actions.py
```diff
@@ -231,10 +254,48 @@ def _chunk_actions(
     chunker = _ActionChunker(
         chunk_size=chunk_size, max_chunk_bytes=max_chunk_bytes, serializer=serializer
     )
-    for action, data in actions:
-        ret = chunker.feed(action, data)
-        if ret:
-            yield ret
+
+    if not flush_after_seconds:
+        for action, data in actions:
```
miguelgrinberg, line 258: If there is no timeout given, then everything works as before.

elasticsearch/helpers/actions.py
```diff
@@ -231,10 +254,48 @@ def _chunk_actions(
...
+            ret = chunker.feed(action, data)
+            if ret:
+                yield ret
+    else:
```
miguelgrinberg, line 262: If there is a timeout, then we put the for-loop that iterates over the data elements in a separate `get_items()` task that puts the items in a queue as they are received. And here on this side we wait on the queue with a timeout, so that we can insert a flush when data does not come for a while.

elasticsearch/helpers/actions.py
```diff
@@ -425,6 +487,13 @@ def streaming_bulk(
 
         serializer = client.transport.serializers.get_serializer("application/json")
 
+        def expand_action_with_meta(
```
miguelgrinberg, line 485: This is another unfortunate complication that I have added to not introduce backwards incompatible changes in the `expand_action()` callback.

elasticsearch/helpers/actions.py
pquentin: Is this because callback types are [contravariant](https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types)? So, a more generic callback is fine, but not a more specific one? And "with meta" types are more specific?

elasticsearch/helpers/actions.py
pquentin: This was introduced in https://github.com/elastic/elasticsearch-py/commit/accb19ab905cc68bcc1690e19f3cd3ad18e7f670, when `raw_action` was a dictionary and `action` a JSON dump of this dictionary. This is indeed no longer needed.

elasticsearch/_async/helpers.py
```diff
@@ -76,10 +79,44 @@ async def _chunk_actions(
...
+        async def get_items() -> None:
+            try:
+                async for item in actions:
+                    await item_queue.put(item)
+            except Exception:
+                await item_queue.put((BulkMeta.done, None))
+                raise
+            await item_queue.put((BulkMeta.done, None))
+
+        item_getter_job = asyncio.create_task(get_items())
```
pquentin: Please use a task group instead to make sure the task will never be forgotten: https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup

miguelgrinberg: Task groups are fairly new, 3.11 according to the docs.

elasticsearch/helpers/actions.py
miguelgrinberg: The specific issue is not related to types, it's just that the types reflect what arguments the callback may receive. Basically the callback is supposed to receive standard actions, we cannot send flush command because if the user is running a customized callback then their code will break when it receives an unexpected item that is not an action.

So instead of invoking the callback as before, now there is an `expand_actions_with_meta()` function that consumes the metadata and only passes actions to the real callback. The types merely reflect this change.

test_elasticsearch/test_server/test_helpers.py
```diff
@@ -75,6 +76,47 @@ def test_bulk_all_documents_get_inserted(sync_client):
...
+
+    timestamps = []
+    for ok, item in helpers.streaming_bulk(
+        sync_client, sync_gen(), index="test-index", refresh=True
+    ):
+        assert ok
+        timestamps.append(time.time())
+
+    # make sure there is a pause between the writing of the 2nd and 3rd items
+    assert timestamps[2] - timestamps[1] > (timestamps[1] - timestamps[0]) * 2
```
miguelgrinberg, line 95: This assert may look strange. The reason I'm validating flushing by looking at timestamp differences is to attempt to not use any fixed timings and avoid a flaky test.

elasticsearch/compat.py
```diff
@@ -76,9 +79,48 @@ def warn_stacklevel() -> int:
...
 
 
+@contextmanager
+def safe_thread(
+    target: Callable[..., Any], *args: Any, **kwargs: Any
+) -> Iterator[Thread]:
+    """Run a thread within a context manager block.
+
+    The thread is automatically joined when the block ends. If the thread raised
+    an exception, it is raised in the caller's context.
```
pquentin, line 89: Out of curiosity, do you have an example traceback that we get with this?

miguelgrinberg: Example test script:

```python
import time
from elasticsearch.compat import safe_thread

def my_thread():
    print("my_thread started")
    1/0

def main():
    print("main started")
    with safe_thread(target=my_thread):
        time.sleep(1)

main()
```

Output:

```
❯ python threadtest.py
main started
my_thread started
  File "/Users/miguelgrinberg/Documents/dev/elasticsearch-py/threadtest.py", line 14, in <module>
    main()
    ~~~~^^
  File "/Users/miguelgrinberg/Documents/dev/elasticsearch-py/threadtest.py", line 10, in main
    with safe_thread(target=my_thread):
         ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/Users/miguelgrinberg/.local/share/mise/installs/python/3.14.0/lib/python3.14/contextlib.py", line 148, in __exit__
    next(self.gen)
    ~~~~^^^^^^^^^^
  File "/Users/miguelgrinberg/Documents/dev/elasticsearch-py/elasticsearch/compat.py", line 105, in safe_thread
    raise captured_exception
  File "/Users/miguelgrinberg/Documents/dev/elasticsearch-py/elasticsearch/compat.py", line 95, in run
    target(*args, **kwargs)
    ~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/miguelgrinberg/Documents/dev/elasticsearch-py/s.py", line 6, in my_thread
    1/0
    ~^~
ZeroDivisionError: division by zero
```

Issue #3075, opened 2025-09-15
Object._serialize() missing 1 required positional argument: 'skip_empty'


selimt:
This started happening with the `9.1.1` release:

```
src/catalog/search/documents.py:59: in save
    return super().save(**kwargs)
/usr/local/lib/python3.10/site-packages/elasticsearch/dsl/_sync/document.py:440: in save
    body=self.to_dict(skip_empty=skip_empty),
src/catalog/search/documents.py:79: in to_dict
    doc = super().to_dict(skip_empty=skip_empty)
/usr/local/lib/python3.10/site-packages/elasticsearch/dsl/document_base.py:593: in to_dict
    d = super().to_dict(skip_empty=skip_empty)
/usr/local/lib/python3.10/site-packages/elasticsearch/dsl/utils.py:606: in to_dict
    v = f.serialize(v, skip_empty=skip_empty)
/usr/local/lib/python3.10/site-packages/elasticsearch/dsl/field.py:152: in serialize
    return self._safe_serialize(data, skip_empty)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = Object(), data = {'data': {'abc': 'def'}}, skip_empty = True
    def _safe_serialize(self, data: Any, skip_empty: bool) -> Any:
        try:
            return self._serialize(data, skip_empty)
        except TypeError:
            # older method signature, without skip_empty
>           return self._serialize(data)  # type: ignore[call-arg]
E           TypeError: Object._serialize() missing 1 required positional argument: 'skip_empty'
/usr/local/lib/python3.10/site-packages/elasticsearch/dsl/field.py:130: TypeError
```

Everything works fine with `9.1.0`

Thanks.

miguelgrinberg:
Hi, I need more details to fully understand this issue. Could you share the entire stack trace chain to see if that provides more insight?

Also, are you using custom `Field` instances in your application, or is this happening with the standard Elasticsearch fields we provide with this package?

selimt:
Here is the full stderr from the unittest:

```
==================================== ERRORS ====================================
__________ ERROR at setup of TestNamespaces.test_namespace_taggables ___________
self = Object(), data = {'data': {'abc': 'def'}}, skip_empty = True
    def _safe_serialize(self, data: Any, skip_empty: bool) -> Any:
        try:
[...]

PR #3076, opened 2025-09-17
Add ABC base class to AttrDict et AttrList

isra17:
Indirectly fix https://github.com/elastic/elasticsearch-py/issues/3075

Not sure if there's a good reason to avoid using these base class, it __might__ break some code that do `isinstance(x, Mapping)` and expect to not match these, but I can't think of a legit scenario.

On the otherhand, this have been a cause of multiple headache, especially when you print a response object and it look like a dict, when you look at the code and it look like a dict.


miguelgrinberg:
I would consider this change backwards incompatible, so aside from its pros or its cons which are debatable, I'm thinking it carries some risk. The fact that you had to add additional checks inside this library to separate "real" collections from our imitation ones suggests this could also happen in application code, and that's not something we want.

leameow:
> I would consider this change backwards incompatible, so aside from its pros or its cons which are debatable, I'm thinking it carries some risk. The fact that you had to add additional checks inside this library to separate "real" collections from our imitation ones suggests this could also happen in application code, and that's not something we want.

Fair.

For what it's worth, there are really only two changes:

```
        if isinstance(data, AttrDict) or not isinstance(data, collections.abc.Mapping):
            data = data.to_dict()
```

We only really need `isinstance(data, AttrDict)` but I kept the other part for backward compatibility...

```
        if isinstance(data, InnerDoc):
            return data.to_dict(skip_empty=skip_empty)
        elif isinstance(data, AttrDict):
            return data.to_dict()
        # somebody assigned raw dict to the field, we should tolerate that
        elif isinstance(data, collections.abc.Mapping):
            return data
```

This change is needed to fix https://github.com/elastic/elasticsearch-py/issues/3075, no matter if we actually make `AttrDict` a `Mapping`.

I stand my point that `AttrDict` not being a `Mapping` is pain anyone using this lib will eventually stumble onto, but I understand if you prefer to keep this issue rather than potentially break some external uses.

The fact that `isinstance(x, (AttrDict, Mapping))` in this codebase and in our codebase shows how annoying it can get... (And many bugs occured because of some check missing it).

I guess I can open a PR that at least fix the original issue on it's own.

miguelgrinberg:
@isra17 My current understanding of this problem is that it occurs only on applications that set inner document values using `AttriDict` instances, and at the same time request that validation of these values be skipped when saving the document.

- If you set your inner doc values from `InnerDoc` instances or `dict` instances, then everything works well with or without validation
- If you always validate the documents that you save, then you can use `AttrDict` as values for your inner documents and again everything works.

Your application just has to set these fields to `AttrDict` and at the same time request the validation to be skipped for the error to occur. Is this characterization correct, or am I missing a more serious use case in which this crash appears?

leameow:
> Your application just has to set these fields to `AttrDict` and at the same time request the validation to be skipped for the error to occur. Is this characterization correct, or am I missing a more serious use case in which this crash appears?

This get triggered when you call `to_dict()` on the document.

leameow:
Here's a stacktrace from our app:

```
 <redacted>.py:526: in  <redacted>
    data=item.to_dict(),
../pants_cache/named_caches/pex_root/venvs/1/s/65db4e66/venv/lib/python3.13/site-packages/elasticsearch/dsl/document_base.py:593: in to_dict
    d = super().to_dict(skip_empty=skip_empty)
../pants_cache/named_caches/pex_root/venvs/1/s/65db4e66/venv/lib/python3.13/site-packages/elasticsearch/dsl/utils.py:606: in to_dict
    v = f.serialize(v, skip_empty=skip_empty)
../pants_cache/named_caches/pex_root/venvs/1/s/65db4e66/venv/lib/python3.13/site-packages/elasticsearch/dsl/field.py:145: in serialize
    return list(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Object()
data = <redacted>
skip_empty = True

    def _safe_serialize(self, data: Any, skip_empty: bool) -> Any:
        try:
            return self._serialize(data, skip_empty)
        except TypeError:
            # older method signature, without skip_empty
>           return self._serialize(data)  # type: ignore[call-arg]
E           TypeError: Object._serialize() missing 1 required positional argument: 'skip_empty'

../pants_cache/named_caches/pex_root/venvs/1/s/65db4e66/venv/lib/python3.13/site-packages/elasticsearch/dsl/field.py:130: TypeError
```

miguelgrinberg:
@isra17 and to help with my understanding of your use case, can you clarify why are you using AttrDict values for these fields instead of working with `InnerDoc` instances? Is there a benefit in doing that?

My expectation would be that you would work with regular dictionaries if an `InnerDoc` class isn't defined. Where does the need to use `AttrDict` come from?

leameow:
These `AttrDict` comes from other documents. Something like:

```
crash_doc["inner"] = existing_doc._source["inner"]
```

miguelgrinberg:
Closing. We'll implement a simpler approach that is less risky.

PR #3078, opened 2025-09-18
Add most recently added fields as top-level exports

miguelgrinberg:
Indirect discovery while working on #3075 is that the most recent batch of `Field` subclasses were not included as top-level exports. This PR adds them.


miguelgrinberg:
> Should this be code generated?

It should, ideally. But there is a mix of manual and generated classes in the import list, so the file cannot be fully generated. I'll see if I can add a unit test that ensures that all the generated fields can be imported, which should be easier to do.

pquentin (approved):

PR #3080, opened 2025-09-18
Support values of inner docs given as AttrDict instances

### LINKED ISSUE #3075: Object._serialize() missing 1 required positional argument: 'skip_empty'

miguelgrinberg:
Fixes #3075.

It appears that some users pass `AttrDict` instances as values for fields of type `Object` or `Nested`. If the document is validated (the default when `save()` is called) then these are handled well, but some users also call `save(validate=False)` or just `to_dict()`, in which case the AttrDicts cause the error described in #3075. This change adds logic that tolerates AttrDict values, and treats them like regular dicts (which are already tolerated).

### DIFF
--- a/test_elasticsearch/test_dsl/test_field.py
+++ b/test_elasticsearch/test_dsl/test_field.py
@@ -24,7 +24,14 @@ import pytest
 from dateutil import tz
 
 from elasticsearch import dsl
-from elasticsearch.dsl import InnerDoc, Range, ValidationException, field
+from elasticsearch.dsl import (
+    AttrDict,
+    AttrList,
+    InnerDoc,
+    Range,
+    ValidationException,
+    field,
+)
 
 
 def test_date_range_deserialization() -> None:
@@ -233,6 +240,33 @@ def test_object_constructor() -> None:
 
     with pytest.raises(ValidationException):
         field.Object(doc_class=Inner, dynamic=False)
+
+
+def test_dynamic_object() -> None:
+    f = field.Object(dynamic=True)
+    assert f.deserialize({"a": "b"}).to_dict() == {"a": "b"}
+    assert f.deserialize(AttrDict({"a": "b"})).to_dict() == {"a": "b"}
+    assert f.serialize({"a": "b"}) == {"a": "b"}
+    assert f.serialize(AttrDict({"a": "b"})) == {"a": "b"}
+
+
+def test_dynamic_nested() -> None:
+    f = field.Nested(dynamic=True)
+    assert f.deserialize([{"a": "b"}, {"c": "d"}]) == [{"a": "b"}, {"c": "d"}]
+    assert f.deserialize([AttrDict({"a": "b"}), {"c": "d"}]) == [
+        {"a": "b"},
+        {"c": "d"},
+    ]
+    assert f.deserialize(AttrList([AttrDict({"a": "b"}), {"c": "d"}])) == [
+        {"a": "b"},
+        {"c": "d"},
+    ]
+    assert f.serialize([{"a": "b"}, {"c": "d"}]) == [{"a": "b"}, {"c": "d"}]
+    assert f.serialize([AttrDict({"a": "b"}), {"c": "d"}]) == [{"a": "b"}, {"c": "d"}]
+    assert f.serialize(AttrList([AttrDict({"a": "b"}), {"c": "d"}])) == [
+        {"a": "b"},
+        {"c": "d"},
+    ]
 
 
 def test_all_fields_exported() -> None:


elasticsearch/dsl/field.py
```diff
@@ -572,7 +572,11 @@ def _serialize(
         if isinstance(data, collections.abc.Mapping):
             return data
 
-        return data.to_dict(skip_empty=skip_empty)
+        try:
```
leameow, line 575: Why not just `isinstance(data, AttrDict)`?

miguelgrinberg: Because this would also include support for other types that are not AttrDict. And because it avoids evaluating the conditional for every field that is saved, so it should perform a bit better.

leameow: > Because this would also include support for other types that are not AttrDict.

Which would probably lead to more crash anyway?

>And because it avoids evaluating the conditional for every field that is saved, so it should perform a bit better.

Handling an exception will cost many time what an `isinstance` cost. But does a few nano seconds performance of this even matter here 🤔 Just seems an unneeded "complex" structure flow to hide a very simple intent.

miguelgrinberg: The exception would be handled only for the fields that have a wrong type. The conditional would have to be evaluated for every single field.

leameow: And to my point, does it even matter in this case? Elasticsearch-dsl serialization/deserialization code is clearly not designed with performance in mind, so I don't really see why this one `isinstance` is seen as a potential issue at the cost of readability. But that's just my 2 cents. (and my last comment on this :) )

miguelgrinberg:
@isra17 Thanks for the discussion and the insights you provided.

I'm going to do some additional testing, and if you don't mind, I'm also going to bring the unit test that you created in your PR, which is different than mine.

As far as the exception vs. no exception method for preventing this error it's really not a big deal, but I have seen other methods that do an instance check like you suggested, so maybe that wins in terms of consistency. But I'm not sure if that is enough. I don't know yet if we need a similar fix for `AttrList` instances in `Nested` fields, which inherit from `Object` and use the same serialize/deserialize logic.

leameow:
> @isra17 Thanks for the discussion and the insights you provided.

Thanks for looking at the issue! I understand this is a bit of a niche error, so it's nice to see it fixed :)

> I'm going to do some additional testing, and if you don't mind, I'm also going to bring the unit test that you created in your PR, which is different than mine.

Of course!

> I don't know yet if we need a similar fix for `AttrList` instances in `Nested` fields, which inherit from `Object` and use the same serialize/deserialize logic.

For what it's worth, I think if `AttrList` would have been an issue we would have seen it in the past since the `[de]serialize` method never dealt with list? Isn't any fields can be a list anyway? My guess is that `to_dict` on the document take care of unrolling list? Seems like `AttrDict.to_dict` does, but I can't find how `DocumentBase` does it 🤔

leameow:
Ha, `Field.serialize` deals with the lists.

elasticsearch/dsl/field.py
leameow: Actually, fwiw, I suppose your method is consistent with [`Field._safe_serialize`](https://github.com/elastic/elasticsearch-py/pull/3041/files#diff-df682b3a1072c61688b2c15ddb0e7fbb9e64f8096f875d41e08cd2821a1da48bR125)

PR #3086, opened 2025-09-22
Support `Annotated` typing hint

miguelgrinberg:
This change adds support for the `Annotated` syntax when defining document fields in the DSL module. Examples:

```python
class TypedDocAnnotated(AsyncDocument):
    ip: Annotated[Optional[str], field.Ip()]
    k1: Annotated[str, field.Keyword(required=True)]
    k2: Annotated[M[str], field.Keyword()]
    k3: Annotated[str, mapped_field(field.Keyword(), default="foo")]
```


examples/quotes/backend/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[project]
+name = "quotes"
+version = "0.1"
+dependencies = [
+    "elasticsearch[async]>=8,<9",
```
pquentin:
```suggestion
    "elasticsearch[async]>=9,<10",
```

Are you planning to backport this?

examples/quotes/backend/quotes.py
```diff
@@ -0,0 +1,178 @@
...
+@app.get("/api/quotes/{id}")
+async def get_quote(id: str) -> Quote:
+    doc = None
+    try:
+        doc = await Quote._doc.get(id)
+    except NotFoundError:
+        pass
+    if not doc:
+        raise HTTPException(status_code=404, detail="Item not found")
+    return Quote.from_doc(doc)
```
pquentin, line 63: Isn't this code the same?

```suggestion
    try:
        doc = await Quote._doc.get(id)
        return Quote.from_doc(doc)
    except NotFoundError:
        raise HTTPException(status_code=404, detail="Item not found")
```

This also applies to `get_quote` and `delete_quote`.

examples/quotes/backend/quotes.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+@app.post("/api/quotes", status_code=201)
+async def create_quote(req: Quote) -> Quote:
+    embed_quotes([req])
+    doc = req.to_doc()
+    doc.meta.id = ""
```
pquentin, line 70: Why is this needed?

examples/quotes/backend/quotes.py
```diff
@@ -0,0 +1,178 @@
...
+    await doc.delete(refresh=True)
+
+
+@app.post('/api/search')
+async def search_quotes(req: SearchRequest) -> SearchResponse:
+    s = Quote._doc.search()
+    if req.query == '':
+        s = s.query(dsl.query.MatchAll())
+    elif req.knn:
+        s = s.query(dsl.query.Knn(field=Quote._doc.embedding, query_vector=model.encode(req.query).tolist()))
```
pquentin: nit: Splitting this into two lines could help with readability:

```suggestion
        query_vector = model.encode(req.query).tolist()
        s = s.query(dsl.query.Knn(field=Quote._doc.embedding, query_vector=query_vector))
```

examples/quotes/README.md
```diff
@@ -0,0 +1,107 @@
+# Quotes
+Quotes database example, which demonstrates the Elasticsearch integration with
+Pydantic models. This example features a React frontend and a FastAPI back end.
+
+![Quotes app screenshot](screenshot.png)
```
pquentin, line 5: Great quotes! Do you maybe have an example that does rely on embeddings?

examples/quotes/public/.keep
pquentin, line 1: What is the purpose of this folder? This looks like a create-react-app artifact that is not actually needed here.

examples/quotes/backend/pyproject.toml
miguelgrinberg: Good catch! I created the project on which this example is based for a talk I gave last year, so the versions are out of date. I will refresh.

examples/quotes/README.md
miguelgrinberg: I don't understand what you mean here. This example uses embeddings, both on their own and combined with BM25 search.

examples/quotes/public/.keep
miguelgrinberg: Correct. This project was initially created with `create-react-app`, and I now refreshed it and moved it to Vite. The public folder doesn't serve any purpose in this example, but I'm guessing Vite expects it to be there, since it serves static files off of it.

examples/quotes/README.md
pquentin: I mean, the results would have been the same with BM25 only

pquentin: So maybe there is something else than "dogs and books" that we can use. That's just a nitpick.

miguelgrinberg: Ah, I see, you are talking about the screenshot. In fact, "dogs and books" does not match the Groucho Marx quote at the top when using BM25, because that quote has "dog" and "book" in it in singular. I have to test this, but maybe using "canine" instead of dog makes the example more clear and returns the same results.

examples/quotes/backend/quotes.py
miguelgrinberg: No, this is close, but not the same. My version doesn't raise any typing errors, yours does. The problem is that our `AsyncDocument.get()` method can return `None` if the document is not found and `AsyncElasticsearch.get()` is not configured to raise the not found exception. That makes all uses of the `get()` method awkward, because you always have to account for the possible `None`.

examples/quotes/backend/quotes.py
miguelgrinberg: This is for security. We do not want the client to decide what the id should be, so by blanking it before calling `save()` we make sure if they provided one it is discarded.

examples/quotes/README.md
miguelgrinberg: The "books and pets" search term returns a few good ones, including the one from Groucho Marx at the top of the list.

PR #3088, opened 2025-09-23
Standardize on pytest.fixture

pquentin:
There's no need to use `pytest_asyncio.fixture` and always using the pytest
prefix is clearer.


pquentin:
Turns out I can't separate this change, this only really works with `pytest.mark.anyo`.

PR #3089, opened 2025-09-23
Support Trio with httpx

pquentin:


elasticsearch/_async/helpers.py
```diff
@@ -53,6 +55,15 @@
 T = TypeVar("T")
 
 
+async def _sleep(seconds: float) -> None:
+    if sniffio.current_async_library() == "trio":
```
miguelgrinberg, line 63: Would it make sense to ask the transport what the current library is, since it stores it? Not super important, but it feels unnecessary to constantly run the detection logic in sniffio.

pyproject.toml
```diff
@@ -41,11 +41,17 @@ keywords = [
 ]
 dynamic = ["version"]
 dependencies = [
-    "elastic-transport>=9.1.0,<10",
+    # TODO revert before merging/releasing
+    "elastic-transport @ git+https://github.com/pquentin/elastic-transport-python.git@trio-support",
```
miguelgrinberg: Adding a comment here just to remind you that need this to be edited. (not sure if your `TODO` comment triggers some alert or linter...)

test_elasticsearch/test_async/test_server/conftest.py
```diff
@@ -16,27 +16,27 @@
...
 
-
-@pytest_asyncio.fixture(scope="function")
+@pytest.fixture(scope="function")
 async def async_client_factory(elasticsearch_url):
-
-    if not hasattr(elasticsearch, "AsyncElasticsearch"):
-        pytest.skip("test requires 'AsyncElasticsearch' and aiohttp to be installed")
-
+    print("async!", elasticsearch_url)
```
miguelgrinberg: debug print?

test_elasticsearch/test_async/test_server/test_rest_api_spec.py
```diff
@@ -39,6 +38,8 @@
 )
 from ...utils import parse_version
 
+# We're not using `pytest.mark.anyio` here because it would run the test suite twice,
+# which does not work as it does not fully clean up after itself.
```
miguelgrinberg, line 42: Probably something that we'll want to address at some point.

test_elasticsearch/test_async/test_transport.py
```diff
@@ -580,8 +582,10 @@ async def test_sniffing_disabled_on_elastic_cloud(self, kwargs):
             == "Sniffing should not be enabled when connecting to Elastic Cloud"
         )
 
-    async def test_sniff_on_start_close_unlocks_async_calls(self):
-        event_loop = asyncio.get_running_loop()
+    async def test_sniff_on_start_close_unlocks_async_calls(self, anyio_backend):
+        if anyio_backend == "trio":
+            pytest.skip("trio does not support sniffing")
```
miguelgrinberg: I think it isn't necessary to skip here, or even use the `anyio` fixture and functions, since this whole class is marked with `asyncio` and not `anyio`.

utils/run-unasync-dsl.py
```diff
@@ -129,7 +128,7 @@ def main(check=False):
                 [
                     "sed",
                     "-i.bak",
-                    "s/pytest.mark.asyncio/pytest.mark.sync/",
+                    "s/pytest.mark.anyio/pytest.mark.sync/",
```
miguelgrinberg: I think this should handle both `anyio` and `asyncio` markers, just in case. Unless we ignore the asyncio marker on purpose just so that the build fails when we mistakenly use it instead of anyio?

elasticsearch/_async/helpers.py
pquentin: Good point!

* sniffio uses caching per-thread, and is actually quite fast.
* Also given this is for sleeping, it does not seem critical.
* But more importantly, exposing the async library from the transport seems annoying! Unless you think using `_async_library` directly is fine

test_elasticsearch/test_async/test_server/test_rest_api_spec.py
pquentin: I spent a few hours on this before giving up. Agreed it would be nice.

utils/run-unasync-dsl.py
pquentin: Done in [#3089 (07e4292)](https://github.com/elastic/elasticsearch-py/pull/3089/commits/07e4292dd815b1849927ae694c1cfafffe28ab40)

test_elasticsearch/test_async/test_transport.py
pquentin: Thanks! Done in [#3089 (b8aaa1f)](https://github.com/elastic/elasticsearch-py/pull/3089/commits/b8aaa1f496776de22b8b43c72ac2a971b0f46b60)

test_elasticsearch/test_async/test_server/conftest.py
pquentin: Thanks! Done in [#3089 (e3cbd6c)](https://github.com/elastic/elasticsearch-py/pull/3089/commits/e3cbd6c402b30947415d8ac8a507f97a879bed33)

pyproject.toml
pquentin: Thanks! There's no alert or linter, no.

pyproject.toml
```diff
@@ -77,6 +80,7 @@ dev = [
     "mapbox-vector-tile",
     "jinja2",
     "tqdm",
+    "trio",
```
agronholm, line 83: This would've been better as `anyio[trio]` to guarantee a trio version compatible with AnyIO.

test_elasticsearch/test_dsl/_async/test_document.py
```diff
@@ -582,21 +582,21 @@ def test_meta_fields_can_be_set_directly_in_init() -> None:
     assert md.meta.id is p
 
 
-@pytest.mark.asyncio
+@pytest.mark.anyio
```
agronholm, line 585: Could these have been removed, and `anyio_mode = "auto"` added to `[tool.pytest.ini_options]` in `pyproject.toml`?

pquentin: No, because some tests are only compatible with asyncio, in which case we stick with `pytest.mark.asyncio`. More generally, I like the explicit markers

pyproject.toml
pquentin: Right, however, this is a dev dependency only

test_elasticsearch/test_dsl/_async/test_document.py
agronholm: AnyIO itself has many such tests too, and there we force the use of the asyncio back-end:
`@pytest.mark.parametrize("anyio_backend", "asyncio")`

pyproject.toml
agronholm: Yes, and if Trio makes a backwards incompatible change in a new release, the CI might break as a result. The `trio` extra in AnyIO exists to shield against that.

test_elasticsearch/test_dsl/_async/test_document.py
pquentin: Good to know, thanks! Not going to rewrite everything now but I'll keep it in mind.

pyproject.toml
pquentin: Done in https://github.com/elastic/elasticsearch-py/pull/3139

PR #3096, opened 2025-10-07
Various fixes in the ES|QL query builder docstrings

miguelgrinberg:


pquentin:
Should we backport this to 9.2?

PR #3098, opened 2025-10-09
Update the compatiblity mode section of the docs

miguelgrinberg:
Clarify how compatibility mode works in the 8.x and newer clients.


miguelgrinberg:
@pquentin Okay, I have removed the section in the connecting page, and have merged the content into the main Compatibility section in the index page. WDYT?

PR #3125, opened 2025-10-23
Always return dense vectors in searches

miguelgrinberg:
In 9.2, dense vectors aren't returned unless they are requested explicitly via the `_source` query parameter. To avoid this breaking change from affecting the DSL module, all `Search` and `AsyncSearch` instances created via the `Document.search()` and `AsyncDocument.search()` methods are initialized with a `_source` property that includes the entire list of fields in the document.

There are also a few minor adjustments to the tests for other apparent changes in 9.2.

### DIFF
--- a/elasticsearch/dsl/search_base.py
+++ b/elasticsearch/dsl/search_base.py
@@ -721,12 +721,13 @@ class SearchBase(Request[_R]):
 
         def ensure_strings(
             fields: Union[
+                bool,
                 str,
                 "InstrumentedField",
                 List[Union[str, "InstrumentedField"]],
                 Dict[str, List[Union[str, "InstrumentedField"]]],
             ],
-        ) -> Union[str, List[str], Dict[str, List[str]]]:
+        ) -> Union[bool, str, List[str], Dict[str, List[str]]]:
             if isinstance(fields, dict):
                 return {k: ensure_strings(v) for k, v in fields.items()}
             elif isinstance(fields, bool):


elasticsearch/dsl/_async/document.py
```diff
@@ -126,9 +126,10 @@ def search(
         Create an :class:`~elasticsearch.dsl.Search` instance that will search
         over this ``Document``.
         """
-        return AsyncSearch(
+        s = AsyncSearch[Self](
             using=cls._get_using(using), index=cls._default_index(index), doc_type=[cls]
         )
+        return s.source(cls._get_field_names())  # type: ignore[arg-type]
```
miguelgrinberg: had to add a type ignore here. I believe mypy is wrong. The type is declared as a union, with one of the elements being `List[str | InstrumentedAttr]`. The argument I'm passing is `List[str]`.

test_elasticsearch/test_dsl/test_integration/_async/test_document.py
```diff
@@ -864,7 +864,7 @@ async def test_dense_vector(
     class Doc(AsyncDocument):
         float_vector: List[float] = mapped_field(DenseVector())
         byte_vector: List[int] = mapped_field(DenseVector(element_type="byte"))
-        bit_vector: str = mapped_field(DenseVector(element_type="bit"))
+        bit_vector: List[int] = mapped_field(DenseVector(element_type="bit"))
```
miguelgrinberg, line 867: This must be another backwards incompatible change. I could set the bit vector using a hex string, but it is returned as an array of signed ints when I ask for it, so I changed it to be `list[int]` all the way.

test_elasticsearch/test_dsl/test_integration/_async/test_document.py
```diff
@@ -873,13 +873,15 @@ class Index:
...
+        float_vector=[1.0, 1.2, 2.3],
+        byte_vector=[12, 23, 34, 45],
+        bit_vector=[18, -43, -112],
     )
     await doc.save(refresh=True)
 
     docs = await Doc.search().execute()
     assert len(docs) == 1
-    assert docs[0].float_vector == doc.float_vector
+    assert [round(v, 1) for v in docs[0].float_vector] == doc.float_vector
```
miguelgrinberg, line 884: And this is yet another minor issue. The floats are not exactly the values I set, probably due to the quantization that we now apply by default. So I need to round before running the comparison.

test_elasticsearch/test_dsl/test_integration/_async/test_search.py
```diff
@@ -135,7 +135,7 @@ async def test_inner_hits_are_serialized_to_dict(
 
 @pytest.mark.anyio
 async def test_scan_respects_doc_types(async_data_client: AsyncElasticsearch) -> None:
-    repos = [repo async for repo in Repository.search().scan()]
+    repos = [repo async for repo in Repository.search().source("organization").scan()]
```
miguelgrinberg: This test failed because it checks for an attribute that is not explicitly defined in the document class. To solve the dense vector field problem I now ask for the known fields explicitly, so this needs to be requested explicitly as well.

elasticsearch/dsl/_async/document.py
tselepakis: It seems to be a mypy issue indeed. Other checkers are able to perform the substitution properly. Another option instead of ignoring the error, would be to cast with the expected type (`List[Union[str, "InstrumentedField"]`) as you do in other places. But I am not sure if this verbosity will make things clear regarding the underlying issue which is the mypy limitation.

test_elasticsearch/test_dsl/test_integration/_sync/test_search.py
```diff
@@ -135,7 +135,7 @@ def test_inner_hits_are_serialized_to_dict(
 
 @pytest.mark.sync
 def test_scan_respects_doc_types(data_client: Elasticsearch) -> None:
-    repos = [repo for repo in Repository.search().scan()]
+    repos = [repo for repo in Repository.search().source("organization").scan()]
```
pquentin: Why is that needed?

elasticsearch/dsl/_async/document.py
```diff
@@ -126,9 +126,10 @@ def search(
         Create an :class:`~elasticsearch.dsl.Search` instance that will search
         over this ``Document``.
         """
-        return AsyncSearch(
+        s = AsyncSearch[Self](
             using=cls._get_using(using), index=cls._default_index(index), doc_type=[cls]
         )
+        return s.source(exclude_vectors=False)
```
pquentin, line 132: Is this ignored on 9.1 and earlier? While we only promise forward compatibility, it's common for users to upgrade their client before Elasticsearch, since it's easier to do. If it's causing an error, this should likely be highlighted in the release notes.

elasticsearch/dsl/search_base.py
```diff
@@ -729,6 +729,9 @@ def ensure_strings(
         ) -> Union[str, List[str], Dict[str, List[str]]]:
             if isinstance(fields, dict):
                 return {k: ensure_strings(v) for k, v in fields.items()}
+            elif isinstance(fields, bool):
```
pquentin, line 733: Should we add `bool` to the type of fields? (Surprised mypy doesn't complain.)

test_elasticsearch/test_dsl/test_integration/_sync/test_search.py
miguelgrinberg: Ah, with the `exclude_vectors=False` option this is not needed. I will remove.

elasticsearch/dsl/_async/document.py
miguelgrinberg: This flag is supported also in 9.1 and 8.19, but not in 9.0 and in any releases before 8.19. We have decided that we can live with this risk. The alternative would be to include an explicit list of fields here, but that breaks another usage of the library that is currently supported, which is to access fields that are not explicitly declared  in the Document class.

PR #3139, opened 2025-10-31
Test a version of Trio that is compatible with AnyIO

pquentin:
As suggested by @agronholm in https://github.com/elastic/elasticsearch-py/pull/3089#discussion_r2472585676

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,7 @@ dev = [
     "mapbox-vector-tile",
     "jinja2",
     "tqdm",
-    "trio",
+    "anyio[trio]",
     "mypy",
     "pyright",
     "types-python-dateutil",


pyproject.toml
agronholm, line 47: Shouldn't we also add a lower bound for AnyIO? Some folks are still using 3.x from what I can see from the download charts.

agronholm:
What also caught my eye was the explicit use of `sniffio`. Why is there a helper function `_sleep()` instead of just using `anyio.sleep()`? And the conftest.py use of sniffio could just be replaced by the use of the `anyio_backend_name` fixture. That would let you drop the explicit dependency on `sniffio`.

pquentin:
Good point! Initially `anyio` wasn't a runtime dependency, but then we needed a queue, and I did not clean up that part.

agronholm:
Yup, makes perfect sense.

PR #3151, opened 2025-11-08
Fix missing assignment

ava0-0sec:
This change corrects identifier escaping in `elasticsearch/esql/esql.py` by properly assigning the result of `str.replace()` back to the local variable before wrapping the identifier in backticks. Previously the `replace()` call produced a new string that was discarded; this leaves inner backticks unescaped. The patch ensures internal backticks are doubled so the rendered ES|QL identifier is quoted correctly.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -45,15 +45,15 @@ spec:
           branch: 'main'
           cronline: '0 10 * * *'
           message: 'Daily run for main branch'
-        Daily 9.0:
-          branch: '9.0'
+        Daily 9.2:
+          branch: '9.2'
           cronline: '0 10 * * *'
-          message: 'Daily run for 9.0 branch'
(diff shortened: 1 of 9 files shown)


miguelgrinberg:
@celeste0-0sec Thanks for the bug fix! Can you please sign the CLA so that I can apply this change?

miguelgrinberg:
@celeste0-0sec I'm unable to merge this until you sign the CLA. If you already signed, the issue is that you signed with an email that is different from the one you are using in this PR. Please sign again, making sure you use the same email. Thanks!

pquentin:
I triggered the CLA again and can confirm that it was signed, thank you!

pquentin:
buildkite test this please

pquentin:
buildkite test this please

PR #3160, opened 2025-11-14
Instrument ping with OTel

pquentin:


test_elasticsearch/utils.py
```diff
@@ -191,3 +191,4 @@ def wait_for_cluster_state_updates_to_finish(client, timeout=30):
     while time.time() < end_time:
         if not client.cluster.pending_tasks().get("tasks", ()):
             break
+        time.sleep(0.1)
```
pquentin, line 194: This is not really related, but this busy loop was adding a lot of noise in OTel mode.

noxfile.py
```diff
@@ -60,7 +60,7 @@ def test_otel(session):
         silent=False,
     )
 
-    argv = pytest_argv() + ["-m", "otel"]
+    argv = pytest_argv() + ["-m", "otel"] + session.posargs
```
pquentin, line 63: Again, not really related, but allowed things like `nox -rs test_otel-3.14 -- test_elasticsearch/test_server/test_otel.py::test_otel_ping -s`

miguelgrinberg (approved):

pmoust (approved):
LGTM

PR #3138, opened 2025-10-29
[DOCS] Clarify upgrades and edit generally

marciw:
👁️ **Preview** [docs/reference/index.md](https://docs-v3-preview.elastic.dev/elastic/elasticsearch-py/pull/3138/reference)

## Summary of changes
- Revised the "Major version upgrades" section in response to https://github.com/elastic/docs-content-internal/issues/421
- Clarified forward compatibility
- Added phrasing and table from .NET docs + made the table more accessible
- Added links to related docs
- Did a general edit for clarity, concision, Elastic style

## ❗ Notes for reviewers
- I'll apply similar changes to the other client docs. This one will serve as the template for updating the others.
- I tried to align this with both the [.NET docs](https://www.elastic.co/docs/reference/elasticsearch/clients/dotnet#_es_version_compatibility) and https://github.com/elastic/devtools-team/pull/936/ (will iterate more when moving the latter to docs-content)

### TODO
- [x] Compare to .NET changes; go with the best of both worlds
- [x] Tweak section heading?
- [ ] ~~Clarify "outages" (whether clients need hot-swapping during major upgrades) per https://github.com/elastic/docs-content/pull/3398#issuecomment-3423887315~~ <-- will do this in a follow-up issue; going to merge this in its current state as a "phase 1" edit

PR #3161, opened 2025-11-14
[DOCS] Continued: Clarify upgrades and edit generally

marciw:
Some quick improvements as part of https://github.com/elastic/docs-content-internal/issues/421.
Continues work started in https://github.com/elastic/elasticsearch-py/pull/3138

**AI-Assisted:** This PR was assisted by Cursor. 
**Usage:** Brainstorming; finding patterns in our docs; general ES research

### DIFF
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -79,7 +79,7 @@ As of version 8.0, {{es}} offers a [REST API compatibility mode](elasticsearch:/
 
 In the Python client, compatibility mode is **always enabled**. You don't need to configure it manually or set an environment variable. The client automatically sends compatibility headers with API requests so that it can communicate with the next major version of {{es}}.
 
-In compatibility mode, your Python client applications can continue running during an upgrade without restart or redeployment. For example, during a rolling upgrade of your {{es}} cluster from 8.x to 9.x, your 8.x Python client continues working throughout the process without any changes or intervention.
+For example, if you're upgrading {{es}} from 8.x to 9.x, you can continue to use the 8.x Python client during and after the {{es}} server upgrade, with the exception of [breaking changes](../release-notes/breaking-changes.md).
 
 **What to expect in compatibility mode:**
 


marciw:
@eedugon this isn't final yet, just hoping it's a step in the right direction. LMKWYT!

docs/reference/index.md
```diff
@@ -66,20 +65,34 @@ Language clients are **forward compatible**: each client version works with equi
...
+As of version 8.0, {{es}} offers a [REST API compatibility mode](elasticsearch://reference/elasticsearch/rest-apis/compatibility.md) for smoother upgrades. In compatibility mode, you can upgrade your {{es}} cluster to the next major version while continuing to use your existing client during the transition.
+
+### How compatibility mode works
+
+In the Python client, compatibility mode is **always enabled**. You don't need to configure it manually or set an environment variable. The client automatically sends compatibility headers with API requests so that it can communicate with the next major version of {{es}}.
 
-For example, if you're upgrading {{es}} from 8.x to 9.x, you can continue to use the 8.x Python client during and after the {{es}} server upgrade, with the exception of [breaking changes](../release-notes/breaking-changes.md).
+In compatibility mode, your Python client applications can continue running during an upgrade without restart or redeployment. For example, during a rolling upgrade of your {{es}} cluster from 8.x to 9.x, your 8.x Python client continues working throughout the process without any changes or intervention.
```
miguelgrinberg: This seems a bit less clear than the original text. Here we are saying "without changes or intervention", but right below we go on to explain that this isn't always true because of breaking changes. We also repeat "working without intervention" in line 92.

I think it would be more accurate to say something like "generally without any changes or interventionm with some exceptions mentioned below" or something along those lines.

marciw: Reverted this change for now and will try again later to clarify 
https://github.com/elastic/elasticsearch-py/pull/3161/commits/c7e8bf54e0f77ab3c8c310369cc59979868f7300

marciw:
pausing this; will continue after https://github.com/elastic/elasticsearch-py/pull/3194/ is merged

docs/reference/index.md
```diff
@@ -66,20 +65,34 @@ Language clients are **forward compatible**: each client version works with equi
...
+### How compatibility mode works
+
+In the Python client, compatibility mode is **always enabled**. You don't need to configure it manually or set an environment variable. The client automatically sends compatibility headers with API requests so that it can communicate with the next major version of {{es}}.
 
 For example, if you're upgrading {{es}} from 8.x to 9.x, you can continue to use the 8.x Python client during and after the {{es}} server upgrade, with the exception of [breaking changes](../release-notes/breaking-changes.md).
 
-In the Python client, compatibility mode is always enabled. 
+**What to expect in compatibility mode:**
+
+If you're using the Python client in compatibility mode to work with the next major version of {{es}}:
```
miguelgrinberg, line 86:
```suggestion
If you're taking advantage of the Python client compatibility mode to work with the next major version of {{es}}:
```

I think the wording should reflect that this is not something that can be turned on or off.

docs/reference/index.md
```diff
@@ -66,20 +65,34 @@ Language clients are **forward compatible**: each client version works with equi
...
+
+* API calls affected by [breaking changes](../release-notes/breaking-changes.md) might fail or behave differently.
```
miguelgrinberg, line 88: This is from a previous PR, but @pquentin noted that the breaking changes here refer to the server's next major release. This link should point to the server's breaking changes page instead of the client's.

PR #3169, opened 2025-11-19
load yaml tests from same branch as client

miguelgrinberg:
This change tries to load the YAML tests for CI from the same branch as the client, instead of always loading `main`. When running a PR build, the branch is not going to exist in the YAML tests repository, so in that case we fallback to `main` as before.

### DIFF
--- a/test_elasticsearch/test_server/test_rest_api_spec.py
+++ b/test_elasticsearch/test_server/test_rest_api_spec.py
@@ -514,6 +514,7 @@ try:
         branch_candidates.append("main")
 
     response = None
+    branch = "main"
     for branch in branch_candidates:
         yaml_tests_url = f"https://api.github.com/repos/elastic/elasticsearch-clients-tests/zipball/{branch}"
         response = http.request("GET", yaml_tests_url)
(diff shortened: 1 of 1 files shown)


test_elasticsearch/test_server/test_rest_api_spec.py
```diff
@@ -549,7 +567,7 @@ def remove_implicit_resolver(cls, tag_to_remove):
             for prefix in ("rest-api-spec/", "test/", "free/", "platinum/"):
                 if pytest_test_name.startswith(prefix):
                     pytest_test_name = pytest_test_name[len(prefix) :]
-            pytest_param_id = "%s[%d]" % (pytest_test_name, test_number)
+            pytest_param_id = "[%s]%s[%d]" % (branch, pytest_test_name, test_number)
```
miguelgrinberg: We do  not have to keep this forever, but initially this will be confirmation of which of all the possible branches was used in the end.

pquentin:
This is failing due to:

```
FAILED test_elasticsearch/test_server/test_rest_api_spec.py::test_rest_api_spec[[main]inference/10_basic[0]] - elastic_transport.ConnectionTimeout: Connection timed out
```

However, that test is part of `FAILING_TESTS`, so it should be ignored.

test_elasticsearch/test_server/test_rest_api_spec.py
```diff
@@ -35,6 +36,7 @@
 
 from elasticsearch import ApiError, ElasticsearchWarning, RequestError
 from elasticsearch._sync.client.utils import _base64_auth_header
+from elasticsearch._version import __versionstr__
```
pquentin, line 39: For what it's worth, `importlib.metadata.version` is the modern way to do that. Not asking you to change it though!

miguelgrinberg: I trust that we are less likely to break this in the future than the Python org.

PR #3196, opened 2025-11-21
Add variable support to YAML test expression parser

miguelgrinberg:
While debugging a YAML test failure in 8.19 I found that our parser does not appear to support variables, as opposite to expressions that are based on the response from the last request. Take the expression `$profile.uid` for example.

- `_lookup('$profile.uid')` is called
- `value` is assigned a dict that was a response in a previous operation
- we enter the for-loop with `step == '$profile'`
- the call to `self._resolve(step)` returns the value stored in `self._state['profile']` and assigns it to `step`
- the `assert step in value` fails

The changes to the parser add support for variables such as `$profile`. The `value` variable is only initialized to `last_response` when the expression does not start with a `$`, which means it is relative to that response. If the expression starts with `$` we initialize to `None`. Then inside the loop we differentiate between a reference to a variable that starts the expression (when `value` is `None`) versus one that is in the middle and should be treated as a key lookup in the current value.

The issue that prompted this change is a block of `match` statements highlighted here: https://github.com/elastic/elasticsearch-clients-tests/blob/8.19/tests/security/50_user_profile.yml#L80-L86, which are currently failing.

### DIFF
--- a/test_elasticsearch/test_server/test_rest_api_spec.py
+++ b/test_elasticsearch/test_server/test_rest_api_spec.py
@@ -411,16 +411,21 @@ class YamlRunner:
         return value
 
     def _lookup(self, path):
-        # fetch the possibly nested value from last_response
-        value = self.last_response
         if path == "$body":
-            return value
+            return self.last_response
+        if path.startswith("$"):
+            value = None
+        else:
+            value = self.last_response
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! LGTM.

PR #3217, opened 2025-12-04
Support for base64 packing of dense vectors

miguelgrinberg:
This PR includes three separate features:
- Support for numpy arrays in the DSL (big change, still work in progress)
- Support for base64 encoding of dense vectors
- Improvements to the quotes example adapted to use the above two features

I have to see if it makes sense to split this into two separate parts for the numpy and base64 changes.


miguelgrinberg:
Closing, as this is superseded by #3218 and #3219.

PR #3219, opened 2025-12-05
Add helper function to pack dense vectors for efficient uploading

miguelgrinberg:
This PR adds a utility function to encode dense vectors in base64 binary encoding for better ingest performance. There is also a benchmark script for this feature.

### DIFF
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -717,7 +717,7 @@ def parallel_bulk(
 def pack_dense_vector(vector: Union["np.ndarray", Sequence[float]]) -> str:
     """Helper function that packs a dense vector for efficient uploading.
 
-    :arg v: the list or numpy array to pack.
+    :arg vector: the list or numpy array to pack.
     """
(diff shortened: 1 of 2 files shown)


elasticsearch/helpers/actions.py
```diff
@@ -708,6 +714,21 @@ def _setup_queues(self) -> None:
             pool.join()
 
 
+def pack_dense_vector(vector: Union["np.ndarray", Sequence[float]]) -> str:
+    """Helper function that packs a dense vector for efficient uploading.
+
+    :arg v: the list or numpy array to pack.
```
pquentin: I like the name, because it means we can support more than numpy in the future if we want to.

pquentin (approved):
Thanks! LGTM.

elasticsearch/helpers/actions.py
miguelgrinberg: Even now, you can pass a list of floats and it will handle it too (I'm converting the list to numpy inside the function though).

PR #3275, opened 2026-01-22
Stop using pytz

pquentin:
It never was a test dependency, and the same information can be obtained from the Python standard library.

Looks like one of our dependency recently stopped using it, which is breaking integration tests: https://buildkite.com/elastic/elasticsearch-py-integration-tests/builds/6491#019be526-55e1-4d20-9f34-7eff3638e4b5.


pquentin:
For the record, the test dependency that dropped pytz is pandas 3.0.0.

PR #3286, opened 2026-01-27
Add async tabs to docs examples

miguelgrinberg:
Add Sync/Async tabs to all documentation examples except the DSL section, which is going to come in a separate PR.

### DIFF
--- a/docs/reference/client-helpers.md
+++ b/docs/reference/client-helpers.md
@@ -12,7 +12,7 @@ You can find here a collection of simple helper functions that abstract some spe
 
 There are several helpers for the bulk API since its requirement for specific formatting and other considerations can make it cumbersome if used directly.
 
-All bulk helpers accept an instance of `Elasticsearch` or `AsyncElasticsearch` classes and an iterable `action` (any iterable, can also be a generator, which is ideal in most cases since it allows you to index large datasets without the need of loading them into memory).
+All bulk helpers accept an instance of `Elasticsearch` class and an iterable `action` (any iterable, can also be a generator, which is ideal in most cases since it allows you to index large datasets without the need of loading them into memory). For asynchronous Python, use the bulk helpers with the `async_` prefix and pass an `AsyncElasticsearch` instance as first argument.
 
 The items in the iterable `action` should be the documents we wish to index in several formats. The most common one is the same as returned by `search()`, for example:
 
--- a/docs/reference/esql-pandas.md
+++ b/docs/reference/esql-pandas.md
@@ -326,10 +326,12 @@ Use the ES|QL CSV import to convert the `employees` dataset to a Pandas datafram
 from io import StringIO
 from elasticsearch import Elasticsearch
 import pandas as pd
+
 client = Elasticsearch(
     "https://[host].elastic-cloud.com",
     api_key="...",
 )
+
 response = client.esql.query(
     query="FROM employees | LIMIT 500",
     format="csv",
@@ -342,15 +344,17 @@ print(df)
 :::{tab-item} Async Python
 :sync: async
 ```python
+import asyncio
 from io import StringIO
 from elasticsearch import AsyncElasticsearch
 import pandas as pd
+
 client = AsyncElasticsearch(
     "https://[host].elastic-cloud.com",
     api_key="...",
 )
 
-async def example():
+async def main():
     response = await client.esql.query(
         query="FROM employees | LIMIT 500",
         format="csv",
--- a/docs/reference/esql-query-builder.md
+++ b/docs/reference/esql-query-builder.md
@@ -36,18 +36,25 @@ To execute this query, you can pass it to the `client.esql.query()` endpoint:
 :::{tab-item} Standard Python
 :sync: sync
 ```python
->>> from elasticsearch import Elasticsearch
->>> client = Elasticsearch(hosts=[os.environ['ELASTICSEARCH_URL']])
->>> response = client.esql.query(query=query)
+from elasticsearch import Elasticsearch
+
+client = Elasticsearch(hosts=[os.environ['ELASTICSEARCH_URL']])
+response = client.esql.query(query=query)
 ```
 :::
 
 :::{tab-item} Async Python
 :sync: async
 ```python
->>> from elasticsearch import AsyncElasticsearch
->>> client = AsyncElasticsearch(hosts=[os.environ['ELASTICSEARCH_URL']])
->>> response = await client.esql.query(query=query)
+import asyncio
+from elasticsearch import AsyncElasticsearch
+
+client = AsyncElasticsearch(hosts=[os.environ['ELASTICSEARCH_URL']])
+
+async def main():
+    response = await client.esql.query(query=query)
+
+asyncio.run(main())
 ```
 :::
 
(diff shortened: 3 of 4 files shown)


docs/reference/client-helpers.md
```diff
@@ -12,7 +12,7 @@ You can find here a collection of simple helper functions that abstract some spe
 
 There are several helpers for the bulk API since its requirement for specific formatting and other considerations can make it cumbersome if used directly.
 
-All bulk helpers accept an instance of `Elasticsearch` class and an iterable `action` (any iterable, can also be a generator, which is ideal in most cases since it allows you to index large datasets without the need of loading them into memory).
+All bulk helpers accept an instance of `Elasticsearch` or `AsyncElasticsearch` classes and an iterable `action` (any iterable, can also be a generator, which is ideal in most cases since it allows you to index large datasets without the need of loading them into memory).
```
pquentin: Should we clarify that you should use async helpers if you have an `AsyncElasticsearch` instance?

docs/reference/connecting.md
```diff
@@ -294,47 +569,48 @@ The async client shouldn’t be used within Function-as-a-Service as a new event
...
 
 ```python
+import os
 
 # Client initialization
 client = Elasticsearch(
     cloud_id="deployment-name:ABCD...",
-    api_key=...
+    api_key=os.environ["ELASTIC_API_KEY"
```
pquentin:
```suggestion
    api_key=os.environ["ELASTIC_API_KEY"]
```

docs/reference/connecting.md
```diff
@@ -294,47 +569,48 @@ The async client shouldn’t be used within Function-as-a-Service as a new event
...
```
pquentin:
```suggestion
    api_key=os.environ["ELASTIC_API_KEY"]
```

docs/reference/esql-pandas.md
```diff
@@ -332,6 +337,30 @@ response = client.esql.query(
...
+)
+
+async def example():
+    response = await client.esql.query(
+        query="FROM employees | LIMIT 500",
+        format="csv",
+    )
+    df = pd.read_csv(StringIO(response.body))
+    print(df)
+```
```
pquentin:
```suggestion

asyncio.run(main())
```
```

File: docs/reference/esql-pandas.md [hunk 5]
```diff
@@ -332,6 +337,30 @@ response = client.esql.query(
...
+:::{tab-item} Async Python
+:sync: async
+```python
+from io import StringIO
+from elasticsearch import AsyncElasticsearch
+import pandas as pd
+client = AsyncElasticsearch(
+    "https://[host].elastic-cloud.com",
+    api_key="...",
+)
```
pquentin, line 355:
```suggestion
import asyncio
import os
from io import StringIO
from elasticsearch import AsyncElasticsearch
import pandas as pd

client = AsyncElasticsearch(
    os.environ["ES_LOCAL_URL"]
    api_key=os.environ["ES_LOCAL_API_KEY"],
)
```

docs/reference/esql-pandas.md
pquentin: (Also applies to the rest of the page.)

docs/reference/esql-query-builder.md
```diff
@@ -30,11 +30,28 @@ FROM employees
...
 ```
+:::
+
+:::{tab-item} Async Python
+:sync: async
+```python
+>>> from elasticsearch import AsyncElasticsearch
+>>> client = AsyncElasticsearch(hosts=[os.environ['ELASTICSEARCH_URL']])
+>>> response = await client.esql.query(query=query)
+```
```
pquentin, line 58: Should we clarify that `python -m asyncio` is needed for this to work?

docs/reference/examples.md
```diff
@@ -33,43 +38,115 @@ doc = {
...
+doc = {
+    'author': 'author_name',
+    'text': 'Interesting content...',
+    'timestamp': datetime.now(),
+}
+
+async def example():
+    resp = await client.index(index="test-index", id=1, document=doc)
+    print(resp['result'])
+```
```
pquentin, line 62: Can we make this example runnable with `import asyncio` and `asyncio.run(example())`?

This applies to the rest of the file, except for one-liners, probably.

docs/reference/getting-started.md
```diff
@@ -19,25 +19,61 @@ This page guides you through the installation process of the Python client, show
...
+```
+:::
 
+::::
 Refer to the [*Installation*](/reference/installation.md) page to learn more.
 
 
 ### Connecting [_connecting]
 
 You can connect to the Elastic Cloud using an API key and the Elasticsearch endpoint.
```
pquentin, line 45: Unrelated to this PR, but we should likely advertise start-local instead.

docs/reference/getting-started.md
```diff
@@ -59,12 +95,32 @@ Time to use Elasticsearch! This section walks you through the basic, and most im
...
+:sync: sync
 ```py
 client.indices.create(index="my_index")
 ```
+:::
+
+:::{tab-item} Async Python
+:sync: async
+```py
+await client.indices.create(index="my_index")
```
pquentin, line 111: While there's value in making all examples runnable, especially for beginners not necessarily fully familiar with asyncio, it also feels like wrapping everything in a function is going to add noise! So I think I like this as it is.

pquentin (commented):
Should we add AWS Lambda/Azure Functions/GCP Cloud Functions examples with asyncio? From a quick online search it seems possible.

Regarding the environment variable names, any thoughts about using the elastic-start-local conventions? ES_LOCAL_URL, ES_LOCAL_PASSWORD and ES_LOCAL_API_KEY

docs/reference/esql-pandas.md
miguelgrinberg: Do you think it is necessary to repeat this in every example? I think it is okay to have it in some, but not sure if having this in every one helps anyone.

docs/reference/client-helpers.md
miguelgrinberg: Yes, good point, I should note that here.

docs/reference/esql-query-builder.md
miguelgrinberg: I think it would be better to rework both examples here so that they are not shown as done in the Python prompt.

docs/reference/getting-started.md
miguelgrinberg: Fully agree, probably more than you since you've suggested adding more `asyncio.run(main())` lines that I thought prudent to include. :)

docs/reference/esql-pandas.md
pquentin: The asyncio import is still missing.

miguelgrinberg:
> Should we add AWS Lambda/Azure Functions/GCP Cloud Functions examples with asyncio?

@pquentin Forgot to answer this. We have a warning about using asyncio in severless functions, due to the limitation that it is not possible to keep the loop running between invocations ([link](https://www.elastic.co/docs/reference/elasticsearch/clients/python/connecting#connecting-faas)). This actually makes sense, so I thought it better to not show examples of a use case that we believe to be sub-optimal. Do you think we should show those examples anyway? Or maybe show the examples, but add a comment in each one saying that using the sync client is a much better option?

miguelgrinberg:
@pquentin I think I addressed all the feedback except these two:
- I did not include examples of FaaS, I explained why in a comment above.
- I did not include start-local, as this is unrelated and should be done for all clients.

PR #3288, opened 2026-01-29
add breaking change note for exclude_vectors option

miguelgrinberg:


pquentin:
Actually, should we add a link from the main release notes page? I don't think most users are going to look at the breaking change page before upgrading a minor version. https://docs-v3-preview.elastic.dev/elastic/elasticsearch-py/pull/3288/release-notes.

miguelgrinberg:
I don't know. Looking at the ES release notes, they do it like this too, there are no references to breaking changes in the main release notes page. For now I'll commit it like this.

Seems like a design or organization issue, I would rather have all release notes in a flat list, but I suppose that needs to be passed on to the docs team to consider?

PR #3301, opened 2026-02-03
Bring 9.3.0 release to parent

miguelgrinberg:


elasticsearch/_version.py
```diff
@@ -15,5 +15,5 @@
 #  specific language governing permissions and limitations
 #  under the License.
 
-__versionstr__ = "9.2.1"
-__es_specification_commit__ = "0d0076662a02bb2513da46f32c2bef6d3b6bf164"
+__versionstr__ = "9.4.0.dev0"
```
pquentin: This change breaks test_user_agent_default in test_elasticsearch/test_server/test_vectorstore/test_vectorstore.py. It would also break a similar test in test_elasticsearch/test_server/test_vectorstore/test_embedding_service.py but this one is skipped for some reason.

miguelgrinberg: Okay, then I'm not going to pursue using `dev0`, not worth the trouble.

Issue #3307, opened 2026-02-06
Send Elastic-Api-Version header when talking to serverless


miguelgrinberg:
Parent issue: https://github.com/elastic/devtools-team/issues/922

PR #3329, opened 2026-02-20
Send Elastic-Api-Version header when talking to serverless

margaretjgu:
Closes https://github.com/elastic/elasticsearch-py/issues/3307

Add `server_mode` parameter to sycn & async clients to support severless. 

When `server_mode="serverless"`, the client sends an `Elastic-Api-Version: 2023-10-31` header instead of the `compatible-with=9 Accept/Content-Type` used for on prem and cloud
- Accepts "stack" (default/current) or `serverless`
- Blocks sniffing when serverless (raises ValueError)
- Followed js approach

### DIFF
--- a/elasticsearch/_async/client/_base.py
+++ b/elasticsearch/_async/client/_base.py
@@ -46,7 +46,7 @@ from elastic_transport import (
 from elastic_transport.client_utils import DEFAULT, DefaultType
 
 from ..._otel import OpenTelemetry
-from ..._version import __versionstr__, _SERVERLESS_API_VERSION
+from ..._version import _SERVERLESS_API_VERSION, __versionstr__
 from ...compat import warn_stacklevel
 from ...exceptions import (
     HTTP_EXCEPTIONS,
--- a/elasticsearch/_sync/client/_base.py
+++ b/elasticsearch/_sync/client/_base.py
@@ -46,7 +46,7 @@ from elastic_transport import (
 from elastic_transport.client_utils import DEFAULT, DefaultType
 
 from ..._otel import OpenTelemetry
-from ..._version import __versionstr__, _SERVERLESS_API_VERSION
+from ..._version import _SERVERLESS_API_VERSION, __versionstr__
 from ...compat import warn_stacklevel
 from ...exceptions import (
     HTTP_EXCEPTIONS,


elasticsearch/_async/client/_base.py
```diff
@@ -64,6 +64,7 @@
 )
 _COMPAT_MIMETYPE_RE = re.compile(r"application/(json|x-ndjson|vnd\.mapbox-vector-tile)")
 _COMPAT_MIMETYPE_SUB = _COMPAT_MIMETYPE_TEMPLATE % (r"\g<1>",)
+_SERVERLESS_API_VERSION = "2023-10-31"
```
miguelgrinberg: To avoid having the version duplicated in the _async and _sync sides, maybe we should relocated this variable to `_version.py`?

test_elasticsearch/test_async/test_transport.py
```diff
@@ -768,3 +768,92 @@ async def test_warning_header(headers):
...
+        client = AsyncElasticsearch(
+            "http://localhost:9200", meta_header=False, node_class=DummyNode
+        )
+        await client.info()
+
+        calls = client.transport.node_pool.get().calls
+        assert 1 == len(calls)
+        headers = calls[0][1]["headers"]
+        assert headers["accept"] == (
+            "application/vnd.elasticsearch+json; compatible-with=9"
```
miguelgrinberg: It would be best to have the "9" retrieved from the version string, so that this test does not break once we move to v10.

test_elasticsearch/test_async/test_transport.py
```diff
@@ -768,3 +768,92 @@ async def test_warning_header(headers):
...
+            meta_header=False,
+            node_class=DummyNode,
+            server_mode="serverless",
+        )
+        await client.info()
+
+        calls = client.transport.node_pool.get().calls
+        assert 1 == len(calls)
+        headers = calls[0][1]["headers"]
+        assert headers["elastic-api-version"] == "2023-10-31"
```
miguelgrinberg: Here too, we can import the version string to use in the test.

test_elasticsearch/test_async/test_transport.py
```diff
@@ -768,3 +768,92 @@ async def test_warning_header(headers):
...
+        await client.search(query={"match_all": {}})
+
+        calls = client.transport.node_pool.get().calls
+        assert 1 == len(calls)
+        headers = calls[0][1]["headers"]
+        assert headers["elastic-api-version"] == "2023-10-31"
+        assert headers["accept"] == "application/json"
+        assert headers["content-type"] == "application/json"
+        assert "compatible-with" not in headers.get("accept", "")
+        assert "compatible-with" not in headers.get("content-type", "")
```
miguelgrinberg, line 824: This and the previous test are pretty much doing the same thing and could be combined into one.

Or was the intention to do one test with GET and one with POST? That could have value, but in that case maybe the test functions should reflect that this is what we are testing.

miguelgrinberg:
Looks pretty good overal, thanks for taking the time to do this!

The only use case that I think is left is the one about namespaces, which I think should inherit the setting from the parent.

```bash
>>> from elasticsearch import AsyncElasticsearch
>>> es = AsyncElasticsearch(hosts=['http://localhost:9200'], server_mode='serverless')
>>> es._is_serverless
True
>>> es.indices._is_serverless
False
```

test_elasticsearch/test_async/test_transport.py
margaretjgu: yes i renamed the tests to make the intention more clear 
`test_serverless_mode_sends_api_version_header_on_get`
`test_serverless_mode_sends_plain_headers_on_post`

elasticsearch/_async/client/_base.py
margaretjgu: done

PR #3331, opened 2026-02-23
Add Vale docs linting support

theletterf:
This PR adds support for Vale docs linting to the repository. Linting only triggers when Markdown files inside the docs folder are edited. Vale linting never blocks pull requests. For more information, refer to [Vale linter](https://www.elastic.co/docs/contribute-docs/vale-linter).

Contributes to https://github.com/elastic/docs-content-internal/issues/670

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -11,6 +11,7 @@ jobs:
   docs-preview:
     uses: elastic/docs-builder/.github/workflows/preview-build.yml@main
     with:
+      enable-vale-linting: true
       path-pattern: docs/**
     permissions:
       deployments: write


theletterf:
Interesting. Could you provide some examples? In theory, Vale only lints Markdown changed lines and ignores code blocks.

pquentin:
Right, I'm talking about what Vale adds to the "Problems" tab in my editor. For example, we use "via" a lot and it doesn't like it.

There are many words like this. Consider this sentence:

> In this example, you can see:

Vale suggests replacing "see" with "refer to" but that never works. Another example:

> In addition to these, `query_vector` or `query_vector_builder` must be given as well.

Vale suggests replacing "In addition" with "Also", which doesn't work. Regarding execute, we have a lot of sentences like this:

> You can execute your search by calling the `.execute()` method that will return a `Response` object.

It would be nice to silent some rules or allow some words on a per-project basis to avoid all the squiggly lines.

theletterf:
Point taken! Could you raise some issues in the vale-rules repo?

PR #3422, opened 2026-05-11
Validate ES|QL source index names

NguyenCong2k:
Summary
reject ES|QL query delimiters in source index names before rendering
cover FROM, TS, and LOOKUP JOIN source index arguments
add regression tests for pipeline, comma, semicolon/newline, and backtick inputs
Test
python -m pytest test_elasticsearch\test_esql.py -q
Result: 43 passed.


NguyenCong2k:
I have completed the Elastic Contributor Agreement. Could you please recheck the CLA status?

miguelgrinberg:
Hi, thanks. Unfortunately this is not a correct solution to the problem, since some of the characters that you are filtering are valid in Elasticsearch index names. I have coded a much simpler solution in #3425.

PR #3445, opened 2026-06-02
Forward opaque_id in async_scan

dfgvaetyj3456356-hash:
## What

Adds `opaque_id` to the transport options extracted by `async_scan()` so the initial search client options and subsequent scroll client options can carry the same `x-opaque-id` behavior as the sync `scan()` helper.

## Why

The sync helper already treats `opaque_id` as a helper-level transport option, but the async helper only forwarded headers/auth options. That meant `opaque_id` was left in request kwargs instead of being applied through `client.options()`, and `scroll_kwargs` could not override it for scroll requests.

## Tests

- `python -m py_compile elasticsearch\_async\helpers.py test_elasticsearch\test_async\test_server\test_helpers.py`
- `git diff --check`
- `python -m pytest test_elasticsearch\test_async\test_server\test_helpers.py -k scan_auth_kwargs -rs` (selected tests skipped locally because no Elasticsearch service is listening on localhost:9200)

### DIFF
--- a/elasticsearch/_async/helpers.py
+++ b/elasticsearch/_async/helpers.py
@@ -480,7 +480,14 @@ async def async_scan(
         # Grab options that should be propagated to every
         # API call within this helper instead of just 'search()'
         transport_kwargs = {}
-        for key in ("headers", "api_key", "http_auth", "basic_auth", "bearer_auth"):
+        for key in (
+            "headers",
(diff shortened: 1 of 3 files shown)


miguelgrinberg (approved):

miguelgrinberg:
Thanks!

PR #3449, opened 2026-06-05
[Issue# 3448] Fix mutable default arguments in reindex functions

### LINKED ISSUE #3448: Potential AttributeError in scan/async_scan when query is a non-dict Mapping type
**Elasticsearch version:** N/A (client-side issue)

**Python version:** All supported versions

**Description:**

The `scan()` and `async_scan()` functions have a potential bug where they unconditionally call `.copy()` on the query object, which may not support this method if it's a Mapping type that isn't a dict.

**Current code:**

Sync version: https://github.com/elastic/elasticsearch-py/blob/6df97e3c21adffeb33d0a4c64dc784e618d85f41/elasticsearch/helpers/actions.py#L785

Async version: https://github.com/elastic/elasticsearch-py/blob/6df97e3c21adffeb33d0a4c64dc784e618d85f41/elasticsearch/_async/helpers.py#L476

```python
if not preserve_order:
    query = query.copy() if query else {}  # Could fail if query is a Mapping without .copy()
    query["sort"] = "_doc"
```

**Problem:** If a user passes a query that is a collections.abc.Mapping subclass (not a dict) that doesn't implement a .copy() method, this will raise an AttributeError. This exception is NOT caught by the existing try/except block which only catches TypeError. 
**Expected behavior:** The function should handle any Mapping-like object safely, converting it to a dict if necessary before attempting to modify it. 

**Suggested fix:**
```python
if not preserve_order:
    if query:
        if isinstance(query, dict):
            query = query.copy()
        else:
            # Convert to dict for non-dict Mapping types
            query = dict(query)
    else:
        query = {}
    query["sort"] = "_doc"
```

**Steps to reproduce:**
```python
from collections.abc import MutableMapping
from elasticsearch.helpers import scan

class CustomMapping(MutableMapping):
    def __init__(self):
        self.data = {"match_all": {}}
    # ... implement required methods but NOT .copy()

client = Elasticsearch(...)
query = CustomMapping()
# This would raise AttributeError
list(scan(client, index="test", query=query, preserve_order=False))
```

Affected files:
Sync: https://github.com/elastic/elasticsearch-py/blob/6df97e3c21adffeb33d0a4c64dc784e618d85f41/elasticsearch/helpers/actions.py#L785
Async: https://github.com/elastic/elasticsearch-py/blob/6df97e3c21adffeb33d0a4c64dc784e618d85f41/elasticsearch/_async/helpers.py#L476

jubins:
## Summary
Fixes #3448 a classic Python gotcha where mutable default arguments (empty dicts `{}`) were used in the `reindex()` and `async_reindex()` functions. This could cause state to be shared between function calls if the dictionaries are modified.

## Problem
The `reindex()` and `async_reindex()` functions had mutable default arguments:

```python
def reindex(
    ...
    scan_kwargs: MutableMapping[str, Any] = {},  # MUTABLE DEFAULT
    bulk_kwargs: MutableMapping[str, Any] = {},  # MUTABLE DEFAULT
):
```
**Why this is a problem:**
In Python, default arguments are evaluated once at function definition time
The same dict object is reused for all function calls
If the function (or any code it calls) modifies these dicts, the changes persist across function calls
This is a well-known Python anti-pattern that can lead to subtle, hard-to-debug issues

**Example of the bug:**
```python
# First call - modifies the default dict
reindex(client, "src", "target")  # Internally adds something to scan_kwargs
```

```python
# Second call - gets the MODIFIED dict from the first call!
reindex(client, "src2", "target2")  # Unexpected behavior!
```

**Tests Added**
Added two unit tests in test_elasticsearch/test_helpers.py (class TestMutableDefaultArguments):
- `test_reindex_does_not_mutate_default_arguments`: Verifies that calling reindex() multiple times without passing scan_kwargs or bulk_kwargs doesn't cause issues due to shared mutable defaults.
- `test_reindex_scan_kwargs_not_shared_between_calls`: Verifies that each call to reindex() receives separate dict instances, ensuring modifications don't persist across calls.

**Impact**
- Low risk: This is a defensive fix that prevents a potential bug
- No breaking changes: The function signature is backward compatible (callers can still pass dicts)
- Improved reliability: Eliminates the possibility of state leakage between reindex calls

### DIFF
--- a/elasticsearch/_async/helpers.py
+++ b/elasticsearch/_async/helpers.py
@@ -580,8 +580,8 @@ async def async_reindex(
     chunk_size: int = 500,
     scroll: str = "5m",
     op_type: Optional[str] = None,
-    scan_kwargs: MutableMapping[str, Any] = {},
-    bulk_kwargs: MutableMapping[str, Any] = {},
+    scan_kwargs: Optional[MutableMapping[str, Any]] = None,
+    bulk_kwargs: Optional[MutableMapping[str, Any]] = None,
 ) -> Tuple[int, Union[int, List[Any]]]:
     """
     Reindex all documents from one index that satisfy a given query
@@ -616,6 +616,11 @@ async def async_reindex(
     :arg bulk_kwargs: additional kwargs to be passed to
         :func:`~elasticsearch.helpers.async_bulk`
     """
+    if scan_kwargs is None:
+        scan_kwargs = {}
+    if bulk_kwargs is None:
+        bulk_kwargs = {}
+
     target_client = client if target_client is None else target_client
     docs = async_scan(
         client, query=query, index=source_index, scroll=scroll, **scan_kwargs
--- a/elasticsearch/helpers/actions.py
+++ b/elasticsearch/helpers/actions.py
@@ -888,8 +888,8 @@ def reindex(
     chunk_size: int = 500,
     scroll: str = "5m",
     op_type: Optional[str] = None,
-    scan_kwargs: MutableMapping[str, Any] = {},
-    bulk_kwargs: MutableMapping[str, Any] = {},
+    scan_kwargs: Optional[MutableMapping[str, Any]] = None,
+    bulk_kwargs: Optional[MutableMapping[str, Any]] = None,
 ) -> Tuple[int, Union[int, List[Dict[str, Any]]]]:
     """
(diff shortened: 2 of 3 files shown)


miguelgrinberg:
This is extremely low priority. Will look at it when I have time.

PR #3491, opened 2026-07-21
fix: stabilize optional imports with TYPE_CHECKING branches

### LINKED ISSUE #3481: fix mypy unused-ignore flakiness on optional imports (serializer.py, document_base.py)
`# type: ignore[assignment]` on the `except ImportError: x = None` fallback in `serializer.py` (pyarrow) and `document_base.py` (annotationlib, UnionType) flips between needed/unused depending on mypy/pyarrow-stubs/Python version. Hit twice already: `generates_9.5_code` (6b0aa4e0, 408d9c92) and main (#3475). #3475 suppressed it via a module-wide `warn_unused_ignores = false` override, which also hides any real unused-ignore added to those files later.

Fix: split type-checking and runtime import paths, mypy always resolves the `if` branch, the `else` is never type-checked (verified: no error with pyarrow installed, uninstalled, or a garbage assignment in the except).

```python
if TYPE_CHECKING:
    import pyarrow as pa
else:
    try:
        import pyarrow as pa
        ...
        __all__.append("PyArrowSerializer")
    except ImportError:
        pa = None
```

Same shape for `annotationlib` and `UnionType`. Remove the two overrides from `pyproject.toml`.

Rejected: annotate `pa: Optional[ModuleType]` upfront, skip the split. Fails with `no-redef`, mypy treats the annotation and the import as conflicting.

Effort: ~30-60 min, 2 files, verify across the Python version matrix.

Solaris-star:
## Summary

Fixes #3481.

`# type: ignore[assignment]` on optional `ImportError` fallbacks for `pyarrow` (`serializer.py`), `annotationlib`, and `UnionType` (`document_base.py`) flips between needed and unused depending on mypy / stub / Python version. That forced module-wide `warn_unused_ignores = false` overrides in `pyproject.toml`, which also hide real unused ignores later.

## Change

Split each optional import into `TYPE_CHECKING` vs runtime paths:

```python
if TYPE_CHECKING:
    import pyarrow as pa
else:
    try:
        import pyarrow as pa
        __all__.append("PyArrowSerializer")
    except ImportError:
        pa = None
```

Same shape for `annotationlib` and `UnionType`. Remove the two mypy overrides from `pyproject.toml`.

This matches the approach described on the issue (and the earlier closed attempt #3483).

## Test plan

- [x] AST parse of both modules
- [ ] CI mypy matrix on PR (authoritative)
- [x] Diff limited to the three files called out in #3481

## Notes

Elastic CLA may be required for first-time contributors: https://www.elastic.co/contributor-agreement

### DIFF
--- a/elasticsearch/dsl/document_base.py
+++ b/elasticsearch/dsl/document_base.py
@@ -36,15 +36,21 @@ from typing import (
 
 from typing_extensions import _AnnotatedAlias
 
-try:
+if TYPE_CHECKING:
     import annotationlib
-except ImportError:
-    annotationlib = None  # type: ignore[assignment]
+else:
+    try:
+        import annotationlib
+    except ImportError:
+        annotationlib = None
 
-try:
+if TYPE_CHECKING:
     from types import UnionType
-except ImportError:
-    UnionType = None  # type: ignore[assignment, misc]
+else:
+    try:
+        from types import UnionType
+    except ImportError:
+        UnionType = None
 
 from typing_extensions import dataclass_transform
 
--- a/elasticsearch/serializer.py
+++ b/elasticsearch/serializer.py
@@ -18,7 +18,7 @@
 import uuid
 from datetime import date, datetime
 from decimal import Decimal
-from typing import Any, ClassVar, Dict, Tuple
+from typing import TYPE_CHECKING, Any, ClassVar, Dict, Tuple
 
 from elastic_transport import JsonSerializer as _JsonSerializer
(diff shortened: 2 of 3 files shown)


margaretjgu:
After reevaluating, this approach seems to overcomplicate the linting fix. Following standard Python/mypy practice ( [python/mypy#8823](https://github.com/python/mypy/issues/8823)), we'll use the unused-ignore error code on the existing type: ignore comments instead of restructuring the imports. 
Closing this in favour of that simpler fix. Thanks for the contribution!

PR #3515, opened 2026-08-07
Feat/codegen updates

tom-whitehead:
- Adds codegen method for retriever API
- Fixes different python type rendering functionality


tom-whitehead:
🤦

PR #3516, opened 2026-08-08
feat: retrievers API codegen, templates and unit tests

### LINKED ISSUE #3513: Support for hybrid search via `retriever` API in DSL
## What
It would be nice to have first class support for the `retriever` API. Specifically, I would like to be able to configure hybrid RRF searches.

## Why
Currently the Python DSL supports the old `rank` API, which I believe is deprecated on 9.x.

As far as I can see there is no support for the newer `retriever` API which I would like to use for RRF hybrid search in the DSL. If I wanted to do this I think I would have to do something like: 
```python
s = AsyncSearch(index="my-index").extra(
    retriever={
        "rrf": {
            "retrievers": [
                {"standard": {...}},
                {"knn": {...}},
            ],
            "rank_constant": 20,
            "rank_window_size": 100,
        }
    }
)[:10]
```

It would be great to be able to configure this with a `retriever` method on `SearchBase`. 

This doesn't look like a small change though as per the [docs](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrievers) the `retriever` API supports seven quite different retrievers, not just RRF that I'm interested in. Presumably these would need to all be implemented together. 

My idea would be to Add a `Retriever` base class in the mould of the `Query` base class.

tom-whitehead:
Resolve #3513 

## Description of changes
- This PR adds first class support for the [retrievers API](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrievers) to the Python DSL. 
- I followed the existing codegen pattern, largely copying the template from query and simplifying/adding new functionality where needed. Most notably there were a couple of places (`InnerRetriever` in `linear` and `RRFRetrieverComponent` in `rrf` that needed some custom logic for the to and from dict serialisation to work. This was all largely written by me and reviewed by Claude.
- Added unit tests in a similar style to the query ones. These were largely generated by Claude and reviewed by me. 

## Testing
- All unit tests pass, including the difficult cases mentioned involving `InnerRetriever` and `RRFRetrieverComponent`. 
- For all retrievers possible I've executed queries against my enterprise cluster and verified they work. This wasn't possible for `RuleRetriever` as I didn't have any rule sets and `TextSimilarityRerankerRetriever` as I didn't have a compatible reranking endpoint/model.
- I've also tested printing queries with `to_dict` and running them straight in the ES console, again against one of my company's indices.


utils/dsl-generator.py
```diff
@@ -58,8 +65,11 @@
 }
 
 
+ACRONYMS = {"ip": "IP", "rrf": "RRF"}
```
tom-whitehead, line 68: I was getting `RrfRetriever` from the `property_to_class_name` function below, but `RRFRetrieverComponent` in `types.py` straight from the spec so I thought it was better to make them consistent.

utils/dsl-generator.py
```diff
@@ -296,25 +306,29 @@ def get_python_type(self, schema_type, for_response=False):
                     self.interfaces.append("PipeSeparatedFlags")
                 return '"types.PipeSeparatedFlags"', None
             else:
-                # generic union type
-                types = list(
-                    dict.fromkeys(  # eliminate duplicates
-                        [
-                            self.get_python_type(t, for_response=for_response)
```
tom-whitehead, line 303: `dict.fromkeys` crashed when `get_python_type` returns a `dict` instead of `None` for `_params_dsl` here as it's not hashable, notably for `RRFRetriever`'s `RRFRetrieverComponent`.

utils/dsl-generator.py
```diff
@@ -296,25 +306,29 @@ def get_python_type(self, schema_type, for_response=False):
...
-                return "Union[" + ", ".join([type_ for type_, _ in types]) + "]", None
+                    return next(iter(types.items()))
+
+                # NOTE: In cases of multiple types in the Union that have different
+                # `DslBase._param_defs`, None will be returned. This would break
+                # `to_dict` serialisation. Fortunately, no such cases exist in the
+                # ElasticSearch schema as it would require a wholesale refactor.
+                params = [param for param in types.values() if param]
+                param = params[0] if len(params) == 1 else None
+                return "Union[" + ", ".join(types) + "]", param
```
tom-whitehead, line 324: Hardcoding `None` here for the `_params_dsl` meant that `RRFRetriever` didn't get `_params_dsl` for the type of it's parameter `retrievers`. This meant that `to_dict` didn't convert the inner retriever properly to a dict but left it as an object.

utils/dsl-generator.py
```diff
@@ -34,17 +34,24 @@
...
 # map with Elasticsearch type replacements
 # keys and values are in given in "{namespace}:{name}" format
 TYPE_REPLACEMENTS = {
     "_types.query_dsl:DistanceFeatureQuery": "_types.query_dsl:DistanceFeatureQueryBase",
 }
 
+# interfaces that wrap a DSL type and so need its shortcut applied on assignment
+# keys are in "{namespace}:{name}" format
+INTERFACE_PARAMS = {
+    "_types:InnerRetriever": {"type": "retriever"},
```
tom-whitehead, line 52: `LinearRetriever` didn't get `_param_dsl` for `retrievers: Sequence[InnerRetriever]` which meant the object wasn't getting properly converted to a dictionary on calls to `to_dict`.

tom-whitehead:
Hey @margaretjgu, thanks for the review. One of these led to a change that I will push (3). The others I don't think are genuine issues 😅. I think the LLM reviewer needed more context than just the PR diff as they all referenced things outside of the PR scope. 

1. Bug: `_retriever` not deep-copied in `_clone`
Nothing is deep-copied in clone. The only copying is done shallow copying of containers, the underlying objects are not copied. The docstring says this too "Performs a shallow copy of all the underlying object". The `_retriever` implementation here is in line with how it's done for `query` and `post_filter` which are not containers ([ref](https://github.com/elastic/elasticsearch-py/blob/00d3c7e9e5d876c13245a082941d732532a16773/elasticsearch/dsl/search_base.py#L493)). 

2. Bug: `R()` passes `_proxied` object back unchanged when it's a `RetrieverComponent`
`RetrieverComponent` is an `AttrDict`, which has no `_proxied` attribute so this isn't correct. Also that conditional is unreachable in the case of an `AttrDict` which has already been converted to a Python `dict` and gone down the `MutableMapping` arm of `R`.

3. Potential issue: `update_from_dict` retriever cast is unsafe
This isn't really an issue as, even though technically `R` can return an `AttrDict` (`RetrieverComponent`), it will always be in a list of retrievers, in an outer `Retriever` if it has a valid request shape. Anything that has an unsafe cast would be an invalid retriever query. The `RetrieverComponent(AttrDict)` represents a `Retriever` wrapped with a weight and/or normalizer inside a list on `rrf` or `linear` outer retriever. I have added a new validation helper function though and removed the `cast` calls – a `ValueError` will be thrown for an invalid retriever request shape instead of a round-trip for a 400. Test case added too.

4. `RetrieverComponent.to_dict` override is pointless
I think the AI review doesn't have the necessary context here 🧐. The override defaults `recursive: bool = True` which is the opposite to `AttrDict.to_dict`. Flipping that boolean flag is the sole point of the class. Without it `RetrieverComponent({"retriever": StandardRetriever(...), "weight": 2.0}).to_dict()` yields `{'retriever': StandardRetriever(query=Match(...)), 'weight': 2.0}` (i.e. the nested retriever remains as an object). There's a test case for this one too (`test_rrf_retriever_component_with_weight_to_dict`).

5. `DiversifyRetriever` passes `lambda_` but the ES API field is lambda
Again the necessary context was not in the review it seems. This is handled in `DslBase._setattr` ([ref](https://github.com/elastic/elasticsearch-py/blob/00d3c7e9e5d876c13245a082941d732532a16773/elasticsearch/dsl/utils.py#L381)). I added `lambda_` to the `RESERVED` mapping. There's a test case for this (`test_diversify_serializes_lambda_without_trailing_underscore`).

Hope this clears everything up.

margaretjgu:
Thanks Tom, we're trialing some ai review tooling to facilitate us with reviews from a surge of external contributions....some may be hit or miss haha. Thanks for addressing the initial AI review at your own discretion. 

Since this is an enhancement and not impacting normal functionality, its currently lower priority on my review list. I will find time to review this within the next few weeks, thanks :)

tom-whitehead:
Hey @margaretjgu, understood, thank you very much.

On the AI reviewer front, the Claude code GHA works really well. It might need some prompt engineering, but you can check out the whole PR branch and also have a step to fetch additional context like the ES spec JSON file too.

utils/templates/retriever.py.tpl
```diff
@@ -0,0 +1,203 @@
...
+
+    # s.retriever = R("standard", query=s.query)
+    if hasattr(name_or_retriever, "_proxied"):
+        return cast(RProxiedProtocol[_T], name_or_retriever)._proxied
+
+    # "standard", query={"match": {"title": "python"}}
+    return Retriever.get_dsl_class(name_or_retriever)(**params)
+
+
+class Retriever(DslBase):
```
margaretjgu, line 134: `_as_retriever` is in `retriever.py` but not in this template. So the next generator run will wipe it, and `search_base.py` also imports it.... please copy the helper here.

utils/dsl-generator.py
margaretjgu: This looks a bit problematic... the union `_param_defs` change in dsl-generator applies to **every union**, not just retrievers...this is a breaking change. This may be the right fix for rrf's retrievers field, but it also means the next full codegen could change `query/aggs/field`. 

Can you run python utils/dsl-generator.py on this branch and say whether those files diff? 
if those two stay contained, I am not worried about the retriever classes themselves. Otherwise, I'm more keen on kjeping it `.extra(retriever=...)` which already works

tom-whitehead: Hey @margaretjgu thanks for reviewing. This isn't problematic, the union type hints generated here are exactly the same as before. I ran the full generation using `nox -s format` in the creation of this PR which runs the generator and formatting. There's no diff on any file other source file other than the creation of retrievers.py and the new interfaces added to `elasticsearch/dsl/types.py`, which is expected. `query`/`aggs`/`field` are all unchanged.

utils/templates/retriever.py.tpl
tom-whitehead: Yeah sorry good point I think I hacked it together then forgot to move it into the template. 

FYI, when I rerun the generation now there are some other changes in the spec that are going to get rolled into this PR. Notably, `DenseVectorIndexOptions` class gets updated to add a bunch of DiskBBQ-specific options.

utils/dsl-generator.py
tom-whitehead: Just re-ran and pushed a full generation now. As mentioned below some other spec changes on `DenseVectorIndexOptions` have been included in this PR as a result.

utils/templates/retriever.py.tpl
tom-whitehead: Done! 😃

PR #3567, opened 2026-09-16
Add a self-managed connection example to the getting started guide

### LINKED ISSUE #2811: Improve quick start guide with some hints on connecting to local hosts
In the quick start documentation page here: https://elasticsearch-py.readthedocs.io/en/v8.17.1/quickstart.html there's only the Cloud example, which is fine.
I think provide an example about non-cloud instances with one line may be helpful, like in this other page: https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/getting-started-python.html .

To be fair there's some hint about direct connection in [this notebook](https://github.com/elastic/elasticsearch-labs/blob/main/notebooks/search/00-quick-start.ipynb), but it is yet another way to do it.

(Another small note is why are there multiple ways to do the same thing? Which one is the preferred one?).

gohri81-tech:
The `Connecting` section of the getting started guide only showed the Elastic Cloud case, so readers who run Elasticsearch on their own machine had to leave the page to find out how to connect, which is what #2811 reports.

This adds a second example right after the Cloud one:

- a sync and an async snippet that connect to `http://localhost:9200` with an API key, using the same `tab-set` / `sync_or_async` structure as the rest of the page,
- a short paragraph about the secured case: the endpoint becomes `https://localhost:9200` and the client needs the certificate authority that Elasticsearch generated on the first start, with a pointer to `ca_certs`, `basic_auth` and `verify_certs` in [Connecting](https://github.com/elastic/elasticsearch-py/blob/main/docs/reference/connecting.md).

Documentation only.

Closes #2811


margaretjgu:
Closing. #2811 asks for a local connection line on the old `Read the Doc`s quickstart. That page is gone. Self managed connections are already in `docs/reference/connecting.md`: `https://localhost:9200` with the certificate authority when security is on, and `http://localhost:9200` only when security is off. Getting started already links there.

The snippet here uses `http://localhost:9200` with an API key. An API key means security is on, so that is the https case.

Please comment on the issue before opening a PR. We usually discuss the approach there, and the issue is assigned before someone starts on it.

PR #3568, opened 2026-09-16
Spell out that the API methods accept keyword arguments only

### LINKED ISSUE #2526: Unexpected ilm.put_lifecycle behavior
**Elasticsearch version** (`8.13.0`):

**`elasticsearch-py` version (`elasticsearch8==8.13.0`)**: 

From what I looked through in the API spec pages, this didn't seem like an issue there per se, but it still might be. In attempting to follow [the layout in `IlmClient.put_lifecycle`](https://github.com/elastic/elasticsearch-py/blob/main/elasticsearch/_sync/client/ilm.py#L363-L373), it shows that `name` is positional, and `policy` is a keyword arg. But when calling like that, I get:

```
File "/path/to/lib/python3.12/site-packages/elasticsearch8/_sync/client/utils.py", line 341, in wrapped
    raise TypeError(
TypeError: Positional arguments can't be used with Elasticsearch API methods. Instead only use keyword arguments.
```
If I call with `name='stringval', policy={'phases':{}}`, it works.

**Steps to reproduce**:

```
ilm_name = 'arbitrary'
phases = {'phases': {}}
client.ilm.put_lifecycle(ilm_name, policy=phases)
```

**Workaround**:

```
ilm_name = 'arbitrary'
phases = {'phases': {}}
client.ilm.put_lifecycle(name=ilm_name, policy=phases)
```

gohri81-tech:
In #2526 the reporter read the signature of `IlmClient.put_lifecycle`, took `name` to be positional and got:

```
TypeError: Positional arguments can't be used with Elasticsearch API methods. Instead only use keyword arguments.
```

The API reference did say "Keyword arguments are required for all calls", but as one sentence in the intro it is easy to miss when you land on a single method page.

This replaces that sentence with a note right under the intro that states the rule explicitly, shows the `TypeError` and shows the same call written with keyword arguments. The existing note about reserved keywords such as `from_` follows it, so both surprises are now next to each other.

Documentation only, no client change.

Closes #2526


margaretjgu:
Please comment on the issue before opening a PR. We usually discuss the approach there, and the issue is assigned before someone starts on it.

PR #3570, opened 2026-09-20
Fix optional Document fields under postponed annotations

### LINKED ISSUE #3554: Every field is required on python < 3.14 with from __future__ import annotations
Users that are not yet on python 3.14 probably have 
```
from __future__ import annotations
```
but that means PEP 563 annotations, that is, e.g. the string `'str | None'` and not a `UnionType[str, None]`. This is not understood by the elasticsearch-py type mapping, and makes every field accidentally required. 

For example, in Python 3.13:
```
from __future__ import annotations
import elasticsearch.dsl as es
from elasticsearch.dsl import mapped_field

class MyDoc(es.Document):
    name: str = mapped_field(es.Keyword(required=True))
    note: str | None = mapped_field(es.Keyword())

print('name._required =', MyDoc._doc_type.mapping['name']._required)   # expect True
print('note._required =', MyDoc._doc_type.mapping['note']._required)    # expect False, actually True
```

Not sure if there is anything that can be done except updating to 3.14, but I also didn't find anything in the docs or an existing report.

GautamVhavle:
## What

Evaluate postponed type annotations before the DSL decides whether a `Document` field is required. If a name cannot be resolved, keep the explicit Field's `required` and `multi` settings.

## Why

`from __future__ import annotations` stores hints such as `str | None` as strings. The mapper only treated live `Union` / `UnionType` objects as optional, so every annotated field became required. That also overwrote `Keyword(required=False)`.

Reproduced on current `main` (Python 3.13):

```python
from __future__ import annotations
from elasticsearch.dsl import Document, Keyword, mapped_field

class MyDoc(Document):
    name: str = mapped_field(Keyword(required=True))
    note: str | None = mapped_field(Keyword())

# before: note._required is True
# after:  note._required is False
```

Closes #3554

## How

- Resolve PEP 563 strings with `typing.ForwardRef` against the class module and class namespace.
- Only infer `required` / `multi` from annotations that actually evaluate.
- If evaluation fails and a Field is present, leave that Field's constructor settings alone.
- If evaluation fails and there is no Field, keep raising `TypeError: Cannot map field ...`.

This matches the fallback requested on the issue: understood annotations keep current behavior; unresolved names do not overwrite an explicit Field.

## Testing

Ran locally on Python 3.13:

- Original reporter reproduction, now passes
- New tests in `test_elasticsearch/test_dsl/_async/test_document.py` (sync copy generated):
  - optional `str | None` with `mapped_field(Keyword())`
  - optional `str | None` with no Field
  - unresolved name keeps `Keyword(required=False)` / `Keyword(required=True)`
  - unresolved name with no Field still raises
- `test_doc_with_type_hints` and `test_doc_with_pipe_type_hints`
- Full document unit tests: 115 passed
- All non-integration DSL tests: 553 passed

Live Elasticsearch integration tests were not run. They skip without a local cluster.

## Notes

No public API change. Python 3.10+ remains the supported range. I have signed, or will sign, the Elastic CLA.


GautamVhavle:
I've signed the Contributor Agreement. Please recheck.

